PR #51, opened 2014-05-14
Foreman Tasks utility script and wrapper

adamruzicka:
for exporting and viewing exported tasks in Dynflow console


ares:
what's the status here? seems useful based on the PR name :-)

adamruzicka:
Hmm, I can't recall why it didn't get in, but I'm sure it could be useful if we brought it back to life

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -19,268 +17,38 @@ If TASK_SEARCH is not defined, it defaults to all tasks in the past 7 days and
...
-      end
-    elsif format == 'csv'
-      CSV.open(export_filename, 'wb') do |csv|
-        csv << ['id', 'state', 'type', 'label', 'result', 'parent_task_id', 'started_at', 'ended_at']
-        tasks.each do |task|
-          csv << [task.id, task.state, task.type, task.label, task.result,
-                  task.parent_task_id, task.started_at, task.ended_at]
-        end
-      end
+    export_task = ::Dynflow::Tasks::Export.new do |t|
```
iNecas: What about having a child class overiding the world and plan methods instead?

adamruzicka: I think we talked about this and agreed on feeding the original rake task the necessary data instead of overriding parts of it. Fixed anyway

adamruzicka:
This was waiting for too long, let's close it for now.

PR #88, opened 2014-09-23
Fixes #9727 - trim action details on index page

jlsherrill:
otherwise actions with a lot of inputs cause the table column to
be very wide


iNecas:
@jlsherrill ping + failing tests

jlsherrill:
@iNecas apologies for this falling off the radar. 

I removed the use of h(), as it seemed to be double escaping the html when calling trunc_with_tooltip().  I tested with an action name including html and it seemed to be properly escaping it.

iNecas:
Thanks!

PR #93, opened 2015-01-21
Add package installation to README

eLobato:


iNecas:
Since it's just readme change, I'm ok with not having an issue opened for this

PR #99, opened 2015-02-23
Fixes #9508 - extract the triggering and waiting for sub plans from bulk actions

iNecas:
Bulk actions is just one of the use-cases for using sub-plans. This patch
extract the common code to be used for triggering external tasks in general.

The usage is as follows: one needs to inherit it's action from the
``Actions::ActionWithSubPlans`` and override the
``create_sub_plans`` method. The ``create_sub_plans`` will be executed in run
phase and the rest of the action will make sure that the execution of the
action will wait till the sub plan is done. It also tracks the relation
between parent and child tasks, as it already did in bulk actions.

An example code could look like this (taken from real Katello actions):

    class SyncAsSubPlan < Actions::ActionWithSubPlans
     input_format do
       param :id, Integer
     end

      def plan(repository)
       plan_self(:id => repository.id)
     end

      def create_sub_plans
       trigger(Repository::Sync, ::Katello::Repository.find(input[:id]))
     end
   end

### DIFF
--- a/app/lib/actions/action_with_sub_plans.rb
+++ b/app/lib/actions/action_with_sub_plans.rb
@@ -3,6 +3,7 @@ module Actions
   class Actions::ActionWithSubPlans < Actions::EntryAction
 
     middleware.use Actions::Middleware::KeepCurrentUser
+    middleware.use Actions::Middleware::SetCurrentTaskId
 
     SubPlanFinished = Algebrick.type do
       fields! :execution_plan_id => String,
@@ -78,7 +79,6 @@ module Actions
       planned, failed = sub_plans.partition(&:planned?)
 
       sub_plan_ids = ((planned + failed).map(&:execution_plan_id))
-      set_parent_task_id(sub_plan_ids)
 
       output[:total_count] = sub_plan_ids.size
       output[:failed_count] = failed.size
@@ -137,12 +137,6 @@ module Actions
 
     def counts_set?
       output[:total_count] && output[:success_count] && output[:failed_count]
-    end
-
-    def set_parent_task_id(sub_plan_ids)
-      ForemanTasks::Task::DynflowTask.
-          where(external_id: sub_plan_ids).
-          update_all(parent_task_id: task.id)
     end
 
     def check_targets!(targets)
--- a/app/lib/actions/action_with_sub_plans.rb
+++ b/app/lib/actions/action_with_sub_plans.rb
@@ -1,0 +1,162 @@
+module Actions
+
+  class Actions::ActionWithSubPlans < Actions::EntryAction
+
+    middleware.use Actions::Middleware::KeepCurrentUser
+
+    SubPlanFinished = Algebrick.type do
+      fields! :execution_plan_id => String,
+              :success           => type { variants TrueClass, FalseClass }
+    end
+
+    def plan(*args)
+      raise NotImplementedError
+    end
+
+    def run(event = nil)
+      case(event)
+      when nil
+        if output[:total_count]
+          resume
+        else
+          sub_plans = create_sub_plans
+          sub_plans = Array[sub_plans] unless sub_plans.is_a? Array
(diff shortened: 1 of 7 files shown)


iNecas:
@bbuckingham I've added the support for the locks of sub-plans to not collide with the parent's

bbuckingham:
@iNecas, I tested this PR out using a combination of foreman+katello+fusor to see the behavior of the fusor deploy action.  That action creates a set of subplan actions to sync a set of repos.  The result was successful (no lock conflicts or other errors).  Nice work!

app/lib/actions/action_with_sub_plans.rb
```diff
@@ -0,0 +1,156 @@
+module Actions
+
+  class Actions::ActionWithSubPlans < Actions::EntryAction
+
+    middleware.use Actions::Middleware::KeepCurrentUser
+    middleware.use Actions::Middleware::SetCurrentTaskId
+
+    SubPlanFinished = Algebrick.type do
+      fields! :execution_plan_id => String,
+              :success           => type { variants TrueClass, FalseClass }
```
pitr-ch: https://github.com/pitr-ch/algebrick/blob/master/lib/algebrick/types.rb#L31-L33 can be used instead

app/lib/actions/action_with_sub_plans.rb
```diff
@@ -0,0 +1,156 @@
...
+    end
+
+    def run(event = nil)
+      case(event)
+      when nil
+        if output[:total_count]
+          resume
+        else
+          sub_plans = create_sub_plans
+          sub_plans = Array[sub_plans] unless sub_plans.is_a? Array
```
pitr-ch: I think `Array()` is more common.

pitr-ch: the condition could be also removed then I think

app/lib/actions/middleware/set_current_task_id.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      private
+
+      def set_current_task_id
+        previous_task_id = Thread.current[:current_task_id]
+        Thread.current[:current_task_id] = action.task.id
+        yield
+      ensure
+        Thread.current[:current_task_id] = previous_task_id
+      end
```
pitr-ch: `Thread.current[:current_task_id]` is really bad :/ we need to come up with a better way. Maybe `ForemanTasks.trigger_sub(parent_task, *args)` or something like that ...

app/lib/actions/action_with_sub_plans.rb
iNecas: You can't do this, as the trigger result is an algerbick object, implementing `to_a` method

pitr-ch: you are right :)

bbuckingham:
@iNecas, Is there a target for getting this merged in and built?  Right now, I am using a checkout work with, but would like to get to gem/rpm soon.  We also need to see about getting it in to the downstream.  Thanks!!

iNecas:
@bbuckingham getting back to this, will try to address the concerns from the review (+ some preparation for the rollbacks, that the locks relate a bit), the updated PR should be ready tomorrow, with the expected release later this week. Would be friday ok with you?

bbuckingham:
@iNecas, Sounds good.  Thanks for the update and effort!

iNecas:
The updated PR is available at #104 (with dependent PR at https://github.com/Dynflow/dynflow/pull/145), closing this PR in favor of these two. The advantage is we don't need to use the Thread.current for passing the task id info + we have the native support for that in Dynlfow (so that one can even link to the sub plans from dynflow console).

@bbuckingham could you please test the new PR against your code? Nothing should be changed from the usage point of view.

bbuckingham:
@iNecas, Sure, I'll run a clean test using the code in the updated PRs.  Thanks!

PR #103, opened 2015-03-13
Fixes #9744 - don't include the test.rake in gem file

iNecas:


domcleal:
Looks reasonable to me!

PR #104, opened 2015-03-13
Fixes #9508 - use Dynflow sub-plans support for sub tasks and bulk actions

iNecas:
Bulk actions is just one of the use-cases for using sub-plans. This patch
extract the common code to be used for triggering external tasks in
general and moves that to the Dynflow.

The usage is as follows: one needs to inherit it's action from the
``Actions::ActionWithSubPlans`` and override the
``create_sub_plans`` method. The ``create_sub_plans`` will be executed in
run phase and the rest of the action will make sure that the execution of
the action will wait till the sub plan is done. It also tracks the relation
between parent and child tasks, as it already did in bulk actions.

An example code could look like this (taken from real Katello actions):

    class SyncAsSubPlan < Actions::ActionWithSubPlans
     input_format do
       param :id, Integer
     end

      def plan(repository)
       plan_self(:id => repository.id)
     end

      def create_sub_plans
       trigger(Repository::Sync, ::Katello::Repository.find(input[:id]))
     end
    end

### DIFF
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -77,6 +77,14 @@ module ForemanTasks
       self.state == 'paused'
     end
 
+    def self_and_parents
+      [self].tap do |ret|
+        if parent_task
+          ret.concat(parent_task.self_and_parents)
+        end
+      end
+    end
+
     def self.search_by_generic_resource(key, operator, value)
       key =  "resource_type" if key.blank?
       key_name = self.connection.quote_column_name(key.sub(/^.*\./,''))
--- a/app/lib/actions/action_with_sub_plans.rb
+++ b/app/lib/actions/action_with_sub_plans.rb
@@ -1,0 +1,30 @@
+module Actions
+
+  class Actions::ActionWithSubPlans < Actions::EntryAction
+
+    middleware.use Actions::Middleware::KeepCurrentUser
+
+    include Dynflow::Action::WithSubPlans
+
+    def plan(*args)
+      raise NotImplementedError
+    end
+
+    def humanized_output
+      return unless counts_set?
+      _("%{total} tasks, %{success} success, %{failed} fail") %
+          { total:   output[:total_count],
+            success: output[:success_count],
+            failed:  output[:failed_count] }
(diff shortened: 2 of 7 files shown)


iNecas:
Depends on https://github.com/Dynflow/dynflow/pull/145

bbuckingham:
Ran a test of fusor which uses ActionWithSubPlans and it ran cleanly, with no unexpected behaviors.  This included both this PR and the referenced dynflow PR.

app/models/foreman_tasks/task.rb
```diff
@@ -77,6 +77,14 @@ def paused?
       self.state == 'paused'
     end
 
+    def with_parents
+      [self].tap do |ret|
+        if parent_task
+          ret.concat(parent_task.with_parents)
+        end
+      end
```
pitr-ch: If you would like to make things more complicated `[self, *(parent_task.with_parents if parent_task)]` ;)

app/models/foreman_tasks/task.rb
```diff
@@ -77,6 +77,14 @@ def paused?
       self.state == 'paused'
     end
 
+    def with_parents
```
pitr-ch: could we think of a better name?

pitr-ch:
1 noop and 1 minor, otherwise ACK

app/models/foreman_tasks/task.rb
iNecas: not really: i'm out of inspiration here :)

iNecas: what about `self_and_parents`?

pitr-ch: yeah, why not

iNecas:
Comments addressed. Waiting for the CI, then merging and releasing new version of foreman-tasks

iNecas:
@bbuckingham foreman-tasks 0.6.13 and dynflow 0.7.7 containt this change.The PR for foreman-packaging is here https://github.com/theforeman/foreman-packaging/pull/582

bbuckingham:
Thanks @iNecas !

PR #106, opened 2015-04-30
Bkearney/10339

bkearney:


lib/tasks/export_tasks.rake
```diff
@@ -0,0 +1,241 @@
+#
+# export_tasks.rake is a debugging tool to extract tasks from the
+# current foreman instance.
+# 
+# Run "foreman-rake export_tasks" to export tasks whcih are not listed
+# as successful. 
+# To export all tasks "foreman-rake export_tasks tasks=all"
+
+desc 'Export Dynflow Tasks'
+  task :export_tasks => :environment do
```
iNecas: indentation -2

lib/tasks/export_tasks.rake
```diff
@@ -0,0 +1,241 @@
...
+        puts "#{count}/#{total}"
+        count += 1
+      end
+
+      File.open(File.join(tmp_dir, "index.html"), 'w') {|file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks)))}
+
+      system("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
+    end
+
+    puts "Created #{export_filename}"
```
iNecas: indentation -2

iNecas:
The first commit should not be there

lib/tasks/export_tasks.rake
iNecas: Also, it should be probably in 'foreman_tasks' namespace, as we do in https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/tasks/dynflow.rake#L1

bkearney: the indentation is at 2, whic his hte standard. yes?
Also.. on the namespace, do you want the file moved to that directory as well?

iNecas: Yes please: the directory under foreman_tasks/tasks would be better fit.
Yes: the standard is 2, but there are still too many 4-line ones, starting here https://github.com/theforeman/foreman-tasks/pull/106/files#diff-10af87cb135065f8f7ed35b14a284ccbR12

bkearney:
@iNecas  all changes have been applied

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -0,0 +1,244 @@
...
+
+      tasks.all.each do |task|
+        File.open(File.join(tmp_dir, "#{task.id}.html"), 'w') {|file| file.write(PageHelper.pagify(renderer.render_task(task)))}
+        puts "#{count}/#{total}"
+        count += 1
+      end
+
+      File.open(File.join(tmp_dir, "index.html"), 'w') {|file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks)))}
+
+      system("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
```
iNecas: `s/system/sh/g` http://rake.rubyforge.org/classes/FileUtils.html#M000018 the behaviour is the same as system, but the important benefit of usnig it is, if something fails when calling the command, the user gets an error describing the problem and the rake task fails without continuing.

iNecas:
Thanks, one last comment. Ack otherwise

bkearney:
@iNecas in there.

iNecas:
Thanks. Merging now

PR #110, opened 2015-06-12
refs #10730 - adds /tasks/bulk_resume

thomasmckay:
for use w/ 'hammer task resume'

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -76,6 +76,86 @@ module ForemanTasks
         render :json => ret
       end
 
+      api :POST, '/tasks/bulk_resume', N_('Resume all paused error tasks')
+      param :search, String, :desc => N_('Resume tasks matching search string')
+      param :task_ids, Array, :desc => N_('Resume specific tasks by id')
+      def bulk_resume
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        if params[:search].nil? && params[:task_ids].nil?
+          scope = scope.where(:state => :paused)
+          scope = scope.where(:result => :error)
+        end
+        scope = scope.where(:id => params[:task_ids]) if params[:task_ids]
+
+        resumed = []
+        failed = []
+        skipped = []
+        scope.find_each do |task|
+          if task.resumable?
+            begin
+              ForemanTasks.dynflow.world.execute(task.execution_plan.id)
+              resumed  << task_hash(task)
+            rescue RuntimeError => e
+              failed << task_hash(task)
+            end
+          else
+            skipped << task_hash(task)
+          end
+        end
+
+        render :json => {
+                          total: resumed.length + failed.length + skipped.length,
+                          resumed: resumed,
+                          failed: failed,
+                          skipped: skipped
+                        }
+      end
+
+      api :GET, '/tasks', N_("List tasks")
+      param :search, String, :desc => N_("Search string")
+      param :page, :number, :desc => N_("Page number, starting at 1")
+      param :per_page,  :number, :desc => N_("Number of results per page to return")
+      param :order, String, :desc => N_("Sort field and order, eg. 'name DESC'")
+      param :sort, Hash, :desc => N_("Hash version of 'order' param") do
+        param :by, String, :desc => N_("Field to sort the results on")
+        param :order, String, :desc => N_("How to order the sorted results (e.g. ASC for ascending)")
+      end
+      def index
+        scope =resource_scope.search_for(params[:search]).select('DISTINCT foreman_tasks_tasks.*')
+        total = scope.count
+
+        ordering_params =  {
+                             sort_by: params[:sort_by] || 'started_at',
+                             sort_order: params[:sort_order] || 'DESC'
+                           }
+        scope = ordering_scope(scope, ordering_params)
+
+
+        pagination_params = {
+                              page: params[:page] || 1,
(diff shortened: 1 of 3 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -76,6 +76,36 @@ def bulk_search
         render :json => ret
       end
 
+      api :PUT, '/tasks/bulk_resume', 'Resume all paused error tasks'
+      def bulk_resume
+        scope = resource_scope_for_index.select('DISTINCT foreman_tasks_tasks.*')
+        scope = scope.where(:state => :paused)
+        scope = scope.where(:result => :error)
```
iNecas: the result is not improtant from this point

iNecas: unless we want to resume also the force-unlocked task, which I agree we should, no problem then

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -76,6 +76,36 @@ def bulk_search
...
+
+        resumed = []
+        failed = []
+        skipped = []
+        scope.all.each do |task|
```
iNecas: we should probably use find_each to avoid loading all the tasks at once into the memmory http://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each

app/controllers/foreman_tasks/api/tasks_controller.rb
thomasmckay: I'll add some CLI options for the resume command. Any suggestions? I was thinking of --search (resume all that match the search) and --no-op (just print out results found, don't actually resume).

I do think the default should probably just be the ones in paused/error

iNecas: Ack on that: especially the search option accepting the scoped search can be really useful

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -76,6 +76,80 @@ def bulk_search
...
+      api :PUT, '/tasks/bulk_resume', 'Resume all paused error tasks'
+      def bulk_resume
+        scope = resource_scope_for_index.select('DISTINCT foreman_tasks_tasks.*')
+        scope = scope.where(:state => :paused)
+        scope = scope.where(:result => :error)
+
+        resumed = []
+        failed = []
+        skipped = []
+        scope.find_each.each do |task|
```
iNecas: the `.each` is not neccessary there

thomasmckay:
@iNecas - This should be ready for review. Thanks!

config/routes.rb
```diff
@@ -14,8 +14,9 @@
     end
 
     namespace :api do
-      resources :tasks, :only => [:show] do
+      resources :tasks, :only => [:show, :index] do
         post :bulk_search, :on => :collection
+        put :bulk_resume, :on => :collection
```
iNecas: any reason for choosing put over post?

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -76,6 +76,86 @@ def bulk_search
...
 
+      api :PUT, '/tasks/bulk_resume', N_('Resume all paused error tasks')
+      param :search, String, :desc => N_('Resume tasks matching search string')
+      param :task_ids, Array, :desc => N_('Resume specific tasks by id')
+      def bulk_resume
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        scope = scope.where(:state => :paused)
+        if params[:search].nil? && params[:task_ids].nil?
```
iNecas: I wonder if the `:state => :paused` should be also just a default scope, overriden by the :search/:task_ids option: reason: when you force unlock some task, it's switched into 'stopped' state, but you can still resume it.

iNecas:
I've tested that it works as expected, some minor comments inline, besides that, ready to merge

config/routes.rb
thomasmckay: PUT is update, POST is create

thomasmckay:
@iNecas seemingly random failures in jenkins unrelated?

config/routes.rb
iNecas: I don't see this actions as standard REST calls, but rather RPCs  in this case, so I would not be that strict about the this division: simplifying that and using POST for this would work for me a bit better, but not a big deal.

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -127,7 +211,7 @@ def active_scope(scope, search_params)
       end
 
       def action_types_scope(scope, search_params)
-        if action_types = search_params[:action_types]
+        if action_types == search_params[:action_types]
```
thomasmckay: @iNecas - i thought this was a simple typo but did you intend it to be an assignment inside an if statement?

iNecas: This is intended to be an assignment

iNecas:
@thomasmckay unrelated

app/controllers/foreman_tasks/api/tasks_controller.rb
thomasmckay: Agreed, moved paused into default section

config/routes.rb
thomasmckay: switched to POST to match other similar api calls

iNecas:
ACK from me: I can merge when you want

thomasmckay:
@iNecas merge away, thanks!

PR #112, opened 2015-06-18
refs #10820 - only collect 60 days of tasks by default

jlsherrill:
using export_tasks rake task


lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -216,6 +217,8 @@ namespace :foreman_tasks do
       tasks = ForemanTasks::Task.where("result != 'success'")
     end
 
+    days = ENV['days'].try(:to_i) || 60
```
iNecas: Maybe printing some informative message about the time limit, so that one has no expectations to get all by default

jlsherrill: done!

iNecas:
Thanks

PR #113, opened 2015-06-22
Fixes #8725 - consistency check on task data after executor start

iNecas:
- [x] add tests

### DIFF
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -17,7 +17,9 @@ module ForemanTasks
                                 end
                               end
       self.label          ||= main_action.class.name
+      changes             = self.changes
       self.save!
+      return changes
     end
 
     def resumable?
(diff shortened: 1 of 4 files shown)


iNecas:
@jlsherrill would you mind reviewing this?

iNecas:
Tests added

cc.patch
```diff
@@ -0,0 +1,52 @@
+commit ce62a96f9312825cb83d6b5ff90e4fe73072af89
```
jlsherrill: did you mean to commit this hash?

jlsherrill:
@iNecas i can't seem to get into the situation that i couldn't seem to get out of last week :)

Any pointers on trying to reproduce that in order to test this?

cc.patch
iNecas: good catch :)

iNecas:
@jlsherrill basically what's done in the tests: https://github.com/theforeman/foreman-tasks/pull/113/files#diff-27c3dbefe444dab41a4a669fe5296314R45

you can:

1.  go into the rails console; find the ForemanTasks::Task representing the currently running ListenOnCandlepin action
2. stop the dynflow executor (server in development, foreman-tasks service in production)
3. in the console, reload the task record (should be in 'stopped' state), update the state to 'running'
4. start the dynflow executor again
5. after it starts, reload the task object in console again: it should be 'stopped' again: the consistency check properly updated the state of the task

jlsherrill:
ACK, worked well, thanks @iNecas !

PR #114, opened 2015-06-22
Refs #10730 - find_each doesn't seem to work on string-based ids

iNecas:


iNecas:
@thomasmckay review please

iNecas:
When trying the bulk resume in production with postgres, I was getting this issue:

```

{
    "error": {
        "message": "PGError: ERROR:  operator does not exist: character varying >= integer\nLINE 1: ...result\" = 'error' AND (\"foreman_tasks_tasks\".\"id\" >= 0) ORDE...\n                                                             ^\nHINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.\n: SELECT  DISTINCT foreman_tasks_tasks.* FROM \"foreman_tasks_tasks\"  WHERE \"foreman_tasks_tasks\".\"state\" = 'paused' AND \"foreman_tasks_tasks\".\"result\" = 'error' AND (\"foreman_tasks_tasks\".\"id\" >= 0) ORDER BY \"foreman_tasks_tasks\".\"id\" ASC LIMIT 1000"
    }
}
```

I'm getting this PR in as the functionality obviously doesn't work as it should and this is a simple fix to make that working

PR #115, opened 2015-06-23
Refs #10905 - task status information in header 

dustints:
depends on https://github.com/Katello/hammer-cli-katello/pull/296

* shows tasks status counts in header

### DIFF
--- a/app/assets/javascripts/foreman_tasks/application.js
+++ b/app/assets/javascripts/foreman_tasks/application.js
@@ -1,0 +1,59 @@
+//= require_tree .
+$(document).ready(function(){
+  var taskCountUpdater = (function(){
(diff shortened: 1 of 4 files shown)


iNecas:
Rebase please

dustints:
@iNecas rebased

iNecas:
status

Some comments:
1. I would be more interested into state (running, paused), than result (pending error),
2. the error number in the status seems missing
3. there will be a lot of stopped/success tasks: a little point in showing that number
4. paused/success will be always zero: I would suggest also the drop down limit to paused and running (not including the result)
5. information about the user's own task might be equally useful

iNecas:
any updates?

PR #120, opened 2015-07-02
Refs #10767 - use the plugin widget API to register the widgets

iNecas:
The original approach was not designed for the plugins and it broke backward
compatibility with Foreman 1.8


iNecas:
@dustints could you review please?

domcleal:
@iNecas I'd suggest bumping requires_foreman to `>= 1.6` too (don't use `>`) since you're relying on `widget`

iNecas:
@domcleal good point, updated

iNecas:
[test]

PR #121, opened 2015-07-02
Fixes #10981 - Migrate logging to new custom logging in core

adamruzicka:


domcleal:
I'd recommend also shipping an example config (YAML) file with the loggers config commented out, so people can easily see the defaults and change it.  This also needs a change to requires_foreman to `>= 1.9.0`.

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -85,15 +85,16 @@ def get_humanized(method)
...
         begin
           changes = task.update_from_dynflow(task.execution_plan.to_hash)
           unless changes.empty?
             fixed_count += 1
-            Rails.logger.warn("Task %s updated at consistency check: %s" % [task.id, changes.inspect])
+            logger.warn("Task %s updated at consistency check: %s" % [task.id, changes.inspect])
           end
         rescue => e
-          Rails.logger.warn("Failed at consistency check for task %s: %s\n %s" % [task.id, e.message, e.backtrace.join("\n")])
+          logger.warn("Failed at consistency check for task %s: %s\n %s" % [task.id, e.message, e.backtrace.join("\n")])
```
iNecas: Could you also search for this kind of exception logging and convert that to use https://github.com/theforeman/foreman/blob/develop/lib/foreman/logging.rb#L60 ?

lib/foreman_tasks/engine.rb
```diff
@@ -4,6 +4,11 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_tasks.load_default_settings', :before => :load_config_initializers do
       require_dependency File.expand_path('../../../app/models/setting/foreman_tasks.rb', __FILE__) if (Setting.table_exists? rescue(false))
+      logging = Foreman::Plugin::Logging.new('foreman-tasks')
+      logging.configure(:loggers => {
```
domcleal: Use the plugin API if you can: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Custom-Plugin-Loggers

lib/foreman_tasks/engine.rb
```diff
@@ -4,6 +4,11 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_tasks.load_default_settings', :before => :load_config_initializers do
       require_dependency File.expand_path('../../../app/models/setting/foreman_tasks.rb', __FILE__) if (Setting.table_exists? rescue(false))
+      logging = Foreman::Plugin::Logging.new('foreman-tasks')
```
domcleal: That's already created by default.

lib/foreman_tasks/dynflow/daemon.rb
```diff
@@ -47,10 +47,11 @@ def run_background(command = "start", options = {})
                        :log_output => true,
                        :ARGV => [command]) do |*args|
         begin
+          ::Logging.reopen
           run(options[:foreman_root])
         rescue => e
           STDERR.puts e.message
-          Rails.logger.fatal e
+          Foreman::Logging.exception("", e)
```
iNecas: "Failed running foreman-tasks daemon"

iNecas:
ACK after fixing the last comment

iNecas:
Thanks!

PR #116, opened 2015-06-26
Refs #10962 - Convert foreman-tasks to use new dynflow version

iNecas:
Depends on https://github.com/Dynflow/dynflow/pull/139 (merged and released in dynflow 0.8.0)

PR #122, opened 2015-07-03
Refs #8725 - make sure to not run consistency checks before the db is migrated

iNecas:


lib/foreman_tasks/dynflow.rb
```diff
@@ -43,14 +43,17 @@ def initialize!
...
-          world.execute_planned_execution_plans
-
-          Task::DynflowTask.consistency_check
+          # don't try to do any rescuing until the tables are properly migrated
+          if !Foreman.in_rake?('db:migrate') && (ForemanTasks::Task.table_exists? rescue(false))
+            # for now, we can check the consistency only when there
+            # is no remote executor. We should be able to check the consistency
+            # every time the new world is created when there is a register
+            # of executors
+            world.consistency_check
```
ehelms: I am getting an undefined method when this is run

iNecas: Yes, there was a change on dynflow 0.8.0, when testing this PR with the 0.7.0 one, you need to update these lines the same way as https://github.com/theforeman/foreman-tasks/pull/116/files#diff-c48e58b017e4345d277b2cbf51e1b159L50, replacing the `consistency_check` and `execute_planned_exection_plans` with `auto_execute`

ehelms:
ACK for me

iNecas:
@ehelms thanks for reporting the issue and testing the fix!

PR #124, opened 2015-07-08
Fixes #10477 - log the dynflow output to proper directory

iNecas:
Use log directory instead of the tmp/pids, that the daemons use by default


ares:
:+1: :+1: :+1:

iNecas:
The failling test was not related to this PR, merging

PR #126, opened 2015-07-21
Fixes #11092 - fix tasks reloading

iNecas:
The previous implementation often caused the browser to crash due to way how
the reloading was implemented. It also fixes the behaviour in the two-pane
mode variant.

### DIFF
--- a/app/views/foreman_tasks/tasks/index.html.erb
+++ b/app/views/foreman_tasks/tasks/index.html.erb
@@ -14,11 +14,14 @@
 </style>
 
 <script>
+
+var currentTwoPaneTask;
+
 $(document).on('click', ".table-two-pane td.two-pane-link", function(e) {
-  var item = $(this).find("a");
-  if(item.length){
+  currentTwoPaneTask = $(this).find("a");
+  if(currentTwoPaneTask.length){
     e.preventDefault();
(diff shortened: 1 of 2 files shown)


bbuckingham:
Tested out this PR on my dev environment.  

The basic flow tested was:
- enable 4 RH repos
- go to Content -> Products
- select the RH product
- select the 4 repos
- click Sync Now
- then monitor the sub-tasks that were generated, (e.g. https://myserver/foreman_tasks/tasks/b4029787-a1a4-4e48-98f7-ed89dc89791f/sub_tasks )
- in doing so, click a task, observe it for a bit.. click another task, observe it for a bit...etc

I did not observe any crashing of the browser (note: using chrome).  I did observe one behavior (which may not be related to these changes, but something we may want to investigate).  

The behavior was:
- when I click the first subtask, I observed the browser polling the server 1 request a time
- when I click the second subtask, I observed the browser polling the server 2 requests at a time
- when I click the third subtask, I observed the browser polling the server 3 requests at a time
- etc.

In other words, the browser is sending redundant requests on a polling interval.

iNecas:
Thanks for pointing out this issue, I've fixed it.

bbuckingham:
ACK

PR #129, opened 2015-08-14
fixes #11358 - add type, value db index to tasks

jlsherrill:


iNecas:
There are indexes on both type and label, Is the combination of them really needed? What is the case it's needed?

iNecas:
ping @jlsherrill

iNecas:
Due to evidence of this helping with searching for a task of specific type, I'm getting it in

PR #130, opened 2015-08-16
fixes #11344 - enable timezone support

shlomizadok:


shlomizadok:
random failure? [test]

app/views/foreman_tasks/tasks/_details.html.erb
```diff
@@ -97,11 +97,11 @@
 </div>
 <div>
   <span class="param-name"><%= _("Started at") %>:</span>
-  <span class="param-value"><%= @task.started_at %></span>
+  <span class="param-value"><%= @task.started_at.in_time_zone %></span>
```
iNecas: `try(:in_time_zone)` would be a bit better, as the value does not have to be set always

app/views/foreman_tasks/tasks/_details.html.erb
```diff
@@ -97,11 +97,11 @@
...
 </div>
 <div>
   <span class="param-name"><%= _("Ended at") %>:</span>
-  <span class="param-value"><%= @task.ended_at %></span>
+  <span class="param-value"><%= @task.ended_at.in_time_zone %></span>
```
iNecas: ditto

app/views/foreman_tasks/tasks/index.html.erb
```diff
@@ -42,7 +42,7 @@ $(document).on('click', ".table-two-pane td.two-pane-link", function(e) {
       </td>
       <td><%= task.state %></td>
       <td><%= task.result %></td>
-      <td><%= task.started_at %></td>
+      <td><%= task.started_at.in_time_zone %></td>
```
iNecas: ditto

app/views/foreman_tasks/tasks/_details.html.erb
shlomizadok: Actually it is always set as in_time_zone falls back to UTC if no time zone is set.

app/views/foreman_tasks/tasks/_details.html.erb
shlomizadok: ditto ;-)

app/views/foreman_tasks/tasks/_details.html.erb
iNecas: It's more about the `started_at` being nil, no matter what timezone it's in

shlomizadok: Changed.

iNecas:
Thanks @shlomizadok works as expected!

PR #133, opened 2015-09-03
Fixes #11674 - Bookmarks for task states

xprazak2:
I've added sortable column `ended_at` and bookmarks that should point to potentially problematic tasks requiring user's attention.

### DIFF
--- a/db/seeds.d/61-foreman_tasks_bookmarks.rb
+++ b/db/seeds.d/61-foreman_tasks_bookmarks.rb
@@ -1,0 +1,12 @@
+Bookmark.without_auditing do
+  [
+    { :name => "running", :query => "state = running" },
+    { :name => "failed", :query => "state = paused or result = error or result = warning" }
+
+  ].each do |item|
+    next if Bookmark.find_by_name(item[:name])
+    next if audit_modified? Bookmark, item[:name]
+    b = Bookmark.create({:controller => "foreman_tasks_tasks", :public => true}.merge(item))
+    raise "Unable to create bookmark: #{format_errors b}" if b.nil? || b.errors.any?
+  end
+end
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -23,6 +23,7 @@ module ForemanTasks
     scoped_search :on => :state, :complete_value => true
     scoped_search :on => :result, :complete_value => true
     scoped_search :on => :started_at, :complete_value => false
+    scoped_search :on => :ended_at, :complete_value => false
     scoped_search :on => :parent_task_id, :complete_value => true
     scoped_search :in => :locks,  :on => :resource_type, :complete_value => true, :rename => "resource_type", :ext_method => :search_by_generic_resource
     scoped_search :in => :locks,  :on => :resource_id, :complete_value => false, :rename => "resource_id", :ext_method => :search_by_generic_resource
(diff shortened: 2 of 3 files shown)


db/migrate/20150903081526_add_foreman_tasks_bookmarks.rb
```diff
@@ -0,0 +1,25 @@
+class AddForemanTasksBookmarks < ActiveRecord::Migration
+  def up
+    Bookmark.where(:name => "pending",
```
iNecas: I would recude the number of different bookmarks: this states often don't tell anything to the end user. I would suggest two bookmars:

* running: state = running
* failed: state = paused OR result = error OR result = warning

db/migrate/20150903081526_add_foreman_tasks_bookmarks.rb
```diff
@@ -0,0 +1,25 @@
+class AddForemanTasksBookmarks < ActiveRecord::Migration
+  def up
+    Bookmark.where(:name => "pending", 
+                   :query => "result = pending", 
+                   :controller => "foreman_tasks_tasks",
+                   :public => true).first_or_create
+    Bookmark.where(:name => "paused",
```
iNecas: Seeds would be a better place for this than migrations, see https://github.com/theforeman/foreman/blob/develop/db/seeds.d/15-bookmarks.rb

xprazak2:
Number of bookmarks reduced to 2 and file moved from migrations to seeds.d

db/seeds.d/61-foreman_tasks_bookmarks.rb
```diff
@@ -0,0 +1,16 @@
...
+                   :controller => "foreman_tasks_tasks",
+                   :public => true).first_or_create
+    Bookmark.where(:name => "failed",
+                   :query => "state = paused || result = error || result = warning",
+                   :controller => "foreman_tasks_tasks",
+                   :public => true).first_or_create
+  end
+
+  def down
+    Bookmark.where(:controller => "foreman_tasks_tasks").map(&:destroy)
```
adamruzicka: This would destroy user-created bookmarks too, are we ok with it?
Wouldn't `#each` be more suited here, since we don't want to keep the results?

iNecas: The seeds are not migrations: so just moving the file to different place is not enough. I guess inspiring more form the way we seed the bookmarks in foreman is the best thing we can do https://github.com/theforeman/foreman/blob/develop/db/seeds.d/15-bookmarks.rb

iNecas: Also, it seems `and` and `or` operations are somehow preferred over `&&` and `||`

xprazak2:
I changed it into the same format as we have in Foreman.

db/seeds.d/61-foreman_tasks_bookmarks.rb
```diff
@@ -0,0 +1,12 @@
+Bookmark.without_auditing do
+  [
+    { :name => "running", :query => "state = running" },
+    { :name => "failed", :query => "state = paused or result = error or result = warning" }
+
+  ].each do |item|
+    next if Bookmark.find_by_name(input[:name])
+    next if audit_modified? Bookmark, input[:name]
```
iNecas: `s/item/input/`

iNecas:
Works well, merging!

PR #134, opened 2015-09-03
Fixes #11682: Raise Foreman::Exception for bulk action errors

ehelms:


ehelms:
For some context, this was inspired by this PR (https://github.com/Katello/katello/pull/5357#issuecomment-135787867) and the need to detect exceptions of this type, and display to the user a message that makes sense.

iNecas:
Thanks @ehelms

PR #141, opened 2015-09-24
Refs #10755 - Support for recurring actions and task groups

adamruzicka:
TODO:
- [x] recurring actions
- [x] task groups
- [x] permissions
- [x] tests
- [ ] improve UI
- [ ] ability to clone (and edit) recurring logic
- [x] cleanup stale comments

### Task Groups
Do this in action's plan method
```ruby
task.add_missing_task_groups(task_groups)
```

### Recurring logic
Use the middleware in an action
```ruby

middleware.use ::Actions::Middleware::RecurringLogic
``` 

To start a recurring action
```ruby
recurring_logic = ::ForemanTasks::RecurringLogic.new_from_cronline('* * * * *')
recurring_logic.start(action_class, plan_arg1, plan_arg2...)
```

~~Task Groups (for actions with sub plans)
Subtasks which want to be in the same task groups as the task which spawned them must use the InheritTaskGroups middleware. This middleware makes the child task inherit task groups from the parent before its plan phase and it makes the parent collect task groups from its children at the end of its run phase. That means the parent task has to set its task groups during the plan phase. At the end of the parent's run phase, it adds the children's task groups to its own.~~

~~tl;dr version:~~
+ ~~`middleware.use InheritTaskGroups`~~
+ ~~the children inherit task groups from the parent before the children's plan phase~~
+ ~~parent adds all of children's task groups to its own at the end of parent's run phase~~
+ ~~sibling tasks don't modify each others' task groups~~

### DIFF
--- a/app/controllers/foreman_tasks/recurring_logics_controller.rb
+++ b/app/controllers/foreman_tasks/recurring_logics_controller.rb
@@ -1,0 +1,33 @@
+module ForemanTasks
+  class RecurringLogicsController < ::ApplicationController
+
+    before_filter :find_recurring_logic, :only => [:show, :cancel]
+
+    def index
+      @recurring_logics = filter(resource_base)
+    end
+
+    def show
+    end
+
+    def cancel
+      @recurring_logic.cancel
+      redirect_to :action => :index
+    end
+
+    def controller_name
+      'foreman_tasks_recurring_logics'
+    end
+
+    private
+
+    def find_recurring_logic
+      @recurring_logic ||= ::ForemanTasks::RecurringLogic.find(params[:id])
+    end
+
+    def filter(scope)
+      scope.search_for(params[:search]).paginate(:page => params[:page])
+    end
+
+  end
+end
--- a/app/helpers/foreman_tasks/foreman_tasks_helper.rb
+++ b/app/helpers/foreman_tasks/foreman_tasks_helper.rb
@@ -1,0 +1,144 @@
+module ForemanTasks
+  module ForemanTasksHelper
+
+    def recurring_logic_state(recurring_logic)
+      icon, status = case recurring_logic.state
+             when 'active'
+               'glyphicon-info-sign'
+             when 'finished'
+               ['glyphicon-ok-sign', 'status-ok']
+             when 'cancelled'
+               ['glyphicon-warning-sign', 'status-error']
+             else
+               'glyphicon-question-sign'
+             end
+      content_tag(:i, '&nbsp'.html_safe, :class => "glyphicon #{icon}") + content_tag(:span, recurring_logic.humanized_state, :class => status)
+    end
+
+    def recurring_logic_action_buttons(recurring_logic)
+      buttons = []
+      if authorized_for(:permission => :edit_recurring_logics, :auth_object => recurring_logic)
+        buttons << link_to(N_("Cancel"), cancel_foreman_tasks_recurring_logic_path(recurring_logic), :method => :post, :class => 'btn btn-danger') unless %w(cancelled finished).include? recurring_logic.state
+      end
+      button_group buttons
+    end
+
+    def recurring_logic_next_occurrence(recurring_logic)
+      if %w(finished cancelled).include? recurring_logic.state
+        N_('-')
+      else
+        recurring_logic.next_occurrence_time
+      end
+    end
+
+    def time_f(f, attr, field_options = {}, time_options = {}, html_options = {})
+      f.fields_for attr do |fields|
+        field(fields, attr, field_options) do
+          fields.time_select attr, time_options, html_options
+        end
+      end
+    end
+
+    def datetime_f(f, attr, field_options = {}, datetime_options = {}, html_options = {})
+      f.fields_for attr do |fields|
+        field(fields, attr, field_options) do
+          fields.datetime_select attr, datetime_options, html_options
+        end
+      end
+    end
+
+    def inline_checkboxes_f(f, attr, field_options = {}, checkboxes = {}, options = {})
+      field(f, attr, field_options) do
+        checkboxes.map do |key, name|
+          [f.check_box(key, options), " #{name} "]
+        end.flatten.join('')
+      end
+    end
+
+    def trigger_selector(f, triggering = Triggering.new, options = {})
+      render :partial => 'common/trigger_form', :locals => { :f => f, :triggering => triggering }
+    end
+
+    private
+
+    def future_mode_fieldset(f, triggering)
+      tags = []
+      tags << text_f(f, :start_at_raw, :label => _('Start at'), :placeholder => 'YYYY-mm-dd HH:MM')
+      tags << text_f(f, :start_before_raw, :label => _('Start before'), :placeholder => 'YYYY-mm-dd HH:MM')
+      content_tag(:fieldset, nil, :id => "trigger_mode_future", :class => "trigger_mode_form #{'hidden' unless triggering.future?}") do
+        tags.join.html_safe
+      end
+    end
+
+    def recurring_mode_fieldset(f, triggering)
+      tags = []
+      tags << selectable_f(f, :input_type, %w(cronline monthly weekly daily hourly), {}, :label => _("Repeats"), :id => 'input_type_selector')
+      tags += [
+        cronline_fieldset(f, triggering),
+        monthly_fieldset(f, triggering),
+        weekly_fieldset(f, triggering),
+        time_picker_fieldset(f, triggering)
+      ]
+
+      content_tag(:fieldset, nil, :id => 'trigger_mode_recurring', :class => "trigger_mode_form #{'hidden' unless triggering.recurring?}") do
+        tags.join.html_safe
+      end
+    end
+
+    def cronline_fieldset(f, triggering)
+      options = [
+        _('is minute (range: 0-59)'),
+        _('is hour (range: 0-23)'),
+        _('is day of month (range: 1-31)'),
+        _('is month (range: 1-12)'),
+        _('is day of week (range: 0-6)')
+      ].map { |opt| content_tag(:li, opt) }.join
+      help = content_tag(:span, nil, :class => 'help-inline') do
+        popover(_('Explanation'),
+                _("Cron line format 'a b c d e', where:<br><ol type=\"a\">#{options}</ol>"))
+      end
+      content_tag(:fieldset, nil, :class => "input_type_form #{'hidden' unless triggering.input_type == :cronline}", :id => "input_type_cronline") do
+        text_f f, :cronline, :label => _('Cron line'), :placeholder => '* * * * *', :help_inline => help
+      end
+    end
+
+    def monthly_fieldset(f, triggering)
+      content_tag(:fieldset, nil, :id => "input_type_monthly", :class => "input_type_form #{'hidden' unless triggering.input_type == :monthly}") do
+        text_f(f, :days, :label => _('Days'), :placeholder => '1,2...')
+      end
+    end
+
+    def weekly_fieldset(f, triggering)
+      content_tag(:fieldset, nil, :id => 'input_type_weekly', :class => "input_type_form #{'hidden' unless triggering.input_type == :weekly}") do
+        f.fields_for :days_of_week do |days_of_week|
+          inline_checkboxes_f(days_of_week,
+                              :weekday,
+                              { :label => _("Days of week") },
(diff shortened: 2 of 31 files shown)


db/migrate/20150907131503_create_task_groups.rb
```diff
@@ -0,0 +1,20 @@
+class CreateTaskGroups < ActiveRecord::Migration
+  def up
+    create_table :foreman_tasks_task_groups do |t|
+      t.string :type, index: true, null: false
+      t.integer :recurring_logic_id, index: true
```
iNecas: Shouldn't this association param be on the `recurring_logics` to avoid the sparse table (other TaskGroup STI tables would not care about this

app/models/foreman_tasks/task_group.rb
```diff
@@ -0,0 +1,8 @@
+module ForemanTasks
+  class TaskGroup < ActiveRecord::Base
+
+    has_and_belongs_to_many :tasks,
```
iNecas: I have pretty bad experience with habtm, could we introduce `TaskGroupMember` model, or something like that and use `has_many :though` instead?

app/views/foreman_tasks/recurring_logics/index.html.erb
```diff
@@ -0,0 +1,29 @@
+<% title _("Recurring logics") %>
+<% title_actions help_path%>
```
iNecas: missing space at the end of ` %>`

app/lib/actions/recurring_logic/create.rb
```diff
@@ -0,0 +1,17 @@
+module Actions
+  module RecurringLogic
+    class Create < EntryAction
+
+      middleware.use Actions::Middleware::InheritTaskGroups
+
+      def plan(recurring_logic)
+        recurring_logic.task_group = ::ForemanTasks::TaskGroups::RecurringLogicTaskGroup.new
```
iNecas: Why separate action for this? Shouldn't this be part of the model (before create?)

app/lib/actions/recurring_action.rb
```diff
@@ -0,0 +1,62 @@
...
+      else
+        task.add_missing_task_groups(recurring_logic.task_group)
+        plan_self :action => action.to_s,
+                  :serialized_args => serialized_args
+      end
+      trigger_repeat(recurring_logic, action, *args) if recurring_logic.can_continue?
+    end
+
+    def create_sub_plans
+      [input[:action].constantize].map { |action_class| trigger action_class, *deserialized_args }
```
iNecas: The `create_sub_plans` dosn't have to return just array: single trigger call is just fine, no need to do the `map` there

app/lib/actions/recurring_action.rb
```diff
@@ -0,0 +1,62 @@
...
+    middleware.use Actions::Middleware::InheritTaskGroups
+
+    def delay(delay_options, recurring_logic, action, *args)
+      task.add_missing_task_groups(recurring_logic.task_group)
+      action_class = action.is_a?(Class) ? action : action.constantize
+      action = action.to_s if action.is_a?(Class)
+      action_class.serializer_class.new [recurring_logic, action] + args
+    end
+
+    def plan(recurring_logic, action, *args)
```
iNecas: `action_class` would be a bit better for reading

app/lib/actions/recurring_action.rb
```diff
@@ -0,0 +1,62 @@
...
+      serialized_args = serializer_class.new(args).tap { |s| s.perform_serialization! }.serialized_args
+      if recurring_logic.new_record?
+        plan_action(RecurringLogic::Create, recurring_logic)
```
iNecas: This seems like something we could do in a model (to initiate the delayed execution), to limit the behaviour of this action to only run the sub-task and delay the next execution.

app/controllers/foreman_tasks/recurring_logics_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  class RecurringLogicsController < ::ApplicationController
+
+    def index
+      @recurring_logics = filter(resource_base)
+    end
+
+    private
+
+    def filter(scope)
+      scope.search_for(params[:search]).
```
iNecas: single line woul de just enough here

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -0,0 +1,53 @@
+module ForemanTasks
+
+  require 'parse-cron'
+
+  class RecurringLogic < ActiveRecord::Base
+    include Authorizable
+
+    has_one :task_group,
```
iNecas: One line should be enough for everyone here

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -0,0 +1,53 @@
...
+            :class_name => TaskGroups::RecurringLogicTaskGroup.name
+
+    delegate :tasks, :to => :task_group
+
+    scoped_search :on => :id, :complete_value => false
+    scoped_search :on => :max_iteration, :complete_value => false, :rename => :iteration_limit
+    scoped_search :on => :iteration, :complete_value => false
+    scoped_search :on => :cronline, :complete_value => true
```
iNecas: validating a cron_line might be a good idea

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  require 'parse-cron'
+
+  class RecurringLogic < ActiveRecord::Base
+    include Authorizable
+
+    has_one :task_group,
+            :class_name => TaskGroups::RecurringLogicTaskGroup.name
+
+    delegate :tasks, :to => :task_group
```
iNecas: something line `has_many :tasks, :through => :task_group` would be better, as it leverages the activerecord AREL

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -0,0 +1,53 @@
...
+      (end_time.nil? || next_occurrence_time(time) < end_time) &&
+        (max_iteration.nil? || iteration < max_iteration)
+    end
+
+    def finished?
+      !can_continue? && task_group.tasks.all? { |task| %w(stopped paused).include? task.state }
+    end
+
+    def cancel_tasks
+      tasks.reject { |task| task.state != 'stopped' }.each { |task| task.cancel }
```
iNecas: using scope would be more efective, the `active` scope shold be the best fit, as sending cancel event to paused tasks wouldn't do anything anyway https://github.com/theforeman/foreman-tasks/blob/master/app/models/foreman_tasks/task.rb#L35

iNecas: `.each { |task| cancel}/.seach(&:cancel)`

app/lib/actions/middleware/recurring_logic.rb
```diff
@@ -0,0 +1,31 @@
+module Actions
+  module Middleware
+    class RecurringLogic < ::Dynflow::Middleware
+
+      def delay(delay_options, *args)
```
iNecas: Some high-level comment describing the purpose of this middleware might help further generations to understand the intention

app/lib/actions/middleware/recurring_logic.rb
```diff
@@ -0,0 +1,31 @@
...
+        pass(delay_options, *args).tap do
+          if delay_options[:recurring_logic_id]
+            task.add_missing_task_groups(recurring_logic(delay_options).task_group)
+          end
+        end
+      end
+
+      def plan(*args)
+        pass(*args).tap do
+          task_group = task.task_groups.find { |tg| tg.class == ::ForemanTasks::TaskGroups::RecurringLogicTaskGroup }
```
iNecas: `tg.is_a? ::ForemanTasks::TaskGroups::RecurringLogicTaskGroup` would read a bit better.

iNecas:
The middleware version works much nicely! I'm missing ability to cancel the recurring logic.

iNecas:
I'm thinking about recurring logic details page, which seems missing right now. Ideally, it should be customized by the specific use, so for the remote execution, there would be a list of job invocations it triggered, etc. Also in the recurring logics index page, there should be info about what task it is about, when was the last occurrence, etc.

app/helpers/foreman_tasks/foreman_tasks_helper.rb
```diff
@@ -0,0 +1,190 @@
...
+          [f.check_box(key, options), " #{name} "]
+        end.flatten.join('')
+      end
+    end
+
+    def trigger_selector(f, options = {})
+      tags = []
+      form_name = f.options[:html][:id]
+      script =<<-END
+          | $(function() { trigger_#{form_name}_selector_binds() });
```
iNecas: Why not extracting this code into a js file?

app/helpers/foreman_tasks/foreman_tasks_helper.rb
```diff
@@ -0,0 +1,190 @@
...
+          |     var o = form.find('fieldset#trigger_mode_recurring fieldset#end_time_limit_form');
+          |     if($(this).val() === 'true') {
+          |       o.show();
+          |     } else {
+          |       o.hide();
+          |     };
+          |   });
+          | };
+      END
+      f.fields_for 'triggering' do |trigger_fields|
```
iNecas: What about extracting the html generation into the erb, and just rendering from here, as we do in foreman https://github.com/theforeman/foreman/blob/develop/app/helpers/application_helper.rb#L42

adamruzicka:
Would it make sense to have something like `FutureTrigger` and `RecurringTrigger` as separate classes extending the `ImmediateTrigger` base class?

app/views/foreman_tasks/recurring_logics/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+      <td><%= link_to(recurring_logic.cron_line, foreman_tasks_recurring_logic_path(recurring_logic)) %></td>
+      <td><%= link_to(recurring_logic.tasks.count, foreman_tasks_tasks_url(:search => "task_group.id = #{recurring_logic.task_group.id}")) %></td>
+      <td><%= format_task_input(recurring_logic.tasks.first, true) %></td>
+      <td><%= recurring_logic.tasks.order(:started_at).where('started_at IS NOT NULL').last.try(:started_at) || N_("-") %></td>
+      <td><%= recurring_logic_next_occurrence recurring_logic %></td>
+      <td><%= recurring_logic.iteration %></td>
+      <td><%= format_recurring_logic_limit recurring_logic.max_iteration %></td>
+      <td><%= format_recurring_logic_limit recurring_logic.end_time.try(:in_time_zone) %></td>
+      <td><%= recurring_logic_state(recurring_logic) %></td>
+      <!-- TODO: Permissions -->
```
iNecas: TODO before merge or after merge? If after, we should have an issue covering that

app/views/foreman_tasks/recurring_logics/_tab_related.html.erb
```diff
@@ -0,0 +1,4 @@
+<div>
+  <%= render 'foreman_tasks/task_groups/tab_related', :task_group => task_group %>
+</div>
+
```
iNecas: Trailing line

app/models/foreman_tasks/triggering.rb
```diff
@@ -0,0 +1,94 @@
...
+        parse_start_at!
+        parse_start_before!
+      end
+    end
+
+    ALLOWED_MODES = [:immediate, :future, :recurring]
+    ALLOWED_INPUT_TYPES = [:cronline, :monthly, :weekly, :daily, :hourly]
+
+    TIME_FORMAT = "%Y-%m-%d %H:%M"
+    TIME_REGEXP = /^\d{4}-\d{2}-\d{2} \d{2}:\d{2}$/
```
iNecas: `\A` (`\Z`) should be preferred over `^` (`$`), compare:

```
line = "hello\nhi"
line =~ /^hello$/
# => 0
line =~ /\Ahello\Z/
# => nil
```

iNecas:
Dynflow 0.8.8 with the required changes released

iNecas:
Waiting for green tests

adamruzicka:
@iNecas the test are green now

iNecas:
ACK, lets squash the commits and merge

iNecas:
Thanks for the patience @adamruzicka I know it took a while to get this in :)

PR #143, opened 2015-09-30
Refs #11895 - optimize humanized data loading

iNecas:
On tasks index page, call only humainzed methods that are needed. Also, cache
the humanized data on first call, as the humanization might be relatively
expensive sometimes. Also, we catch all the exceptions when calling the
humanization, as it's not critical and the errors at humanization should not
prevent the user to see other task data.


iNecas:
[test]

iNecas:
The CI was failing due to https://github.com/theforeman/foreman/pull/2771, which should be resolved, but I don't think we need to wait for the results on this one.

PR #145, opened 2015-10-02
Fixes #12041 - Add async support detection

ares:


ares:
https://github.com/theforeman/foreman/pull/2776 is required to be merged first, that's why it also bumps Foreman requirements to 1.10+

ares:
Bumping Foreman dependency should also imply higher minor (at least) version, so maybe that would be time for 0.8.

app/controllers/foreman_tasks/concerns/hosts_controller_extension.rb
```diff
@@ -8,16 +8,20 @@ module HostsControllerExtension
...
-        task = ForemanTasks.async_task(::Actions::Foreman::Host::ImportFacts,
-                                       detect_host_type,
-                                       params[:name],
-                                       params[:facts],
-                                       params[:certname],
-                                       detected_proxy.try(:id))
+        type = params.fetch(:facts, {}).fetch(:_type, {}) || 'puppet'
+        trigger_type = FactImporter.importer_for(type).support_background ? :async_task : :sync_task
+
+        task = ForemanTasks.public_send(trigger_type,
```
iNecas: What if we moved https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks.rb#L40 logic from `sync_task` to `trigger_task` and then used the `trigger_task` here as well?

ares: +1 for that, but should I push it into this PR? would that deserve some version bump? it seems like public API

ares:
Rebased, the dependency is bumped to 1.11 as the linked commit did not make it to 1.10.

ares:
what's the status here?

adamruzicka:
Is this still needed? If yes, it will need a rebase, otherwise it would be good to close it

ares:
honestly I don't know already :-) closing this until I figure out

ares:
actually I think we need it for non-puppet cfgmgmt systems that would like to use background processing, reopening with hope I'll get to rebasing

adamruzicka:
Is this still needed? If yes, it will need a rebase, otherwise it would be good to close it

adamruzicka:
@ares ping

ares:
This is now rebased. Note that in order to use this, you must set FOREMAN_TASKS_MONKEYS env variable to true https://github.com/theforeman/foreman-tasks/blob/2f4735cb6ed882529fb02a5c9701013c5a8eb74b/lib/foreman_tasks/engine.rb#L131

I think it's still valuable for ansible facts parsing. Perhaps we could even now remove the need for the env variable.

adamruzicka:
CC @ezr-ondrej , @xprazak2 need more eyes on this

@ares While you're at it, could you take a look at the action it is using? Currently there's some backward compatibility stuff for foreman < 1.16 which could be dropped. Also I'd expect the action to mimic what the controller currently does. They seem to have divereged a bit over the time

app/controllers/foreman_tasks/concerns/hosts_controller_extension.rb
```diff
@@ -1,13 +1,23 @@
...
-        task = ForemanTasks.async_task(::Actions::Foreman::Host::ImportFacts,
-                                       detect_host_type,
-                                       params[:name],
-                                       params[:facts],
-                                       params[:certname],
-                                       detected_proxy.try(:id))
+      extend ActiveSupport::Concern
+
+      included do
+        alias_method_chain :facts, :dynflow
```
ezr-ondrej, line 7: We should prepend instead

ares:
Thanks, closing as no longer relevant

PR #147, opened 2015-11-12
Fixes #12460 - Rails 4 compatible changes with Rails 3 (find_or_create/where)

dLobatog:
Some parts of foreman-tasks use find_or_create, that have to be replaced
by Rails 4 syntax. We can do this with where and first_or_create and
have it working on both Rails versions.


johnpmitsch:
can we close in favor of #148 ?

PR #148, opened 2015-11-19
rails4 compatability - updated

komidore64:


cpeters:
ACK

config/foreman-tasks.yaml
```diff
@@ -0,0 +1,26 @@
+:foreman-tasks:
```
unorthodoxgeek: why did you add this file?

unorthodoxgeek:
it's just a styling thing, but I prefer this:

```ruby
Foo.where(bar: "baz").first
```

over this:
```ruby
Foo.find_by(bar: "baz")
```

It keeps using the same API everywhere, and it just seems like the more "right" thing to do... But it's just an opinion, doesn't have to be done.

Also - did you run tests against a rails4 version of foreman?

unorthodoxgeek:
tried running tests, got this:
```
test/controllers/api/tasks_controller_test.rb:1:in `require': cannot load such file -- foreman_tasks_test_helper (LoadError)
```

komidore64:
@unorthodoxgeek, we use `find_by` in Katello and I guess that just bled into our fixes of this project as well.

unorthodoxgeek:
I'm very confused, why did this PR close? what are the other PRs?

cpeters:
@unorthodoxgeek 

I created a redmine issue for the rails4 upgrade and a pull request.
https://github.com/theforeman/foreman-tasks/pull/151

PR #153, opened 2015-12-08
Refs #12145 - Adds start_before explanation

adamruzicka:
basically theforeman/foreman_remote_execution#75

@stbenjam I'd still get this in to make the UI at least a bit less confusing until http://projects.theforeman.org/issues/12686 is done


stbenjam:
Yea that's cool, APJ

iNecas:
Thanks!

PR #151, opened 2015-12-01
Fixes #12654 - Rails 4 compatability changes

cpeters:

PR #155, opened 2015-12-09
Refs #10970 - Fixes task details page

johnpmitsch:
find_resource is not working when showing a task page in katello in Rails 4, It will not use the full uuid as the id param.


iNecas:
Should be fixed as part of #151

PR #156, opened 2015-12-10
fixes #12763 - puppetclass importer should be async

unorthodoxgeek:


app/lib/actions/foreman/puppetclass/import.rb
```diff
@@ -0,0 +1,32 @@
+module Actions
+  module Foreman
+    module Puppetclass
+      class Import < Actions::EntryAction
+        def resource_locks
+          :import_puppetclasses
+        end
+
+        def plan(params)
+          plan_self(params)
```
iNecas: This is the dafult implementation of the plan method in the action - no need to redefine unless other intention

app/controllers/foreman_tasks/concerns/environments_extension.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      included do
+        alias_method_chain :obsolete_and_new, :dynflow
+      end
+
+      def obsolete_and_new_with_dynflow
+        task = ForemanTasks.async_task(::Actions::Foreman::Puppetclass::Import,
+                                       params)
+
+        notice _("Added import task to queue, it will be run shortly")
```
iNecas: A link to the task details would be interesting here.

unorthodoxgeek:
@iNecas - fixed according to comments, please re-review

iNecas:
[test]

app/controllers/foreman_tasks/concerns/environments_extension.rb
iNecas: I have note realized the link is not shown properly in the notification window, so it just renders in page escaped. Sry about that - we would need a better infra in UI to make it better. So the version is ok.

lib/foreman_tasks/engine.rb
```diff
@@ -69,6 +69,8 @@ class Engine < ::Rails::Engine
 
       config.to_prepare do
         ::Api::V2::HostsController.send :include, ForemanTasks::Concerns::HostsControllerExtension
+        ::PuppetclassesController.send :include, ForemanTasks::Concerns::EnvironmentsExtension
```
iNecas: Do we need to extend this controller as well? I don't see it defining the `obsolete_and_new`

unorthodoxgeek: well, technically, the definition is at Foreman::Controller::Environments.

I could include it there, but this seems to be more verbose...

iNecas: Right, I have misread it completely before (haven't noticed the `Foreman::Controller::Environments` is a concern (I would love to see the modules in Foreman to follow the directory structure…). With this fact, I agree this is better.

app/controllers/foreman_tasks/concerns/environments_extension.rb
unorthodoxgeek: reverted the reverted reverted code.

iNecas: :ok_hand: :)

iNecas:
Thanks @unorthodoxgeek

PR #157, opened 2015-12-10
Fixes #12766 - Picks right days when using weekly recurring

adamruzicka:


iNecas:
It seems it should not be hard to put some tests there as well.

iNecas:
Thanks @adamruzicka

PR #160, opened 2016-01-04
Fixes #13006 - unable to view task details page

johnpmitsch:
sessions are lazily loaded, and @rack_request.session will return ```#<ActionDispatch::Request::Session:0x7f76982e76b8 not yet loaded>
``` if not accessed first. This seems to have broke with the rails 4 changes


iNecas:
#159 fixes the issue, Thanks @johnpmitsch for the report and patch

PR #161, opened 2016-01-05
Fixes #13001 - use start_at for immediate tasks as well

ares:


ares:
not sure why [test] failed

iNecas:
[test] the reason why the tests failed should be fixed now

ares:
looks better :green_apple:

stbenjam:
Hm, not sure I like the closeness of the two names, started_at and start_at is confusing.

Why not scheduled_at?

ares:
do you want to change that existing column in foreman-tasks or rename scoped_search definition?

iNecas:
Let's talks about the naming somewhere else (an issue or different PR) as it's not directly about what's this PR about. ACK for the actual subject of this PR

stbenjam:
You're adding start_at as a user-exposed searching option, that seems relevant to discuss here. 

I guess I'm a bit late to the party since the column is named, but I'd find it confusing as a non-native speaker to have two fields named so close together.

iNecas:
Since the `start_at` will now be set bot for queued and immediate executions, I'm not having issues with the name, and as a non-native speaker, I have not issues with that either (although I'm pretty biased:)

ares:
wouldn't `scheduled_at` indicate scheduling in future? tbh I don't have strong opinion but I'd prefer having the search name same as DB attribute and `start_at` sounds best to me, maybe renaming `started_at` to something else would be better since it won't be used that often...

stbenjam:
Why wouldn't started_at be used? In the other PR, my point was I want to look at the job invocations for ones that started in the last 24 hours.  

I don't care, I give up, I think it's named poorly to have two fields like this but apparently I'm too late.

ares:
> Why wouldn't started_at be used? 

I'd say that `start_at` is more useful information. You'd probably want to see all tasks that should have started in last 24 hours, using `started_at` you wouldn't see those that were not picked, in case of dynflow not running, you wouldn't see any I think. If you'd want to see tasks in specific state you could combine `start_at` with `state`.

Anyway I'd suggest opening a foreman-tasks refactoring issue to rename one of the column.

iNecas:
Agreed

PR #162, opened 2016-01-05
Fixes #13007 - fix the way we determine test to be skipped

iNecas:


stbenjam:
ACK from me, looks good

PR #164, opened 2016-01-07
Refs #13007 - compatibility fix to run tests against foreman 1.10

iNecas:


stbenjam:
I do wonder how stable this will be, you're touching minitest internals here, right? Can we not protect the callback with a permission?

Ideally foreman's plugin interface should let us tell it that some routes should be allowed to not have permissions.

iNecas:
Yes, it's not stable, as we can see, and I agree the Foreman's plugin interface, or something else to determine, so far I'm not aware of any other way that this one, I'm ok with getting this in, and opening another PR for us to not have to do again later.

stbenjam:
Yea that's fine, I started tests against 1.10-stable here: http://ci.theforeman.org/view/Plugins/job/test_plugin_pull_request/2002/

APJ

iNecas:
Both versions are passing now, thanks @stbenjam for triggering the 1.10 tests

iNecas:
0.7.10 released, just with this little fix to unblock the rex

stbenjam:
@iNecas Thanks!!! :tada:  Packaging PR's opened.

PR #167, opened 2016-01-20
Fixes #13291 - change scheduling form styling

ares:


iNecas:
ACK, tested and works great

PR #171, opened 2016-02-12
Fixes #13578 - remove the at_exit hook to terminate the world

iNecas:
Since https://github.com/Dynflow/dynflow/commit/4e3a002bd5f798090bf9d9896c5c625695c386d2#diff-35ca92bf5b10414a154a2844535ab5e9R29
we already are terminating the dynflow world in at_exit hook. Doing this
outside for Dynflow can lead to calling `exit` twice on termination finish.


mmoll:
fixes the reported issue for me :+1:

PR #172, opened 2016-02-12
Fixes #13641 - fix the ordering of the tasks menu items

iNecas:
The previous version override the "Reports" label of the Foreman core


ares:
ACK

PR #174, opened 2016-02-16
fixes #13743 - remove 'Recurring logics' menu

thomasmckay:
I removed the menu but left the infrastructure in case there's somewhere less obtrusive to put it.


stbenjam:
Well, it needs to go somewhere - what about a button on the Tasks page?

thomasmckay:
Not returning to this.

PR #176, opened 2016-02-19
Fixes #13814 - update monkey patch for facts importing

ares:


app/models/foreman_tasks/concerns/host_action_subject.rb
```diff
@@ -15,26 +15,30 @@ def available_locks
...
         # +importHostAndFacts+ method into two
-        def importHost(hostname, certname, proxy_id = nil)
+        def import_host(hostname, certname, facts, proxy_id = nil)
+          raise(::Foreman::Exception.new("Invalid Facts, must be a Hash")) unless facts.is_a?(Hash)
           raise(::Foreman::Exception.new("Invalid Hostname, must be a String")) unless hostname.is_a?(String)
 
           # downcase everything
           hostname.try(:downcase!)
           certname.try(:downcase!)
+          facts['domain'].try(:downcase!)
```
stbenjam: Where's the domain get used here?

ares: nowhere, probably should be moved to facts importing when https://github.com/theforeman/foreman-tasks/pull/176/files#diff-a2cb9fe73d15f4e6310e2ef57f76e22bR16 is implemented, good catch but I'm not sure where to put it here, I don't want to wrap another method (Host::Base#import_facts)... I plan to open the split PR against core soon

stbenjam: It's very not obvious, because I guess the facts are passed around by reference and you're modifying it here so it's downcased later?

Maybe create a normalize_facts method and call it from `Plan`? At least it'd be clearer what's going on imho.

ares:
it will likely be replaced by https://github.com/theforeman/foreman/pull/3242, I'll will close if it's accepted

ares:
It was merged upstream, I think we still need it for Foreman 1.11.

app/models/foreman_tasks/concerns/host_action_subject.rb
ares: moved to the plan

iNecas:
How does that change in post 1.11? Any change needed for newer versions of Foreman?

ares:
1.12 already has the logic split so the monkey patching of Host is not needed and should be removed. On the other hand it still requires this patch since the code of importing method has changed. Note that this patch is 1.12 compatible, it's just cloning some code that is already present in 1.12. So as a step 2, we can bump Foreman dependency (introduced Foreman tasks 0.8.0) and remove the monkey patch.

iNecas:
[test]

iNecas:
Could you bump the foreman required version to 1.11 https://github.com/ares/foreman-tasks/blob/fix/13814/lib/foreman_tasks/engine.rb#L39 as part of this PR?

ares:
Actually this code does not bring any incompatibility with older versions. The monkey patch only brings a fix for ignoring puppet proxy configuration for non-puppet imports to older versions and should be compatible with both 1.9 and 1.10.

iNecas:
Thanks @ares

iNecas:
Released in foreman-tasks 0.7.16

PR #177, opened 2016-02-22
Fixes #13833 - Handle humanized input with delayed task

adamruzicka:


ares:
Works fine and fixes the issue, could we return some value like 'N/A', '...', 'unavailable', or 'pending' instead of returning just nil? Otherwise I only see "Run hosts job:" before it gets scheduled.

Otherwise :+1:

iNecas:
[test]

ares:
APJ

iNecas:
[test]

PR #181, opened 2016-04-18
Fixes #14681 - Make end_at timepicker scale better

adamruzicka:


iNecas:
Tested and works great, merged as https://github.com/theforeman/foreman-tasks/commit/329c978a2b9c6240824820d946f9ee51dfc1330c Thanks @adamruzicka

PR #182, opened 2016-04-18
Fixes #14684 - Saves end time

adamruzicka:


adamruzicka:
This should be solved in remote execution

PR #184, opened 2016-04-25
Fixes #14798 - Provide a way to call save without dynflow wrap

ares:


app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -45,6 +45,14 @@ def update_attributes!(*args)
         dynflow_task_wrap(:save) { super(*args) }
       end
 
+      # this can be used when HostActionSubject is used for orchestration but you want to avoid
+      # triggering more tasks by ActiveRecord callbacks within run/finalize phase of your task
+      def save_with_disabled_dynflow_hooks
+        @_disable_dynflow_hooks = true
+        save
```
iNecas: what about just calling `save_without_dynflow_task_wrap` from here. if you have reasons not to, you should save the original value `@_disable_dynflow_hooks` and restore it in `ensure` block

ares:
this concept can be used for other actions like destroy, that's why I prefer `@_disable_dynflow_hooks`, using ensure and original value now

iNecas:
Thinking about the other cases (destroy update), what about:

```
host.disable_dynflow_hooks { |h| h.destory } # would disable it just in the block
host.disable_dynflow_hooks # would disable it on the record until calling `enable_dynflow_hooks`
```

ares:
sounds good, better than setting `@_disable_dynflow_hooks` directly, please see new version

iNecas:
[test]

PR #186, opened 2016-05-16
Fixes #14487 - Export filtered tasks in csv format

daviddavis:

### DIFF
--- a/lib/foreman_tasks/tasks/cleanup.rake
+++ b/lib/foreman_tasks/tasks/cleanup.rake
@@ -3,21 +3,21 @@ namespace :foreman_tasks do
     desc <<DESC
 Clean tasks based on filter and age. ENV variables:
 
-  * FILTER     : scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
-  * AFTER      : delete tasks created after *AFTER* period. Expected format is a number followed by the time unit (s,h,m,y), such as '10d' for 10 days
-  * STATES     : comma separated list of task states to touch with the cleanup, by default only stopped tasks are covered
-  * NOOP       : set to "true" if the task should not actuall perform the deletion
-  * VERBOSE    : set to "true" for more verbose output
-  * BATCH_SIZE : the size of batches the tasks get processed in (1000 by default)
+  * TASK_SEARCH : scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
+  * AFTER       : delete tasks created after *AFTER* period. Expected format is a number followed by the time unit (s,h,m,y), such as '10d' for 10 days
+  * STATES      : comma separated list of task states to touch with the cleanup, by default only stopped tasks are covered
+  * NOOP        : set to "true" if the task should not actuall perform the deletion
+  * VERBOSE     : set to "true" for more verbose output
+  * BATCH_SIZE  : the size of batches the tasks get processed in (1000 by default)
 
-If none of FILTER, BEFORE, STATES is specified, the tasks will be cleaned based
+If none of TASK_SEARCH, BEFORE, STATES is specified, the tasks will be cleaned based
 configuration in settings
 DESC
     task :run => 'environment' do
       options = {}
 
-      if ENV['FILTER']
-        options[:filter] = ENV['FILTER']
+      if ENV['TASK_SEARCH']
+        options[:filter] = ENV['TASK_SEARCH']
       end
 
       if ENV['AFTER']
@@ -38,6 +38,10 @@ DESC
 
       if ENV['BATCH_SIZE']
         options[:batch_size] = ENV['BATCH_SIZE'].to_i
+      end
+
+      if ENV['FILTER']
+        fail "FILTER has been deprecated. Please use TASK_SEARCH instead."
       end
 
       ForemanTasks::Cleaner.run(options)
--- a/lib/foreman_tasks/tasks/export_tasks.rake
+++ b/lib/foreman_tasks/tasks/export_tasks.rake
@@ -2,15 +2,31 @@
 # export_tasks.rake is a debugging tool to extract tasks from the
 # current foreman instance.
 #
-# Run "foreman-rake export_tasks" to export tasks whcih are not listed
-# as successful.
-# To export all tasks "foreman-rake export_tasks tasks=all"
-#   to specify the number of days of tasks to gather:  days=60 (defaults to 60)
+# Run "foreman-rake foreman_tasks:export_tasks" to export tasks
+
+require 'csv'
 
 namespace :foreman_tasks do
-  desc 'Export Dynflow Tasks'
+  desc <<DESC
+Export dynflow tasks based on filter. ENV variables:
 
+  * TASK_SEARCH     : scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
+  * TASK_FILE       : file to export to
+  * TASK_FORMAT     : format to use for the export (either html or csv)
+  * TASK_DAYS       : number of days to go back
+
+If TASK_SEARCH is not defined, it defaults to all tasks in the past 7 days and
+all unsuccessful tasks in the past 60 days. The default TASK_FORMAT is html
+which requires a tar.gz file extension.
+DESC
   task :export_tasks => :environment do
+    deprecated_options = {:tasks => "TASK_SEARCH",
+                          :days => "TASK_DAYS",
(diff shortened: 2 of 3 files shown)


iNecas:
Funny thing: I was checking the ability to influence what tasks get exported with foreman-debug like this (after changing foreman-rake script a bit as described in my commant https://github.com/theforeman/foreman/pull/3534#issuecomment-220356120 - just removing the `-` form the su command):
```
export FILTER='started_at > 2016-05-19' 
foreman-debug
```

I got a cryptic message first, just to find out that foreman-debug already uses the `FILTER` env variable to something else https://github.com/theforeman/foreman/blob/develop/script/foreman-debug#L55.

I suggest renaming the `FILTER` in our case to `TASKS_SEARCH` both here and in the tasks cleanup script to avoid this kind of possible conflicts.

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -218,35 +231,37 @@ namespace :foreman_tasks do
...
 
-      File.open(File.join(tmp_dir, "index.html"), 'w') {|file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks)))}
+        File.open(File.join(tmp_dir, "index.html"), 'w') {|file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks)))}
 
-      sh("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
+        sh("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
+      end
+    elsif format == 'csv'
+      CSV.open(export_filename, 'wb') do |csv|
+        csv << ['id', 'state', 'type', 'label']
```
iNecas: `result`, `started_at`, `ended_at` and `parent_task_id` are pretty useful data to gather as well.

iNecas:
I've tested the patch and it works, few small things found.

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -218,35 +231,37 @@ namespace :foreman_tasks do
...
-      tasks = ForemanTasks::Task
-    else
-      tasks = ForemanTasks::Task.where("result != 'success'")
-    end
-
-    days = ENV['days'].try(:to_i) || 60
-    tasks = tasks.where('started_at > ?', days.days.ago)
-    export_filename = ENV['export'] || "/tmp/task-export-#{DateTime.now.to_i}.tar.gz"
+    tasks = ForemanTasks::Task.search_for(filter)
+    export_filename += (format == 'csv' ? '.csv' : '.tar.gz')
```
iNecas: I would suggest to add the suffix just in case the file name is autogenerated. It feels strange to pass `FILE=my_file.csv` and the data exported in `my_file.csv.csv`

daviddavis: Yea, the only thing that is kind of weird is that we use a `tar.gz` for html formatted exports. Maybe I can put a note about that in the task description though?

iNecas: +1

ekohl:
How about outputting to stdout like standard unix tools?

iNecas:
@ekohl I can image that for the csv format, not that much for the full task export, but I can see a variant, where the FILE would be applicable only for the html format

ekohl:
@iNecas you're right that I was only thinking of the CSV-output

daviddavis:
Ok, I think I addressed everything. I created a redmine for supporting STDOUT for the CSV format. Let me know if you'd prefer me to include that change here and I can.

ekohl:
I think it'd pretty easy to implement ```-``` as the stdout filename here

daviddavis:
If I add STDOUT support I have to content with things like the `puts` calls in this file. Given I'm not really happy with the code in this PR and I hate to add more technical debt, I am almost wondering if for 3.0 I should just try to parse the tar.gz file in the installer and leave rewriting this file for after katello 3.0. Thoughts?

ekohl:
At first I thought it should be simple enough, but after giving it some thought I think you could be right. You don't want to be stuck supporting some quick hack.

iNecas:
So what's the status of this PR?

daviddavis:
@inecas I'm waiting to see what we decide on in https://github.com/Katello/katello-installer/pull/337.

daviddavis:
[test]

iNecas:
Thanks, putting to waiting on contributor for now

iNecas:
Given it also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1314207, I would be ok with getting this in regardless https://github.com/Katello/katello-installer/pull/337. @daviddavis anything you would like to cover here before merging?

daviddavis:
No, looks like there are some intermittent test failures but it should be good to merge.

iNecas:
[test]

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -2,15 +2,29 @@
...
+
+require 'csv'
 
 namespace :foreman_tasks do
-  desc 'Export Dynflow Tasks'
+  desc <<DESC
+Export dynflow tasks based on filter. ENV variables:
+
+  * TASK_SEARCH     : scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
+  * TASK_FILE       : file to export to (minus the file extension)
```
iNecas: (minus the file extension) does not hold anymore

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -2,15 +2,29 @@
...
+  * TASK_FORMAT     : format to use for the export (either html or csv)
 
+If TASK_FILTER is not defined, it defaults to unsuccessful tasks newer than 60
+days ago. The default TASK_FORMAT is html which requires a tar.gz file extension.
+DESC
   task :export_tasks => :environment do
+    deprecated_options = {:tasks => "TASK_SEARCH",
+                          :days => "TASK_SEARCH",
```
iNecas: the days is still pretty useful option IMO, compared to users having to format the date one their own.

iNecas:
Last two comments and we should be ready to go

daviddavis:
@iNecas updated

lib/foreman_tasks/tasks/cleanup.rake
```diff
@@ -3,21 +3,21 @@ namespace :foreman_tasks do
     desc <<DESC
 Clean tasks based on filter and age. ENV variables:
 
-  * FILTER     : scoped search filter (example: 'label = "Actions::Foreman::Host::ImportFacts"')
```
iNecas: README needs to be updated with the change

lib/foreman_tasks/tasks/cleanup.rake
```diff
@@ -3,21 +3,21 @@ namespace :foreman_tasks do
...
+  * BATCH_SIZE  : the size of batches the tasks get processed in (1000 by default)
 
-If none of FILTER, BEFORE, STATES is specified, the tasks will be cleaned based
+If none of TASK_SEARCH, BEFORE, STATES is specified, the tasks will be cleaned based
 configuration in settings
 DESC
     task :run => 'environment' do
       options = {}
 
-      if ENV['FILTER']
```
iNecas: We should raise an error here. Since this is not being called from the foreman-debug, there is no issue with FILTER interfering with other scripts.

iNecas:
Tested and works well, I know I've already said it should be ready to go after the fixes, but some more found. Sorry about that. It should READY be ready to go after that :)

lib/foreman_tasks/tasks/cleanup.rake
daviddavis: You mean like a deprecation error (e.g. `Please use TASK_SEARCH instead of FILTER`)?

iNecas: Yes, to preven't a situation, where somebody would have some filter set to
delete
some tasks, and after this change, we would just delete everything because
no filter was set

-- Ivan

daviddavis:
@iNecas I think it should be good now.

daviddavis:
Looks like the ruby 2.1+ tests are passing as fast_gettext dropped support for < 2.1.

iNecas:
[test]

daviddavis:
@iNecas all green 👍

iNecas:
Thanks @daviddavis, merging now

PR #187, opened 2016-05-19
Fixes #15091 - Handle user timezones properly

adamruzicka:
Also fixes issue where displayed times were depended on local system time (having system set as gmt+2 and user timezone as gmt-11 would lead to displaying times in the gmt-9 zone.


iNecas:
https://github.com/theforeman/foreman-tasks/pull/187/files#diff-02c838ba9b04851fb6e35da3b1835685R46 needs to be zone-specific as well

iNecas:
ACK, tested and works well,  let's squash the commits

iNecas:
Thanks @adamruzicka

PR #61, opened 2014-06-03
Online task import/export

adamruzicka:
requires https://github.com/Dynflow/dynflow/pull/115

PR #191, opened 2016-07-07
Fixes #15598: Output task plan in JSON format

ehelms:
Provides the task execution plan output in JSON format, similar to
the dynflow console. This same JSON output is provided via export.
This allows users to export into a known and parseable format for
advanced processing.

New API:

/foreman_tasks/api/tasks/5c743d80-359c-4dcf-ab66-bf11a7fa2288/plan

Export:

rake foreman_tasks:export_tasks json=true

### DIFF
--- a/app/presenters/foreman_tasks/execution_plan_presenter.rb
+++ b/app/presenters/foreman_tasks/execution_plan_presenter.rb
@@ -1,0 +1,40 @@
+module ForemanTasks
+  class ExecutionPlanPresenter
+
+    attr_accessor :task
+
+    def initialize(task)
+      @task = task
+    end
+
+    def plan
+      output = []
+      @execution_plan = @task.execution_plan
+      output << {'run' => process_flow(@execution_plan.run_flow)}
+      output << {'finalize' => process_flow(@execution_plan.finalize_flow)}
+      output
+    end
+
+    def process_flow(flow)
+      case flow
+      when ::Dynflow::Flows::Sequence
+        steps = flow.flows.collect do |step|
+          process_flow(step)
+        end
+        {'type' => 'sequence', 'steps' => steps}
+      when ::Dynflow::Flows::Concurrence
+        steps = flow.flows.collect do |step|
+          process_flow(step)
+        end
+        {'type' => 'concurrence', 'steps' => steps}
+      when ::Dynflow::Flows::Atom
+        {'type' => 'atom', 'step' => process_atom(flow)}
+      end
+    end
+
+    def process_atom(atom)
+      @execution_plan.steps[atom.step_id].to_hash
+    end
+
+  end
+end
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -29,7 +29,7 @@ module ForemanTasks
       class BadRequest < Apipie::ParamError
       end
 
-      before_filter :find_task, :only => [:show]
+      before_filter :find_task, :only => [:show, :plan]
 
       api :GET, "/tasks/summary", "Show task summary"
       def summary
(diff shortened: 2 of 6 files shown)


adamruzicka:
> This engine [foreman-tasks] is agnostic on background processing tool and can be used with anything...

This implementation is tightly bound to Dynflow. It might be better to have a generic `PlanPresenter` defining an interface, which other backend-specific presenters could implement and provide the bits and pieces specific to those backends and select the right presenter implementation according to the used backend.

Tasks can have a parent-child relationship with other tasks. It would be nice to have at least the IDs of task's parent and children in the export

This takes only the `Run` phase into account, meaning if an execution plan fails in the `Plan` or `Finalize` phase, you won't get any useful info from the export. 
```
# Export of task which failed in plan phase
{"id":"1425aaeb-6e86-4b7e-b46a-2993cced38fe","plan":{"type":"concurrence","steps":[]}}
```

ehelms:
I have added one update which is to break it down by phases and include the finalize but I could not figure out how to include the plan phase output:

```
{
id: "5c743d80-359c-4dcf-ab66-bf11a7fa2288",
  phases: [
    {
       run: {}
    },
    {
      finalize: {}
    }
  ]
}
```

1) What are the other backends outside of Dynflow that are supported today? How does one from the backend from the execution plan or task?

2) What are the IDs that you speak of? Right now the output for each atom is and I see a few IDs, if not those, how to fetch the ID you would like included:

```
step: {
execution_plan_id: "fd49dc92-ada4-43e9-8d93-42e5fff3ab39",
id: 5,
state: "success",
class: "Dynflow::ExecutionPlan::Steps::RunStep",
action_class: "Actions::Candlepin::Environment::Create",
action_id: 4,
error: null,
started_at: "2016-06-29 03:06:49",
ended_at: "2016-06-29 03:06:49",
execution_time: 0.090849526,
real_time: 0.090849526,
progress_done: 1,
progress_weight: 1
}
```

adamruzicka:
> I have added one update which is to break it down by phases and include the finalize but I could not figure out how to include the plan phase output:

You could do something like this to get all the steps in plan phase
```ruby
# Variant 1
root = @execution_plan.root_plan_step
[root] + root.planned_steps(@execution_plan)

# Variant 2
@execution_plan.steps.values.select { |step| step.is_a? ::Dynflow::ExecutionPlan::Steps::PlanStep }
```

1) Well, currently there are none which I know of. But either we should try to keep an open door for people to write additional backend adapters, or stop claiming that we're backend agnostic.

2) I meant tasks, not steps. There are `#sub_tasks` and `#parent_task` associations on the `ForemanTasks::Task` object (see [link](https://github.com/theforeman/foreman-tasks/blob/master/app/models/foreman_tasks/task.rb#L12)), which are used when you have an action with sub-plans

app/presenters/foreman_tasks/execution_plan_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+      output = []
+      @execution_plan = @task.execution_plan
+      output << {'run' => process_flow(@execution_plan.run_flow)}
+      output << {'finalize' => process_flow(@execution_plan.finalize_flow)}
+      output
+    end
+
+    def process_flow(flow)
+      case flow
+      when ::Dynflow::Flows::Sequence
```
iNecas: This should be provided by Dynflow. Probably time to refresh https://github.com/Dynflow/dynflow/pull/115

iNecas: The reason for not having this in foreman-tasks is to have the same format in foreman_tasks and smart_proxy_dynflow.

iNecas:
I think the the multi-backend support is not crutial, but it helps reasoning on what belongs to dynflow and what belongs to foreman tasks.

BTH. I would rather refresh https://github.com/theforeman/foreman-tasks/pull/51, https://github.com/theforeman/foreman-tasks/pull/61 and https://github.com/Dynflow/dynflow/pull/115 than extending the export rake, that proved already to be quite fragile to changes

iNecas:
Closing in favor of https://github.com/Dynflow/dynflow/pull/115

ehelms:
All for handling the export in dynflow but I feel like the API should
provide me a way to get this info per task.

iNecas:
I'm ok with that, but the API should use the Dynflow export capability for that purpose instead of using the Dynflow internals here for that purpose.

ehelms:
+1 want to keep this open for that purpose then? I'll adapt it

iNecas:
Agreed

PR #192, opened 2016-07-08
Fixes #15624 - support for filtering data in presentation

iNecas:
Depends on:

 - [x] https://github.com/Dynflow/dynflow/pull/190

This allows to set a `present` middleware to filter the data that are
shown to the user. To use it, one needs to define a `present`
middleware:

    class MyMiddleware < ::Dynlfow::Middleware
      def present
        action.output['password'] = '******'
        pass
      end
    end

and then load the middleware, either globally:

    initializer "my_plugin.set_dynflow_middlewares", :before => :finisher_hook do
      ForemanTasks.dynflow.config.on_init do |world|
        world.middleware.use MyMiddleware
      end
    end

or inside action (or parent action) to be used with:

    class MySecretAction < Dynflow::Action
       middleware.use MyMiddleware
    end

This should affect all the input/output data displayed to user: either
in tasks details or dynflow conosle, but should not affect the
execution.

### DIFF
--- a/foreman-tasks.gemspec
+++ b/foreman-tasks.gemspec
@@ -26,7 +26,7 @@ DESC
   s.test_files = `git ls-files test`.split("\n")
   s.extra_rdoc_files = Dir['README*', 'LICENSE']
 
-  s.add_dependency "dynflow", '~> 0.8.8'
+  s.add_dependency "dynflow", '~> 0.8.13'
   s.add_dependency "sequel" # for Dynflow process persistence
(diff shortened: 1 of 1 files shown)


djwhatle:
Thanks for the quick PR's Ivan! Erik Nelson tested this out in a sandbox environment and it looks like it will suit our needs well.

Is there a BZ already existing for getting this feature into downstream? If not, I'll create one.

Again, thanks for your help.

adamruzicka:
Tested and works great, ACK

PR #194, opened 2016-07-22
Fixes #15778 - fix DynflowTask route helpers

ares:


ares:
I suggest bumping the version after this patch is merged. We'll need to point to some foreman-tasks version when Foreman starts using it for background processing. It seems this is only thing that's missing.

iNecas:
Thanks @ares

PR #195, opened 2016-08-11
Fixes #16065 - convert to strong params

iNecas:


adamruzicka:
Tested with REX, works as expected, ACK

PR #197, opened 2016-08-30
Fixes #16113 - Display tasks in planning state properly

adamruzicka:
Second commit solves this issue in a bit cleaner way, but requires https://github.com/Dynflow/dynflow/pull/196


adamruzicka:
Dependencies updated

iNecas:
Thank you @adamruzicka

PR #198, opened 2016-09-02
Refs #16058 - extract common code for remote execution and ansible

iNecas:
- [x] add tests - cheating a bit to track it as short term issue http://projects.theforeman.org/issues/16465
- [x] update remote execution

### DIFF
--- a/app/lib/actions/helpers/with_delegated_action.rb
+++ b/app/lib/actions/helpers/with_delegated_action.rb
@@ -1,0 +1,57 @@
+module Actions
+  module Helpers
+    module WithDelegatedAction
+      include ::Actions::Helpers::WithContinuousOutput
+
+      def plan_delegated_action(proxy, klass, options)
+        case proxy
+        when :not_defined
+          if klass.is_a?(String)
+            raise ::Foreman::Exception.new(_('No proxy defined for execution'))
+          else
+            delegated_action = plan_action(klass, options)
+          end
+        when :not_available
+          raise ::Foreman::Exception.new(_('All proxies with the required feature are unavailable at the moment'))
+        when ::SmartProxy
+          delegated_action = plan_action(::Actions::ProxyAction, proxy, klass, options)
+        end
+
+        input[:delegated_action_id] = delegated_action.id
+        delegated_action
+      end
+
+      def humanized_output
+        delegated_output
+      end
+
+      def continuous_output_providers
+        super.tap do |ret|
+          ret << delegated_action if delegated_action.respond_to?(:fill_continuous_output)
+        end
+      end
+
+      def delegated_output
+        return @delegated_output if @delegated_output
+        action = delegated_action
+        @delegated_output = case action
+                            when NilClass
+                              {}
+                            when ::Actions::ProxyAction
+                              action.proxy_output(true)
+                            else
+                              action.output
+                            end
+      end
+
+      def delegated_action
+        # TODO: make it easier in dynflow to load action data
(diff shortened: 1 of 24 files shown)


app/lib/actions/helpers/with_delegated_action.rb
```diff
@@ -0,0 +1,52 @@
...
+    module WithDelegatedAction
+
+      include ::Dynflow::Action::Cancellable
+      include ::Actions::Helpers::WithContinuousOutput
+
+      def plan_delegated_action(proxy, klass, options)
+        case proxy
+        when :not_defined
+          if klass.is_a?(String)
+            raise _(_'No proxy defined for execution')
```
adamruzicka: One `_` should be enough

app/lib/actions/helpers/with_delegated_action.rb
```diff
@@ -0,0 +1,52 @@
...
+          else
+            delegated_action = plan_action(klass, options)
+          end
+        when :not_available
+          raise _(_'All proxies with the required feature are unavailable at the moment')
```
adamruzicka: Ditto

app/lib/actions/proxy_action.rb
```diff
@@ -91,15 +90,31 @@ def on_data(data)
...
+      input[:proxy_action_name]
     end
 
     def proxy
       ProxyAPI::ForemanDynflow::DynflowProxy.new(:url => input[:proxy_url])
     end
 
-    def proxy_output
-      output[:proxy_output]
+    def proxy_output(life = false)
```
adamruzicka: `life` as in `live_output`?

adamruzicka:
Gemspec for `foreman-tasks-core` is missing

iNecas:
@adamruzicka added the misisng spec

app/lib/actions/proxy_action.rb
iNecas: right, updated

app/lib/actions/helpers/with_delegated_action.rb
iNecas: Fixed

lib/foreman_tasks_core/callback_action.rb
```diff
@@ -0,0 +1,11 @@
+module ForemanTasksCore
+  class CallbackAction < ::Dynflow::Action
+    def plan(callback, data)
+      plan_self(:callback => callback, :data => data)
+    end
+
+    def run
+      SmartProxyDynflowCore::Callback::Request.send_to_foreman_tasks(input[:callback], input[:data])
```
adamruzicka: Just wondering if this is the right place for this to be in? I know this will probably be only used from within `SmartProxyDynflowCore`, but `ForemanTasks{,Core}` doesn't have any kind of dependency on SPDC.

adamruzicka:
I inlined one last comment, but I don't feel strongly about it. Either way let's wait for Mr. Jenkins' approval and get this merged

tl; dr: APJ

lib/foreman_tasks_core/callback_action.rb
iNecas: Updated, we still reference it in SharableAction, but there is not other easy way how to do that. Anyway, no reason for this class to be here

adamruzicka:
No more comments from me

iNecas:
I squashed the commits and waiting for jenkins to merge and release.

iNecas:
Jenskins is green, I'm going to release new gems so that foreman_ansible and forman_remote_execution can proceed

PR #202, opened 2016-09-20
Fixes #15283 - add indexes to make the locks request faster

iNecas:
We are searching on various lock attributes to find out if we don't
have any active task locking the resource.

https://github.com/theforeman/foreman-tasks/blob/v0.8.2/app/models/foreman_tasks/lock.rb#L62

If the amount of locks
it too high, it can lead to performance degradation.


adamruzicka:
The autogenerated index names are too long, you will need to come up with a shorter name for both of the indexes and set it explicitly

iNecas:
Tests are passing, merging

iNecas:
Merged as 4868e9d

PR #205, opened 2016-10-03
Fixes #16779 - add missing indexes

iNecas:
The support for defining index at column definition was added in Rails here
https://github.com/rails/rails/commit/9a0d35e820464f872b0340366dded639f00e19b9
We need to fix migrations that happened before Rails 4.2

### DIFF
--- a/app/models/foreman_tasks/lock.rb
+++ b/app/models/foreman_tasks/lock.rb
@@ -52,7 +52,7 @@ module ForemanTasks
 
     # returns true if it's possible to aquire this kind of lock
     def available?
-      not colliding_locks.any?
+      not colliding_locks.exists?
     end
 
     # returns a scope of the locks colliding with this one
--- a/db/migrate/20161003091412_add_missing_indexes.rb
+++ b/db/migrate/20161003091412_add_missing_indexes.rb
@@ -1,0 +1,30 @@
+class AddMissingIndexes < ActiveRecord::Migration
+  # since this is just making sure the indexes we defined before are
+  # really added we don't define the down direction here.
+  # The support for defining index at column definition was added in Rails here
+  # https://github.com/rails/rails/commit/9a0d35e820464f872b0340366dded639f00e19b9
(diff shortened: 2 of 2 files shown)


db/migrate/20161003091412_add_exclusive_lock_indexes.rb
```diff
@@ -0,0 +1,7 @@
+class AddExclusiveLockIndexes < ActiveRecord::Migration
+  def change
+    add_index(:foreman_tasks_locks, [:name, :resource_type, :resource_id, :exclusive],
+              :name => 'index_foreman_tasks_locks_all_lock_properties')
```
iNecas: In theory, it shouldn't be necessary, but I would rather give it a try on some larger scale DB before replacing the one index fully.

db/migrate/20161003091412_add_exclusive_lock_indexes.rb
```diff
@@ -0,0 +1,7 @@
...
+    add_index(:foreman_tasks_locks, :task_id)
```
adamruzicka: There's already an index on this created in `20131209122644_create_foreman_tasks_locks.rb`

iNecas: I've revisited this patch to just eliminate the missed indexes, when the migration happened on rails < 4.2

adamruzicka:
Is adding missing indexes enough? Originally in this PR you were adding an index on a group of columns
```ruby
add_index(:foreman_tasks_locks, [:name, :resource_type, :resource_id, :exclusive],
          :name => 'index_foreman_tasks_locks_all_lock_properties')
```

Is this not needed anymore?

iNecas:
I think the issue originally was no indexes around `:exclusive` were there, after adding this, I think the queries should be much more effective without need for adding the combined indexes.

iNecas:
I would start adding the combined indexes once we are sure this ones are not enough

adamruzicka:
Ok, looks good to me, ACK

PR #206, opened 2016-10-05
Fixes #12729 - Add dynflow navbar links

adamruzicka:
requires https://github.com/Dynflow/dynflow/pull/204

The only issue here would be if whole foreman was mounted on some unusual location instead of `/`

### DIFF
--- a/lib/foreman_tasks/dynflow.rb
+++ b/lib/foreman_tasks/dynflow.rb
@@ -91,6 +91,9 @@ module ForemanTasks
           end
         end
 
+        set(:custom_navigation) do
+          { "Back to Foreman" => "/#{ForemanTasks::TasksController.controller_path}" }
+        end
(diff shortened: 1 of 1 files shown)


lib/foreman_tasks/dynflow.rb
```diff
@@ -91,6 +91,9 @@ def web_console
           end
         end
 
+        set(:custom_navigation) do
+          { "Back to Foreman" => "/#{ForemanTasks::TasksController.controller_path}" }
```
iNecas: If we use don't use the Foreman name, we will avoid branding problems later, "Back to tasks" sounds equally to me, while avoiding the issues

adamruzicka: Fixed

iNecas:
Thanks @adamruzicka

PR #207, opened 2016-10-13
Refs #16899 - core helper is renamed in 1.14

ares:


ares:
@iNecas fixed to keep compatiblity as you requested, could you please take a look?

adamruzicka:
Looks good to me

iNecas:
Thanks @ares!

PR #215, opened 2016-11-16
Fixes #17361 - ensure migrations_paths are set

iNecas:


iNecas:
closed in favor of https://github.com/theforeman/foreman/pull/4027

PR #218, opened 2016-12-13
Fixes #17649 - rename 'Recurring logics' to 'Tasks schedules'

shlomizadok:


shlomizadok:
Test failure unrelated

ohadlevy:
@iNecas it looks like testing is broken in general?

iNecas:
I hoped https://github.com/theforeman/foreman-infra/pull/259 will fix the issues, but apparently, the test_plugin_pull_request in not managed by jenkins job builder, if I find some time, I will open a PR against infra to do so, but for now, I've reflected the changes to the job configuration directly

iNecas:
[test]

iNecas:
What I'm not sure about is changing just  the label vs. the model/api paths etc… IMO we should also change the hammer endpoint, otherwise it would be even more confusing than it is right now. Thoughts?

shlomizadok:
I am against changing API paths / endpoints, nor hammer commands.
I think the request was mainly a semantic menu change.

ares:
I'd like to avoid another `LookupKey` vs `SmartVariable` mismatch. Think about resource type in filters or documentation in /apidoc. I agree that the name is not good but I think it should be consistent everywhere.

adamruzicka:
+1 for keeping it consistent

adamruzicka:
What's the status here?

ares:
Is there a chance someone will take over this or should it be closed?

adamruzicka:
>  ares commented on May 16, 2017
> Is there a chance someone will take over this or should it be closed?

Sounds like no to me, closing for now.

PR #220, opened 2017-01-02
Fixed gettext rake task for plugins

lzap:
When extracting strings for plugins via

    be rake plugin:gettext[foreman_discovery]

the extension in tasks creates irrelevant file in

    foreman_discovery/locale/action_names.rb

which is then extracted. This avoids that.


adamruzicka:
@lzap Needs a rebase

lzap:
Rebased.

lib/tasks/gettext.rake
```diff
@@ -7,16 +7,13 @@ gettext_find_task = begin
...
-                 .all_action_names
-                 .uniq
-                 .map { |n| %[_("#{n}")] }
-                 .join("\n")
+      # not relevant when extracting plugins
+      if text_domain == "foreman"
+        storage_file = "#{locale_path}/action_names.rb"
+        puts "writing action translations to: #{storage_file}"
+        File.write storage_file, "# Autogenerated!\n" +
+          Actions::EntryAction. all_action_names. uniq. map { |n| %[_("#{n}")] }. join("\n")
```
adamruzicka:
```
Offenses:

lib/tasks/gettext.rake:15:11: C: Align the operands of an expression spanning multiple lines.
          Actions::EntryAction. all_action_names. uniq. map { |n| %[_("#{n}")] }. join("\n")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

lzap:
Ok bot try now.

adamruzicka:
[test]

lib/tasks/gettext.rake
```diff
@@ -7,16 +7,13 @@ gettext_find_task = begin
...
-      puts "writing action translations to: #{storage_file}"
-
-      File.write storage_file,
-                 "# Autogenerated!\n" +
-                 Actions::EntryAction
-                 .all_action_names
-                 .uniq
-                 .map { |n| %[_("#{n}")] }
-                 .join("\n")
+      # not relevant when extracting plugins
```
iNecas: This is not valid: it's relevant for plugins, that define actions for foreman-tasks (see https://github.com/Katello/katello/blob/master/locale/action_names.rb),

Better approach would be first introduce a mechanism that plugins that use the actions would use for extracting just names of their actions, and then removing this code completely. Without doing this, this PR just introduces regression IMO

PR #223, opened 2017-01-03
Fixes #17914 - Improve index performance

tbrisker:
This commit addresses a couple of inefficiencies in the controllers.
the most siginificant being a `SELECT DISTINCT foreman_tasks_tasks.*`
that was added on the resource scopes to prevent duplicates when
searching on the lock properties. The duplicates were caused by an
unneeded ext_method that created a join on the lock table, instead of
using the built in scoped search capabilities to handle has many
relations.
Another inefficiency was including id fields in non-explicit searches
making those also result in more complex queries.
In addition, the scoped search declerations were updated to use the new
:relation property instead of :in.

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -90,7 +90,6 @@ module ForemanTasks
       def bulk_resume
         scope = resource_scope
         scope = scope.search_for(params[:search]) if params[:search]
-        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
         if params[:search].nil? && params[:task_ids].nil?
           scope = scope.where(:state => :paused)
           scope = scope.where(:result => :error)
@@ -131,7 +130,7 @@ module ForemanTasks
         param :order, String, :desc => N_("How to order the sorted results (e.g. ASC for ascending)")
       end
       def index
-        scope =resource_scope.search_for(params[:search]).select('DISTINCT foreman_tasks_tasks.*')
+        scope = resource_scope.search_for(params[:search])
         total = scope.count
 
         ordering_params =  {
@@ -178,7 +177,7 @@ module ForemanTasks
       private
 
       def search_tasks(search_params)
-        scope = resource_scope_for_index.select('DISTINCT foreman_tasks_tasks.*')
+        scope = resource_scope_for_index
         scope = ordering_scope(scope, search_params)
         scope = search_scope(scope, search_params)
         scope = active_scope(scope, search_params)
--- a/app/controllers/foreman_tasks/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/tasks_controller.rb
@@ -103,8 +103,7 @@ module ForemanTasks
 
     def filter(scope)
       scope.search_for(params[:search], :order => params[:order]).
-          paginate(:page => params[:page]).select('DISTINCT foreman_tasks_tasks.*')
+          paginate(:page => params[:page])
     end
-
   end
 end
--- a/app/models/foreman_tasks/recurring_logic.rb
+++ b/app/models/foreman_tasks/recurring_logic.rb
@@ -15,7 +15,7 @@ module ForemanTasks
       has_many :task_groups, -> { uniq }, :through => :tasks
     end
 
-    scoped_search :on => :id, :complete_value => false
+    scoped_search :on => :id, :complete_value => false, :only_explicit => true
     scoped_search :on => :max_iteration, :complete_value => false, :rename => :iteration_limit
     scoped_search :on => :iteration, :complete_value => false
     scoped_search :on => :cron_line, :complete_value => true
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -29,20 +29,20 @@ module ForemanTasks
     end
 
 
-    scoped_search :on => :id, :complete_value => false
+    scoped_search :on => :id, :complete_value => false, :only_explicit => true
     scoped_search :on => :label, :complete_value => true
     scoped_search :on => :state, :complete_value => true
     scoped_search :on => :result, :complete_value => true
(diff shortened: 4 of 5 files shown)


iNecas:
Unfortunately it's not equivalent to the original search. 

```ruby
ForemanTasks::Task.search_for("resource_type = Host::Managed AND resource_id = 1").to_sql
```

in version before the patch generates:

```
"SELECT \"foreman_tasks_tasks\".* FROM \"foreman_tasks_tasks\" INNER JOIN \"foreman_tasks_locks\" ON \"foreman_tasks_locks\".\"task_id\" = \"foreman_tasks_tasks\".\"id\" WHERE (((foreman_tasks_locks.\"resource_type\" = 'Host::Managed') AND (foreman_tasks_locks.\"resource_id\" = '1')))"
```

while the new version generates:

```
SELECT "foreman_tasks_tasks"."id" AS t0_r0, "foreman_tasks_tasks"."type" AS t0_r1, "foreman_tasks_tasks"."label" AS t0_r2, "foreman_tasks_tasks"."started_at" AS t0_r3, "foreman_tasks_tasks"."ended_at" AS t0_r4, "foreman_tasks_tasks"."state" AS t0_r5, "foreman_tasks_tasks"."result" AS t0_r6, "foreman_tasks_tasks"."external_id" AS t0_r7, "foreman_tasks_tasks"."parent_task_id" AS t0_r8, "foreman_tasks_tasks"."start_at" AS t0_r9, "foreman_tasks_tasks"."start_before" AS t0_r10, "foreman_tasks_locks"."id" AS t1_r0, "foreman_tasks_locks"."task_id" AS t1_r1, "foreman_tasks_locks"."name" AS t1_r2, "foreman_tasks_locks"."resource_type" AS t1_r3, "foreman_tasks_locks"."resource_id" AS t1_r4, "foreman_tasks_locks"."exclusive" AS t1_r5 FROM "foreman_tasks_tasks" LEFT OUTER JOIN "foreman_tasks_locks" ON "foreman_tasks_locks"."task_id" = "foreman_tasks_tasks"."id" WHERE ((("foreman_tasks_tasks"."id" IN (SELECT "task_id" FROM "foreman_tasks_locks" WHERE "foreman_tasks_locks"."resource_type" = 'Host::Managed' )) AND ("foreman_tasks_tasks"."id" IN (SELECT "task_id" FROM "foreman_tasks_locks" WHERE "foreman_tasks_locks"."resource_id" = '1' ))))
```

The core difference is that with the new approach, you can't find tasks that belong to resource of specific type and id - it leads to or instead.

Btw. as far as I understand the behaviour @ohadlevy seen with performance, it seems that was an issue with Mysql, and using `uniq` instead of `select('distinct')` worked for him (while on postgres this two approaches generate the very same query)

tbrisker:
@iNecas I'm not sure I follow, the issue was that the added `select distinct "foreman_tasks_tasks".*` query needed to get rid of the duplicates generated by the original query caused an extra table being created (even if using `.uniq` instead of writing the sql by hand). Also, looking at the query from the new version it does AND, not OR, only it doesn't return duplicates afaict. The extra left outer join there is unneeded and I have opened a PR against scoped search to remove it.

tbrisker:
@iNecas I have updated the test to include the example you gave as well.

iNecas:
The problem is, in the new query, the resource_type and resource_id are not in the same condition.

So if you're searching for `resource_type = User and resource_id = 1` , in your case, it would not find all the tasks that belong to the user with id 1, but it would find all tasks with resource of type User and has a resource of *any* type with id 1. So in the tests, if the task is linked to another resource through the lock (such as host), and the `host.id` is the same as the `user.id`, it would include it in the result as well, which is not how it should work.

I agree i would not expect difference between using `uniq` and `distinct` in the query, but @ohadlevy reported it had impact.

tbrisker:
@iNecas Updated the test with that case, seems to still work fine. IIUC, the difference @ohadlevy was reporting with `.uniq` instead of `DISTINCT` is that it didn't completely break on mysql because the sql was generated correctly.

iNecas:
I'm not aware of the fact the task index would be broken completely on mysql.

Here is the example it beaks on https://gist.github.com/iNecas/a92ce9507f7c579fbb46b0a117a63b9b

iNecas:
Btw. I've gave it a try on mysql, and I have not seen any difference in sql or explain response with `DISTINCT` vs. `uniq` so I'm not sure what Ohad actually  saw

adamruzicka:
@iNecas, @tbrisker what's the status here?

iNecas:
Putting on waiting on contributor, as it's not equivalent on how it worked before and failing on this case: https://gist.github.com/iNecas/a92ce9507f7c579fbb46b0a117a63b9b

adamruzicka:
@tbrisker Needs a rebase

tbrisker:
@iNecas @adamruzicka thanks, i'm still trying to think of a solution that will correctly handle the case outlined by @iNecas - the problem is we were relying on a side effect of the problematic ext_method to correctly handle searching the polymorphic association in the locks table. 
In general, scoped_search treats each clause as a condition on the searched object (task in this case), so that the result is correct in that case with respect to the task - it does have a lock with that resource id AND a lock with that resource type - just not the same lock. 
I don't know how common that search is, but maybe it should be changed to have a specific search for each resource type (or at least the main one). That would allow searching by "hostname=example.com" instead of "resource_type = host and resource_id = wait_let_me_go_figure_out_what_was_the_id_of_example.com_again"

iNecas:
Just keep in mind we need this also for searching by repository, content view etc.

adamruzicka:
@tbrisker what's the status here?

iNecas:
How would that solve the performance issue?

ShimShtein:
@iNecas, It won't solve the performance issue, but it should solve the search from your gist.

The performance shouldn't be very bad too, since it will be converted to two simple conditions (by custom method).

iNecas:
The gist is broken just in this PR, not in the master

iNecas:
Closing due to inactivity

PR #219, opened 2016-12-21
fixes #17816 - adds rubocop support

ohadlevy:
This commit adds rubocop support and update the code base
as well for trivial fixes, there are no functional changes to the
codebase.

additional cops might be enabled further on in the .rubocop_todo.yml file

PR #224, opened 2017-01-11
Fixes #17816 - adds rubocop support

adamruzicka:
based on @ohadlevy 's #219


adamruzicka:
[test]

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -104,28 +94,27 @@ def get_humanized(method)
...
-               :state => data[:state].to_s,
-               :result => data[:result].to_s)
+      new(:external_id => execution_plan_id,
+          :state => data[:state].to_s,
+          :result => data[:result].to_s)
     end
 
-    private
-
+    private_class_method :model_name
```
iNecas: I'm not sure we want this method to actually be private? Based on the history, seems this should be used by external code (routes generation)

iNecas: I don't see this methods being private in foreman

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -51,13 +50,11 @@ DESC
...
       def template(filename)
-        File.join(Gem::Specification.find_by_name("dynflow").gem_dir, 'web', 'views', "#{filename}.erb")
+        File.join(Gem::Specification.find_by(name: 'dynflow').gem_dir, 'web', 'views', "#{filename}.erb")
       end
 
       def erb(file, options = {})
-        unless @cache[file]
-          @cache[file] = Tilt.new(template(file))
-        end
+        @cache[file] = Tilt.new(template(file)) unless @cache[file]
```
iNecas: Does rubocop have anything againts `@cache[file] ||= Tilt.new(template(file))`?

app/models/foreman_tasks/task/dynflow_task.rb
adamruzicka: So we can drop the `private` completely, right?

lib/foreman_tasks/tasks/export_tasks.rake
adamruzicka: Apparently not

app/lib/actions/proxy_action.rb
```diff
@@ -17,8 +17,8 @@ def plan(proxy, klass, options)
       plan_self(options.merge(:proxy_url => proxy.url, :proxy_action_name => klass.to_s))
     end
 
-    def run(event = nil)
-      with_connection_error_handling(event) do |event|
+    def run(an_event = nil)
```
adamruzicka: Disabled  the cop enforcing this altogether

app/models/foreman_tasks/task/dynflow_task.rb
iNecas: yes

iNecas:
Waiting for jenkins.

adamruzicka:
Jenkins is green, merging. Thanks for reviewing @iNecas !

PR #225, opened 2017-01-20
Fixes #18189 - Round completed tasks percentage

johnpmitsch:
The task percentage is excessively long and can be
rounded to a whole number.


ares:
Why not using `round`? 99.1 will results in 100% :-) but I don't insist, it's quite edge case...

johnpmitsch:
@ares thats a fair point, I assume the progress will reach an actual 100, so no need to round up. updated!

ares:
:+1:

johnpmitsch:
@ares what is the status of this PR?

iNecas:
@johnpmitsch it's merged :)

PR #226, opened 2017-01-24
Fixes #18060 - ship default GC tuning

jlsherrill:


jlsherrill:
These likely could be tuned a bit more, but these defaults have shown a great improvement across multiple users for reducing the scale and speed of memory growth.

iNecas:
@jlsherrill do you hae some data measured with/without this tunings? What actions were run by that time?

jlsherrill:
Here are my memory consumption results with and without GC from my tests a month or so ago testing Content view publish with capsule syncs (10 capsules):

graph

I didn't do nearly as stringent timing tests with and without the GC tuning, but the couple of tests i did do showed possibly a 1-3% decrease in performance although it wasn't entirely an apples to apples comparison as CV publish may just be getting 1-3% slower as mongo and postgresql have more data.  Let me know if there is more data you'd like me to gather.

iNecas:
Looks nice: let me try this out in few other examples before merging: I expect to have it finalized by the end of the week

iNecas:
Thanks @jlsherrill, pulling this in.

PR #227, opened 2017-02-02
Fixes #18382 - Save action prior to save/destroy on action triggering 

jlsherrill:


thomasmckay:
This change fixes the problem of changes not being saved to candlepin

app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -71,17 +71,17 @@ def sync_action!
       end
 
       def plan_create_action
```
iNecas: since now all the methods (plan_{create,update,destroy}_action) look the same, i guess we have just one `plan_hook_action` and use it for all `after_create`, `after_update` and `after_destroy`

jlsherrill:
@iNecas where did you want me to make a comment saying that they are called before save/destroy happens?  I could add it to just action_triggering.rb, but maybe explaining action triggering in the README and adding it there might be better?

iNecas:
Above the section, where `create_action`, `update_action`, `delete_action` are being defined

app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -116,17 +106,18 @@ def dynflow_task_wrap(method)
...
                  when :destroy
                    destroy_action
                  else
                    raise 'unexpected method'
                  end
-        ensure_not_in_transaction! if action
+        ensure_not_in_transaction! if @_dynflow_hook_action
         yield.tap do |result|
           execute_planned_action if result
+          @_dynflow_hook_action = nil
```
iNecas: The reset should go to ensure block, to make sure w always do this

jlsherrill: 👍

jlsherrill:
@iNecas updated

iNecas:
Thanks @jlsherrill

PR #228, opened 2017-02-03
Fixes #18385 - fix the exporting task

ares:


iNecas:
Thanks, I've disabled rubocop checks on those lines, to make it passing, and merged as 13ee398

PR #229, opened 2017-02-09
Fixes #17528 - Use batch processing for bulk actions

adamruzicka:
requires https://github.com/Dynflow/dynflow/pull/214

### DIFF
--- a/app/lib/actions/action_with_sub_plans.rb
+++ b/app/lib/actions/action_with_sub_plans.rb
@@ -3,6 +3,7 @@ module Actions
     middleware.use Actions::Middleware::KeepCurrentUser
 
     include Dynflow::Action::WithSubPlans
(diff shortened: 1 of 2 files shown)


app/lib/actions/action_with_sub_plans.rb
```diff
@@ -3,6 +3,7 @@ class Actions::ActionWithSubPlans < Actions::EntryAction
     middleware.use Actions::Middleware::KeepCurrentUser
 
     include Dynflow::Action::WithSubPlans
+    include Dynflow::Action::WithBulkSubPlans
```
iNecas: I don't think we should include this into the `ActionWithSubPlans` but rather into `BulkAction` and `RunHostsJob` action: the action with sub plans doesn't have to be bulk at the end

adamruzicka:
Closing in favor of #236

PR #230, opened 2017-02-10
Fixes #18456 - Fix scoped_search deprecations

dLobatog:
The scoped_search :in relations are deprecated and :relation should be
used instead to prevent it breaking when they completely drop the
feature.


iNecas:
Thanks @dLobatog

PR #233, opened 2017-02-17
Fixes #18552 - load erros only if value is present

ares:


app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -152,7 +152,7 @@ def execute_planned_action
           run = ::ForemanTasks.dynflow.world.execute(@execution_plan.id)
           if @dynflow_sync_action
             run.wait
-            if run.value.error?
+            if run.value.try(:error?)
```
iNecas: I wonder if we shouldn't check the reason and raise another error when that happens http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Promises/Future.html#reason-instance_method

When one can get into this state? When the tasks is not running?

adamruzicka: ping @ares

ares: note for myself - this only happens when I do image based provisioning and I receive facts from chef-client

adamruzicka:
@ares What's the status here?

adamruzicka:
@ares ping

ares:
Sadly, I don't have the reproducer anymore, most likely the issue is between orchestration and transaction in case of image based provisioning, however I didn't see it anywhere else, let's close for now.

PR #235, opened 2017-02-24
Fixes #18673 - Remove initialization code in favor of Dynflow

dLobatog:
Initialization code for Rails has moved to Dynflow on 0.8.20. The only
things that should remain in Tasks with regards to that are - the web
console permissions setup and the persistence adaptor for Foreman.

The loggers, task and script to start the executor also moved to
Foreman, so the plugin should merely use them.

This relies on https://github.com/theforeman/foreman/pull/4316 - however, I'm finding a small issue:

I found that the persistence adapter saves these classes in the database ('Dynflow::ActiveJob::QueueAdapters::JobWrapper') - when I run a Job using ActiveJob, this label will be set. No name or other fields are set, so I decided to not show those jobs in the /tasks page.

Since these jobs are supposed to be small things for checking notifications and stuff like that, it's probably fine to not have a task for them. However, I would want to have that at some point. How can I retrieve attributes from the job on `on_execution_plan_save`?


lib/foreman_tasks/dynflow/persistence.rb
```diff
@@ -11,7 +11,8 @@ def save_execution_plan(execution_plan_id, value)
         begin
           on_execution_plan_save(execution_plan_id, value)
```
dLobatog: Here, at this point, I would like to be able to access properties from the job, such as name, comments, etc.. to actually save ActiveJob jobs with a proper name in the Tasks table. I was not able to do this, and I'm not sure if it's possible without changing Dynflow cc @iNecas 

My idea was to have each job provide an arbitrary hash with things like name, a comment explaining what the job does, possible debugging steps, etcetera, but I cannot pick random stuff from the job here.

If such an interface existed, we would be able to show nicer names on the /tasks table than the class name, e.g: "Pulp::Repository::Sync" could show up as "Pulp: Syncing repository EPEL", and a comment explaining what the task does, etc..

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -102,8 +102,9 @@ def find_dynflow_task
     end
 
     def filter(scope)
-      scope.search_for(params[:search], :order => params[:order])
-           .paginate(:page => params[:page]).select('DISTINCT foreman_tasks_tasks.*')
+      scope.where.not(:label => 'Dynflow::ActiveJob::QueueAdapters::JobWrapper').
+        search_for(params[:search], :order => params[:order]).
+        paginate(:page => params[:page]).select('DISTINCT foreman_tasks_tasks.*')
```
ShimShtein: Does it make sense to use `distinct` method instead?
``` ruby
paginate(:page => params[:page]).distinct()
```

bin/dynflow-executor
```diff
@@ -1,46 +0,0 @@
...
-BANNER
-
-  opts.on('-h', '--help', 'Show this message') do
-    puts opts
-    exit 1
-  end
-  opts.on('-f', '--foreman-root=PATH', "Path to Foreman Rails root path. By default '#{options[:foreman_root]}'") do |path|
-    options[:foreman_root] = path
-  end
-  opts.on('-c', '--executors-count=COUNT', 'Number of parallel executors to spawn. Overrides EXECUTORS_COUNT environment varaible.') do |count|
```
ShimShtein: Where did this code go? Couldn't find it in neither foreman nor dynflow commits?

lib/foreman_tasks/dynflow/persistence.rb
ShimShtein: I was thinking in this direction too, but from a different angle - I would like to customize the parameters that are printed for each task - the parameters could be too long, or contain sensitive information.
@iNecas, maybe we need a `#humanize` method in the dynflow task.

bin/dynflow-executor
dLobatog: Nowhere, I asked on IRC a few people and apparently it's not being used anywhere these days

ShimShtein: This means we are ditching command line options for dynflow executor. We will still be able to set various options through environment variables and rake task, but we won't get nice help for it.
@iNecas, is it acceptable?

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -102,8 +102,8 @@ def find_dynflow_task
     end
 
     def filter(scope)
-      scope.search_for(params[:search], :order => params[:order])
-           .paginate(:page => params[:page]).select('DISTINCT foreman_tasks_tasks.*')
+      scope.search_for(params[:search], :order => params[:order]).
+        paginate(:page => params[:page]).distinct
```
dLobatog:
```
[3] pry(#<ForemanTasks::TasksController>)> scope.search_for(params[:search], :order => params[:order]).paginate(:page => params[:page]).select('DISTINCT foreman_tasks_tasks.*').to_sql
=> "SELECT  DISTINCT foreman_tasks_tasks.* FROM \"foreman_tasks_tasks\"  ORDER BY \"foreman_tasks_tasks\".\"started_at\" DESC NULLS LAST  LIMIT 20 OFFSET 0"
[4] pry(#<ForemanTasks::TasksController>)> scope.search_for(params[:search], :order => params[:order]).paginate(:page => params[:page]).distinct().to_sql
=> "SELECT  DISTINCT \"foreman_tasks_tasks\".* FROM \"foreman_tasks_tasks\"  ORDER BY \"foreman_tasks_tasks\".\"started_at\" DESC NULLS LAST  LIMIT 20 OFFSET 0"
```

bin/dynflow-executor
dLobatog: AFAIK, we're not ditching anything, this code is plain and simply not used (not because of any changes I made, it wasn't used before)

lib/foreman_tasks/dynflow/persistence.rb
dLobatog: https://github.com/theforeman/foreman-tasks/pull/235/files#diff-a00196c0418a614248a22c431ef0d163R22 - after changing some stuff in the helper, to support `def humanize_name` etc.. methods in the job, I got it to show all the information just fine

ShimShtein:
:+1:

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -41,12 +41,18 @@ def execution_plan
       @execution_plan ||= ForemanTasks.dynflow.world.persistence.load_execution_plan(external_id)
     end
 
+    def label
```
iNecas: Could you put this logic into separate PR. Btw. I'm just looking into this code from performance perspective: I might try to reduce the the calls to laod teh execution plan from here.

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -70,7 +76,27 @@ def cli_example
 
     def main_action
       return @main_action if @main_action
-      execution_plan.root_plan_step.action execution_plan
+      action = execution_plan_action
+      if activejob?
```
adamruzicka: Because of this the result is sometimes an `Action` and sometimes an `ActiveJob`. This can lead to errors in [here](https://github.com/dLobatog/foreman-tasks/blob/d82379fed94983d3c88144361ac2d20faba4b3c2/app/models/foreman_tasks/task/dynflow_task.rb#L18), where `ActiveJob` doesn't respond to `caller_execution_plan_id` and raises `NoMethodError`. I hit this several times today

app/models/foreman_tasks/task/dynflow_task.rb
ShimShtein: Why putting it in a separate PR? It's adaptation of tasks to ActiveJob.

dLobatog: I think it's fine - this one can just remove the initialization code, and the other one includes the 'UI' adaption part.

foreman-tasks.gemspec
```diff
@@ -29,7 +29,7 @@ DESC
   s.extra_rdoc_files = Dir['README*', 'LICENSE']
 
   s.add_dependency "foreman-tasks-core"
-  s.add_dependency "dynflow", '~> 0.8.17'
+  s.add_dependency "dynflow", '~> 0.8.21'
```
adamruzicka: Is there another Dynflow release coming soon or should it be '0.8.20'?

dLobatog: Correct 😄

dLobatog:
Latest update removes the `deploy/` directory. If the foreman-tasks bin dir is deployed by Foreman, it needs the systemd config too. https://github.com/theforeman/foreman-packaging/pull/1549 deploys it.

adamruzicka:
[test]

adamruzicka:
@dLobatog needs a rebase

dLobatog:
Updated so that it takes into account the improvements made by @ShimShtein , which includes https://github.com/Dynflow/dynflow/pull/236

I have yet to test this thoroughly, but feel free to comment in the meantime 😄

dLobatog:
Rebased

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -102,8 +102,8 @@ def find_dynflow_task
     end
 
     def filter(scope)
-      scope.search_for(params[:search], :order => params[:order])
-           .paginate(:page => params[:page]).select('DISTINCT foreman_tasks_tasks.*')
+      scope.search_for(params[:search], :order => params[:order]).
```
adamruzicka: Rubocop will probably bug you about this when the tests run
```
app/controllers/foreman_tasks/tasks_controller.rb:105:66: C: Place the . on the next line, together with the method name.
      scope.search_for(params[:search], :order => params[:order]).
                                                                 ^
```

lib/foreman_tasks/dynflow/configuration.rb
```diff
@@ -1,146 +1,11 @@
...
-
-    attr_accessor :lazy_initialization
-
-    # what rake tasks should run their own executor, not depending on the external one
-    attr_accessor :rake_tasks_with_executor
-
-    # if true, the ForemanTasks::Concerns::ActionTriggering will make
-    # no effect. Useful for testing, where we mignt not want to execute
-    # the orchestration tied to the models.
-    attr_accessor :disable_active_record_actions
```
adamruzicka: When running the tests locally it fails with 
```ruby
foreman-tasks/test/foreman_tasks_test_helper.rb:11:in `<top (required)>': undefined method `disable_active_record_actions=' for #<ForemanTasks::Dynflow::Configuration:0x005580871810c0> (NoMethodError)
```

it needs to be added to the configuration in dynflow or foreman

foreman-tasks.gemspec
```diff
@@ -32,7 +32,6 @@ same resource. It also optionally provides Dynflow infrastructure for using it f
   s.add_dependency "dynflow", '~> 0.8.24'
   s.add_dependency "sequel" # for Dynflow process persistence
```
dLobatog: Nope, it's automatically added to the Gemfile.lock.. same with sinatra

adamruzicka:
The tests are indeed passing. If we agree on keeping the Dynflow tests here I'm ready to ack this

iNecas:
[test]

adamruzicka:
We can't really argue with the tests, can we? :) +1 from me. @iNecas anything else to add from your point of view?

iNecas:
Thanks @dLobatog and @adamruzicka for huge amount of work at this

PR #236, opened 2017-02-27
Fixes #18963 - Use batch processing

adamruzicka:


app/models/foreman_tasks/task.rb
```diff
@@ -158,6 +158,30 @@ def add_missing_task_groups(groups)
...
+      else
+        cancelled = 0
+        failed = unsuccessful.count
+      end
+      {
+        :cancelled => cancelled,
+        :failed    => failed,
+        :pending   => pending,
+        :success   => success,
+        :total     => cancelled + failed + pending + success
```
iNecas: This number doesn't match for bulk actions with a batch. I wonder if we should not move it to the bulk action helper, probably rebasing on https://github.com/theforeman/foreman-tasks/pull/229/files

app/models/foreman_tasks/task.rb
```diff
@@ -158,6 +158,30 @@ def add_missing_task_groups(groups)
       (groups - task_groups).each { |group| task_groups << group }
     end
 
+    def cancelled?
+      execution_plan.errors.map(&:exception).any? { |exception| exception.class == ::ForemanTasks::Task::TaskCancelledException }
```
iNecas: I'm a bit concerned about the `cancelled` calculation: I would suggest removing this from here, and perhaps introducing different state when we want to query based on this. I suspect this being a potential reason to this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1426204

adamruzicka:
Rebased onto https://github.com/theforeman/foreman-tasks/pull/229

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -117,5 +117,16 @@ def self.new_for_execution_plan(execution_plan_id, data)
     def self.model_name
       superclass.model_name
     end
+
+    private
+
+    def map_result(result)
+      result = :cancelled if result == :error && cancelled?
```
iNecas: 👍

app/models/foreman_tasks/task.rb
```diff
@@ -158,6 +158,25 @@ def add_missing_task_groups(groups)
       (groups - task_groups).each { |group| task_groups << group }
     end
 
+    def sub_tasks_counts
+      success   = sub_tasks.where(:result => 'success').count
+      pending   = sub_tasks.where(:result => 'pending').count
+      failed    = sub_tasks.where(:result => %w(error warning)).count
+      cancelled = sub_tasks.where(:result => 'cancelled').count
```
iNecas: This is not working fully with batch planning of subtasks: the chart in rex looks like window's file copy dialog https://xkcd.com/612/ :). We need to include:

1. the count of unplanned tasks in in :pending when the task is not stopped
2. the count of unplanned tasks in :cancelled, if the task is stopped

app/models/foreman_tasks/task.rb
```diff
@@ -158,6 +158,20 @@ def add_missing_task_groups(groups)
       (groups - task_groups).each { |group| task_groups << group }
     end
 
+    def sub_tasks_counts
+      result = %w(cancelled error pending success warning).zip([0].cycle).to_h
+      result.update sub_tasks.group(:result).count
+      sum = result.values.reduce(:+)
+      if respond_to?(:main_action) && main_action.respond_to?(:total_count)
+        result[:total] = main_action.total_count
+        key = state == 'stopped' ? 'cancelled' : 'pending'
```
iNecas: While we know what's going on here right now, I bet we will not be that clever few months later: please add an explanatory comment why we are doing this

adamruzicka: Done

iNecas:
It's ok now, thanks @adamruzicka

PR #238, opened 2017-03-01
Fixes #18758 - Allow ActiveJobs to be displayed in the tasks lists

dLobatog:
The jobs created by ActiveJob would not show up properly in the tasks
list. Small adaptions should be made to allow foreman-tasks to pick up
the 'humanized' methods and display it.

Can be merged at any point, before or after: https://github.com/theforeman/foreman-tasks/pull/235

### DIFF
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -15,11 +15,11 @@ module ForemanTasks
       self.start_at       = string_to_time(utc_zone, data[:start_at]) if data[:start_at]
       self.start_before   = string_to_time(utc_zone, data[:start_before]) if data[:start_before]
       self.parent_task_id ||= begin
-                                if main_action.caller_execution_plan_id
+                                if main_action.present? && main_action.caller_execution_plan_id
                                   DynflowTask.where(:external_id => main_action.caller_execution_plan_id).first!.id
                                 end
                               end
-      self.label          ||= main_action && main_action.class.name
+      self.label          ||= label
       changes = self.changes
       save!
       changes
@@ -69,12 +69,19 @@ module ForemanTasks
       execution_plan(false)
     end
 
+    def label
+      return execution_plan_action.input['job_class'] if active_job?
+      return main_action.class.name if main_action.present?
+      self[:label]
+    end
+
     def input
-      main_action.respond_to?(:task_input) && main_action.task_input
+      return execution_plan_action.input['job_arguments'] if active_job?
+      main_action.task_input if main_action.respond_to?(:task_input)
     end
 
     def output
-      main_action.respond_to?(:task_output) && main_action.task_output
+      main_action.task_output if main_action.respond_to?(:task_output)
     end
 
     def failed_steps
@@ -98,17 +105,47 @@ module ForemanTasks
 
     def main_action
       return @main_action if defined?(@main_action)
-      @main_action = execution_plan && execution_plan.root_plan_step.try(:action, execution_plan)
+      if active_job?
+        action = execution_plan_action
+        active_job_action(action.input['job_class'], action.input['job_arguments'])
+      else
+        @main_action = execution_plan && execution_plan.root_plan_step.try(:action, execution_plan)
+      end
+    end
+
+    # The class for ActiveJob jobs in Dynflow, JobWrapper is not expected to
+    # implement any humanized actions. Individual jobs are expected to implement
+    # humanized_* methods for foreman-tasks integration.
+    def active_job_action(klass, args)
+      return if klass.blank?
+      klass.constantize.new(*args)
+    end
+
+    def active_job?
+      return false unless execution_plan.present? &&
+                          execution_plan.root_plan_step.present?
+      execution_plan_action.class == ::Dynflow::ActiveJob::QueueAdapters::JobWrapper
+    end
+
+    def execution_plan_action
+      execution_plan.root_plan_step.action(execution_plan)
+    end
+
+    def execution_scheduled?
+      main_action.nil? || main_action.respond_to?(:execution_plan) &&
+        main_action.execution_plan.state == :scheduled ||
+        execution_plan.state == :scheduled
     end
 
     def get_humanized(method)
       @humanized_cache ||= {}
-      if [:name, :input, :output, :error].include?(method)
-        method = "humanized_#{method}".to_sym
-      end
+      method = find_humanize_method_kind(method)
       Match! method, :humanized_name, :humanized_input, :humanized_output, :humanized_errors
-      return N_('N/A') if method != :humanized_name && (main_action.nil? || main_action.execution_plan.state == :scheduled)
-      return N_(label) if method == :humanized_name && main_action.nil?
+      if method != :humanized_name && execution_scheduled?
+        return N_('N/A')
+      elsif method == :humanized_name && main_action.nil?
+        return N_(label)
+      end
       @humanized_cache[method] ||= begin
                                      if main_action.respond_to? method
                                        begin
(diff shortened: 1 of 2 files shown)


ares:
not a full review, just one comment from the quick scan, should we use active_job instead of activejob everywhere? Just to avoid another operatingsystem and the fact, methods like `underscore` or `constantize` would not work

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -70,7 +76,27 @@ def cli_example
 
     def main_action
       return @main_action if @main_action
-      execution_plan.root_plan_step.action execution_plan
+      action = execution_plan_action
+      if activejob?
```
adamruzicka: I'm still hitting [this](https://p.teknik.io/X0Wv8)

adamruzicka:
Does this include #235 now?

dLobatog:
@adamruzicka my mistake, the point is that without the other PR this one won't work at all though

iNecas:
[test]

iNecas:
One last unit test failure http://ci.theforeman.org/job/test_plugin_pull_request/3940/#showFailuresLink

iNecas:
besides the last test failure, it requires rebase

adamruzicka:
@dLobatog ping

dLobatog:
@adamruzicka I'm currently rebasing this - will ping you once it works (page shows the class name of the job or humanized_name if defined in the job)

ares:
@dLobatog the failures seems to be related

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -97,16 +103,38 @@ def cli_example
 
     def main_action
       return @main_action if defined?(@main_action)
-      @main_action = execution_plan && execution_plan.root_plan_step.try(:action, execution_plan)
+      if active_job?
+        action = execution_plan_action
+        active_job_action(action.input['job_class'], action.input['job_arguments'])
```
adamruzicka: If the action is `:scheduled` or `:planning`, the input is still unset, leading to 
```
2018-01-22T13:11:17  [dynflow] [W] Error on on_execution_plan_save event
 | NoMethodError: undefined method `constantize' for nil:NilClass
 | /home/aruzicka/git/foreman-tasks/app/models/foreman_tasks/task/dynflow_task.rb:118:in `active_job_action'
 | /home/aruzicka/git/foreman-tasks/app/models/foreman_tasks/task/dynflow_task.rb:108:in `main_action'
```

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -97,16 +103,38 @@ def cli_example
 
     def main_action
```
adamruzicka: This method sometimes returns an `Action` and sometimes the class of the active job. In either case we try to call `caller_execution_plan_id` on it here https://github.com/theforeman/foreman-tasks/pull/238/files#diff-a00196c0418a614248a22c431ef0d163R17 . Either all our ActiveJobs should inherit from a single class which would basically just stub this method and return `nil`, or preferably make the `main_action` always return the action and create another method for obtaining the job's class

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -97,16 +103,38 @@ def cli_example
...
+      execution_plan_action.class == ::Dynflow::ActiveJob::QueueAdapters::JobWrapper
+    end
+
+    def execution_plan_action
+      execution_plan.root_plan_step.action(execution_plan)
     end
 
     def get_humanized(method)
       @humanized_cache ||= {}
-      if [:name, :input, :output, :error].include?(method)
```
adamruzicka: When called from task's details, the `method` variable is already `:humanized_$something`. The way it was before this part only changed the variable if the `method` was not already in the `humanized_$something` shape. Your variant changes it every time (sometimes setting it to nil) and thus breaking task details for all tasks

```
2018-01-23T10:09:50 b3f6d069 [app] [F] ActionView::Template::Error (Value (NilClass) '' is not matching any of: humanized_name; humanized_input; humanized_output; humanized_errors.):
2018-01-23T10:09:50 b3f6d069 [app] [F]     173:   </div>
 |     174: </div>
 |     175: 
 |     176: <% unless @task.humanized[:output].blank? %>
 |     177: <div>
 |     178:   <span class="param-name list-group-item-heading"><%= _("Output") %>:</span>
 |     179:   <span class="param-value">
2018-01-23T10:09:50 b3f6d069 [app] [F]   
2018-01-23T10:09:50 b3f6d069 [app] [F] /home/aruzicka/git/foreman-tasks/app/models/foreman_tasks/task/dynflow_task.rb:136:in `get_humanized'
 | /home/aruzicka/git/foreman-tasks/app/models/foreman_tasks/task/dynflow_task.rb:94:in `humanized'
 | /home/aruzicka/git/foreman-tasks/app/views/foreman_tasks/tasks/_details.html.erb:176:in `_ab3cdce398436b4e1a9f9768905bfabf'
 | /home/aruzicka/git/foreman-tasks/app/views/foreman_tasks/tasks/show.html.erb:105:in `block in _790ce4eaa756fd806681718e86a53324'
 | app/helpers/form_helper.rb:296:in `form_for'
 | /home/aruzicka/git/foreman-tasks/app/views/foreman_tasks/tasks/show.html.erb:93:in `_790ce4eaa756fd806681718e86a53324'
 | app/controllers/concerns/application_shared.rb:15:in `set_timezone'
 | app/models/concerns/foreman/thread_session.rb:32:in `clear_thread'
 | app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in `set_topbar_sweeper_controller'
 | lib/middleware/catch_json_parse_errors.rb:8:in `call'
 | lib/middleware/session_safe_logging.rb:17:in `call'
 | lib/middleware/tagged_logging.rb:18:in `call'
```

dLobatog: I fixed this here https://github.com/theforeman/foreman-tasks/pull/238/files#diff-a00196c0418a614248a22c431ef0d163R151

app/models/foreman_tasks/task/dynflow_task.rb
dLobatog: How did you get this error, for example by scheduling a REX job ?

adamruzicka: I think I scheduled an active job to be executed in the future and went to look into the tasks index before it was started

iNecas:
It seems there is still some topic unresolved. Also tests are failling, probably needs rebasing.

ares:
based on last comment, setting WoC, I see @dLobatog pushed the new version, but it's not clear, whether all was resolved

dLobatog:
Last update just fixes a cop (use `self[:attr] instead of read_attribute(:attr)`)

dLobatog:
@adamruzicka Updated to fix the `SampleJob.set(wait_until: Date.tomorrow.noon).perform_later` and the broken regex, thanks for the review!

san7ket:
Tested it with creating a job 
```
 def humanized_name
    "Create RSS notifications"
  end
```

Works as expected.

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -120,6 +157,13 @@ def get_humanized(method)
                                    end
     end
 
+    def find_humanize_method_kind(method)
+      return method if method.match?(/humanized_.*/)
```
iNecas: Tests failures seem related, At least line connects directly to this failure http://ci.theforeman.org/job/test_plugin_pull_request/4879/database=postgresql,label=fast,ruby=2.3/testReport/junit/(root)/ForemanTasks__DynflowConsoleAuthorizerTest__user%20without%20edit_foreman_tasks%20permissions/test_0001_can_not_see_any_tasks/

Most of the failures in http://ci.theforeman.org/job/test_plugin_pull_request/4879/ seemed to be related to that.

dLobatog:
@iNecas @adamruzicka Fixed the failures on 2.3. Turns out `string.match?(regex)` is a new feature of 2.4, so I switched back to `regex =~ string` :smile: hopefully that was the last thing

iNecas:
Thanks @dLobatog, @adamruzicka and @san7ket

PR #239, opened 2017-03-03
Refs #18231 - task export moved to dynflow

lzap:
Counterpart for https://github.com/Katello/katello-packaging/pull/329

### DIFF
--- a/extra/dynflow-debug.sh
+++ b/extra/dynflow-debug.sh
@@ -3,5 +3,17 @@
 # This file provides additional debug information for foreman-debug tool and is
 # symlinked as /usr/share/foreman/script/foreman-debug.d/60-dynflow_debug
 
+export_csv() {
+  echo "COPY ($1) TO STDOUT WITH CSV;" | su - postgres -c '/usr/bin/psql -d foreman' > $2
+}
+
 add_files /var/log/foreman/dynflow_executor*.log*
 add_files /var/log/foreman/dynflow_executor*.output*
+
+# Foreman Tasks fast export (for HTML version use foreman-rake foreman_tasks:export_tasks)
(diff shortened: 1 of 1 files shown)


extra/dynflow-debug.sh
```diff
@@ -3,5 +3,17 @@
...
 
+export_csv() {
+  echo "COPY ($1) TO STDOUT WITH CSV;" | su - postgres -c '/usr/bin/psql -d foreman' > $2
+}
+
 add_files /var/log/foreman/dynflow_executor*.log*
 add_files /var/log/foreman/dynflow_executor*.output*
+
+# Foreman Tasks fast export (for HTML version use foreman-rake foreman_tasks:export_tasks)
+export_csv "select dynflow_execution_plans.* from foreman_tasks_tasks join dynflow_execution_plans on (foreman_tasks_tasks.external_id = dynflow_execution_plans.uuid) where foreman_tasks_tasks.started_at > 'now'::timestamp - '${DYNFLOW_EXPORT_MONTHS:-2} months'::interval and foreman_tasks_tasks.result = 'success'" "$DIR/dynflow_execution_plans.csv"
```
iNecas: One think I've noticed as a user: why to limit the result on success: it's more useful for me to see all tasks at once, and the non-successful ones should be smaller in count, so the resulting size should not matter much

lzap: Sure, it must have been some kind of copy paste error, not sure why it was there.

I also increased default months to 6 which is saner to me, this export is quick and not much data generated, we should just prevent it does not export all in case customer have cleanup task disabled or something... Half-a year is sane I think.

iNecas:
Thanks @lzap, merging now, works as expected

PR #242, opened 2017-03-16
Fixes #18919 - Pull DynflowTask status in after_commit

mdellweg:
reinitiante update in case we missed on it

### DIFF
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -2,8 +2,14 @@ module ForemanTasks
   class Task::DynflowTask < ForemanTasks::Task
     include Algebrick::TypeCheck
 
+    after_commit :pull_update_from_dynflow
+
     delegate :cancellable?, :progress, to: :execution_plan
     scope :for_action, ->(action_class) { where(label: action_class.name) }
+
+    def pull_update_from_dynflow
+      execution_plan.save unless (state == execution_plan.state.to_s)
+    end
 
     def update_from_dynflow(data)
       utc_zone = ActiveSupport::TimeZone.new('UTC')


iNecas:
Planning additional tasks in finalize phase is a bit complicated, as there might be other issues with the task being run before the transaction of planning was committed. Could you share with us the specific case you are trying to solve? The right solution would be, either triggering the task from run, not finalize (I know it might not always be possible), the other option would be to just plan the the next task from the finalize phase, and let dynflow to trigger the execution phase once the original tasks finishes.

mdellweg:
I wanted to trigger ::Actions::Katello::Repository::Update from a models after_save hook.
The corresponding save resided in the finalize-phase of another Task (as database access in 'run' is a bad idea). Using the after_commit hook obviosly put the call outside the transaction and solved the problem there.
On the other hand, i have a hard time planing the creation of a Katello::Product with attached Katello::Repositories. The products candlepin id is not available upfront, but initiating the creation of the repositories in finalize hits the same problem as before.
Could you share a minimal example where the next task is planned in and triggered after finalization of an asynchronous task? I do not find it in the docs.

iNecas:
Could you describe more the use case in Katello's terms: I'm wondering if we could find different way to model this that would fit better into the design. The fact that the candlepin id is not known upfront should not be that much of an issue, as Dynflow can handle future references to the values, as we do here for example (https://github.com/Katello/katello/blob/028b832a2975a45d8610817536f7f456d6e97139/app/lib/actions/katello/host/register.rb#L35)

mdellweg:
Ideally I want to create a product and add repositories as part of the same action (minimal example attached).
[gen_product.rb.txt](https://github.com/theforeman/foreman-tasks/files/868056/gen_product.rb.txt)

Unfortunately, i get an error of the form
_Katello::Errors::CandlepinError:Runtime Error RESTEASY001185: Could not find resource for relative : /products//content/1490363035471_
Indicating, that the Candlepin ID between 'product' and 'content' is missing.
I did not find a way to propagate that information to the right place, and initiating the repository creation in finalize did not work properly either.
Any ideas to solve this are welcome.

iNecas:
One way to achieve this would be via sub-plans, where you actually drive the sub-tasks from a parent task. It's a bit more verbose, but this is the most versatile way to put the tasks together without needed to know too much about their internal structure

```ruby
module Actions
  class GenProduct < Actions::EntryAction
    def plan(name, url, organization)
      sequence do
        create_product = plan_action(CreateProduct, :name => name, :organization_id => organization.id)
        plan_action(CreateRepository, :product_id => create_product.output[:product_id], :name => name, :url => url)

      end
    end

    def humanized_name
      _("Generate Product with Repository")
    end

    class CreateProduct < Actions::Base
      middleware.use ::Actions::Middleware::KeepCurrentUser
      include ::Dynflow::Action::WithSubPlans

      def create_sub_plans
        product = ::Katello::Product.new
        product.name = input[:name]
        product.description = input[:name]
        trigger(::Actions::Katello::Product::Create, product, Organization.find(input[:organization_id])).tap do
          output[:product_id] = product.id
        end
      end
    end

    class CreateRepository < Actions::Base
      middleware.use ::Actions::Middleware::KeepCurrentUser
      include ::Dynflow::Action::WithSubPlans

      def create_sub_plans
        uniq_name = input[:name] + ' repo1'
        label = ::Katello::Util::Model.labelize(uniq_name)
        unprotected = true
        product = ::Katello::Product.find(input[:product_id])
        gpg_key = product.gpg_key
        repository = product.add_repo(label, uniq_name, input[:url], 'yum', unprotected, gpg_key)
        repository.arch = 'noarch'
        repository.mirror_on_sync = true
        repository.verify_ssl_on_sync = true
        trigger(::Actions::Katello::Repository::Create, repository, false, false)
      end
    end
  end
end
```

mdellweg:
Thank you! That helps a lot.
Sadly, the sub-plans are not documented well...

PR #245, opened 2017-03-28
Fixes #18963 - Fix wrong rebase

adamruzicka:
this was originally fixed in https://github.com/theforeman/foreman-tasks/pull/229/commits/1066089b21ef592c6023a2d16fc6eac15336d0da and got lost during rebasing


iNecas:
I've already reviewed that in the original PR, merging, thanks @adamruzicka for pointing out

PR #246, opened 2017-04-03
Fixes #19147 - handle situations when the plan is not available

iNecas:
There are various reasons that can cause the execution plan for a task
not being available. We should still try to not fail completely at those
situations and try to show at least the data we have available in
foreman_tasks_tasks table.

### DIFF
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -49,7 +49,7 @@ module ForemanTasks
       execution_plan = ForemanTasks.dynflow.world.persistence.load_execution_plan(external_id)
       # don't use invalid execution plans for our purposes
       if execution_plan.respond_to?(:valid?) && !execution_plan.valid?
-        raise executor_plan.exception
+        raise execution_plan.exception
       else
         @execution_plan = execution_plan
       end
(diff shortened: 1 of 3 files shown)


app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -102,6 +119,10 @@ def self.consistency_check
             logger.warn('Task %s updated at consistency check: %s' % [task.id, changes.inspect])
           end
         rescue => e
+          # if we fail updating the data from dynflow, it usually means there is something
+          # odd with the data consistency and at this point it is not possible to resume, switching
+          # the tsk to stopped/error
+          task.update_attributes(:state => 'stopped', :result => 'error')
```
beav: This will help a lot when debugging issues, thanks!

beav:
change looks good to me. This will help a lot with tasks that are in running state but are not running for whatever reason.

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -46,7 +46,14 @@ def progress
 
     def execution_plan(silence_exception = true)
       return @execution_plan if defined?(@execution_plan)
-      @execution_plan = ForemanTasks.dynflow.world.persistence.load_execution_plan(external_id)
+      execution_plan = ForemanTasks.dynflow.world.persistence.load_execution_plan(external_id)
+      # don't use invalid execution plans for our purposes
+      if execution_plan.respond_to?(:valid?) && !execution_plan.valid?
+        raise executor_plan.exception
```
adamruzicka: `s/or/ion/`

iNecas:
Nice catch @adamruzicka! Fix pushed

adamruzicka:
APJ

PR #247, opened 2017-04-12
Fixes #16922 - Clean old tasks automaticaly

mbacovsky:
Setup to cleanup tasks older then 30 days.
While the config is in foreman-tasks.yaml it is reflected by
'foreman_tasks:cleanup:config' rake task.
The time was set to 7:45 pm so it happens before nightly syncs
and do not collide with other default tasks of Foreman

### DIFF
--- a/deploy/foreman-tasks.cron.d
+++ b/deploy/foreman-tasks.cron.d
@@ -1,0 +1,8 @@
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+RAILS_ENV=production
+FOREMAN_HOME=/usr/share/foreman
+
+# Clean up the session entries in the database
+45 19 * * *     foreman    /usr/sbin/foreman-rake foreman_tasks:cleanup >>/var/log/foreman/cron.log 2>&1
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -13,11 +13,11 @@
 # Cleaning configuration: how long should the actions be kept before deleted
 # by `rake foreman_tasks:clean` task
 #
-#  :cleanup:
+  :cleanup:
 #
 # the period after which to delete all the tasks (by default all tasks are not being deleted after some period)
 #
-#    :after: 365d
+    :after: 30d
 #
 # per action settings to override the default defined in the actions (self.cleanup_after method)
 #


mbacovsky:
Packaging PR will follow

deploy/foreman-tasks.cron.d
```diff
@@ -0,0 +1,8 @@
+SHELL=/bin/sh
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+
+RAILS_ENV=production
+FOREMAN_HOME=/usr/share/foreman
+
+# Clean up the session entries in the database
+45 19 * * *     foreman    /usr/sbin/foreman-rake foreman_tasks:cleanup >>/var/log/foreman/cron.log 2>&1
```
iNecas: What's the reasoning behind this time?

mbacovsky: As was said in PR description: The time was set to 7:45 pm so it happens before nightly syncs
and do not collide with other default tasks of Foreman. Feel free to suggest more reasonable time.

iNecas: Oh, right, have missed that, makes sense

adamruzicka:
@mbacovsky What's the status here?

bkearney:
@iNecas @mbacovsky @adamruzicka  bump on this

ehelms:
I'm going to be a bit of the negative nancy here for a second, so I do apologize. This will by default, automatically delete tasks from a users database without their knowledge of such a thing occurring. If a user has requirements to keep data around for a particular lifetime, or they are monitoring trends over time by analyzing tasks this will erase that without the user having any knowledge or ability to configure it. Could they update the cronjob themself? Yep! But if this gets updated, install of the new job would wipe out customization most likely.

I feel that this should be a documented feature, or that it should be deployed commented out with a user having to opt-in to this by un-commenting the script. This almost feels like it would be better included as part of foreman-maintain. e.g.

```
foreman-maintain cleanup-tasks --cron-job --cron-entry 45 19 * * * 
```

iNecas:
@ehelms that's why we put http://projects.theforeman.org/issues/20352/ as a prerequisity for doing this, so that the data are not completely lost.

ehelms:
@iNecas thats a good solution to help with no lost data. I however, stand by assertion that any auto data deletion should be opt-in, not opt-out as is the case with this change.

iNecas:
The thing is, with the amount of data in tasks, almost any user will need to do this sooner or later. So far, people don't seem to care much about this: but they care when the system stops working for some reason. Please raise a needinfo on the reporter in BZ and let's discuss it there for better visibility for the people that care about this

lzap:
I vote for cleaning all *successful* tasks after one month and all failed/pending/error tasks after one year. Foreman database must not be abused as audit, if user want auditing, we should write into audit logs or something.

Does this patch make any difference between green and red finished tasks?

beav:
@lzap I agree, I think 1 month and 1 year are suitable. The only exception I can think of is to not delete the most recent reposync task for a particular repo, so the web UI does not show "never synced".

I think auditing is useful in some cases but this is not the best database to store historical records in.

lzap:
Let's only delete sync tasks after one year or something? Sounds like a plan guys?

ehelms:
Could we do this through the application somehow? Or make it configurable through the application? I still stand by that this hsould be opt-in, not opt-out.

adamruzicka:
@ehelms Would being able to enable/disable this by setting a flag in the installer work for you?

We could then make the transition to automatic cleanup smoother by recommending it at first and if everything goes well make a note in the docs about it and enable it by default after a release or two when basically everyone is using it or at least knows about it.

lzap:
@ehelms if this is opt-in, most users won't care enabling this ultimately leading to gigantic databases. Relation database is not the tool for recording purposes. Ideally tasks should automatically put all finished tasks out of the RDBM writing them to an audit log or something. But I don't see this feature coming.

ehelms:
Installer configuration would work for me.

@lzap I don't disagree with the sentiment, but we can't suddenly start deleting data. Some users may be using the historical data, some users may be using it for some sort of compliance purpose, some users might be using it for who knows what. So my point is, because we don't know, we can't just start deleting data out from under users without their knowledge.

Users may not even realize this option exists, but we can document and advertise it to the best of our abilities. I think that's the best we can do?

I would also ask, will this cause in performance issues? Or spikes in performance when it runs for larger users? I assume this is fairly light weight, but I still figure, given our past, we should ask the question in case we need to warn users to expect a spike when it runs.

bkearney:
+1 to an installer option and|or exposing it via the settings.
-1 to different times. If there is an archive model, use that.

sideangleside:
I lean strongly to the side of enabling deletion of tasks by default, if (and only if) those tasks are written somewhere else. That can be either a different table in the database or an external log. I am indifferent as to where. 

Having the user opt-in would imply that 

- the user read the docs
- the user understands the (long term) ramifications of the installer option. 

The downside of opt-out is that the user doesn't know that they need to delete tasks until it is usually too late and they are suffering from performance / disk space issues.  IMHO, having the app 'keep itself healthy' is the right thing to do. Just make sure you preserve the user's data _somewhere_ because they might need it.

adamruzicka:
The installer PR is now opened, so far opt-in, I'm still waiting for the majority to agree on either option.

> I would also ask, will this cause in performance issues? Or spikes in performance when it runs for larger users?

Given it will run every day it should clean up only a "day's worth" of tasks, which shouldn't cause any major issues, but it would be probably nice to have someone measure it with large scale deployment before we roll this out.

iNecas:
Anyway, I'm ok with merging this and keeping the default decision in the installer. We can start with opt-in and turn into opt-out later. I would like to keep get this thing going.

iNecas:
Merged prematurely this PR, as I have not realized it was opened/updated by #278. For future reference, let's close on PR if another PR is supposed to replace it to avoid this in te future. Reverted as eef71c51103e42facef8d670fa286669d99e91b4

PR #249, opened 2017-05-17
Fixes #19570 - Check exception class properly

adamruzicka:


iNecas:
Looks good, waiting for stable builds to that jenkins can run, ACK from me

PR #254, opened 2017-05-30
Fixes #18356 - Assume incoming output is UTF-8

adamruzicka:
`net-ssh` read the data as bytes and assumes it is `ASCII-8BIT` encoded. Sadly we have no reliable way of detecting the encoding of the incoming data.

### DIFF
--- a/lib/foreman_tasks_core/runner/base.rb
+++ b/lib/foreman_tasks_core/runner/base.rb
@@ -42,7 +42,7 @@ module ForemanTasksCore
       end
 
       def publish_data(data, type)
-        @continuous_output.add_output(data, type)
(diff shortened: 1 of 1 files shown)


lib/foreman_tasks_core/runner/base.rb
```diff
@@ -42,7 +42,7 @@ def close
       end
 
       def publish_data(data, type)
-        @continuous_output.add_output(data, type)
+        @continuous_output.add_output(data.force_encoding('UTF-8'), type)
```
iNecas: shouldn't we override the `publish_method` in rex to call `super(data.force_endocing('UTF-8'), type)` there

adamruzicka:
Closing since this can be done on REX's side

PR #257, opened 2017-06-12
Fixes #17347 - Stop autorefresh leaking to other pages

adamruzicka:
When starting the autorefresh we store the id of the form, which contains the uuid of current task. The refresh is cancelled if element with that id doesn't exist on page


iNecas:
Although it was not reproducible on newer versions (the reproducer was working against Foreman 11, while with Foreman 15, there seemed to be a difference with how turbolinks are being used. Anyway, the change looks sane and prevents reloading if the turbolink, or something else decides to not refresh the page

PR #258, opened 2017-06-14
Fixes #20080 - Allow runners to set timeout

adamruzicka:
- [x] add info about timeout to continuous output (https://github.com/theforeman/foreman-tasks/pull/258/files#r123245188)
- [X] add separate issue for this change, for better tracking of what needs to be done where and why

### DIFF
--- a/lib/foreman_tasks_core/runner/base.rb
+++ b/lib/foreman_tasks_core/runner/base.rb
@@ -43,6 +43,7 @@ module ForemanTasksCore
 
       def timeout
         # Override when timeouts and regular kills should be handled differently
+        publish_data('Timeout for execution passed, trying to stop the job', 'debug')
         kill
       end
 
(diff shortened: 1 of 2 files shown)


lib/foreman_tasks_core/runner/base.rb
```diff
@@ -41,6 +41,16 @@ def close
         # if cleanup is needed
       end
 
+      def timeout
+        # Override when timeouts and regular kills should be handled differently
+        kill
```
iNecas: What about adding a debug message to `@continuous_output` about the timeout passed, something like `Time out for the execution passed, trying to kill` (perhaps the working could be better: it sounds pretty violent when put out of the context :)

iNecas:
I will agree with the prprocessor here: please add a separate issue for this so that we can track the differences needed + assign the proper tasks version to it once merged.

adamruzicka:
Comments addressed

iNecas:
Thanks @adamruzicka

PR #261, opened 2017-06-28
Fixes #20136 - ignore create permission authorization

ares:
has no effect until https://github.com/theforeman/foreman/pull/4030 is merged but does not do any harm without


ares:
Fixed the rubocop issue, tests should be green now

adamruzicka:
Merged, thanks @ares !

ares:
Awesome, thanks @aruzicka! Any estimation for when can we expect a new tasks release? Once it's out, test in core PR should be green and the permission PR wuld be unblocked, since Katello merged their part too.

PR #263, opened 2017-07-06
Fixes #20230 - Add time ended to error tasks widget

tbrisker:
Also made the table fixed with tooltips on overflow.

### DIFF
--- a/app/views/foreman_tasks/tasks/dashboard/_latest_tasks_in_error_warning.html.erb
+++ b/app/views/foreman_tasks/tasks/dashboard/_latest_tasks_in_error_warning.html.erb
@@ -1,15 +1,17 @@
 <h4><%= link_to _("Latest Warning/Error Tasks"), foreman_tasks_tasks_path(:order=>'started_at DESC') %></h4>
-<table class="table table-striped">
+<table class="table table-striped table-fixed">
   <tr>
-    <th><%= _("Name") %></th>
-    <th><%= _("State") %></th>
-    <th><%= _("Result") %></th>
+    <th class="col-md-5"><%= _("Name") %></th>
+    <th class="col-md-2"><%= _("State") %></th>
+    <th class="col-md-2"><%= _("Result") %></th>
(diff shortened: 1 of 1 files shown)


app/views/foreman_tasks/tasks/dashboard/_latest_tasks_in_error_warning.html.erb
```diff
@@ -1,15 +1,17 @@
...
-<table class="table table-striped">
+<table class="table table-striped table-fixed">
   <tr>
-    <th><%= _("Name") %></th>
-    <th><%= _("State") %></th>
-    <th><%= _("Result") %></th>
+    <th class="col-md-5"><%= _("Name") %></th>
+    <th class="col-md-2"><%= _("State") %></th>
+    <th class="col-md-2"><%= _("Result") %></th>
+    <th class="col-md-3"><%= _("Ended at") %></th>
```
adamruzicka: Would it make sense to drop the 'at' from the header or reword it some other way? `Ended at: 5 minutes ago` feels weird, but maybe it's just me.

iNecas: Just `Ended` should work fine, or?

tbrisker: changed to Ended

iNecas:
Thanks @tbrisker

iNecas:
and @adamruzicka for review

beav:
(sorry for the post-merge review, I didn't catch the issue til just now 🎣 )

If a task is in paused state, it won't have an ended_at and the widget will error. I had to use this line instead:

`<td><%= task.ended_at ? _('%s ago') % time_ago_in_words(task.ended_at) : '' %></td>`

Other than that, the change looks good.

tbrisker:
@beav thanks for the catch, opened https://github.com/theforeman/foreman-tasks/pull/266

PR #264, opened 2017-07-10
Fixes #20250 - Open dynflow console in new window

adamruzicka:


iNecas:
Nice and clean :) Thanks @adamruzicka

PR #266, opened 2017-07-11
Refs #20230 - Show task start time on dashboard widget

tbrisker:


app/views/foreman_tasks/tasks/dashboard/_latest_tasks_in_error_warning.html.erb
```diff
@@ -11,7 +11,7 @@
       <td class="ellipsis"><%= link_to task.humanized[:action], defined?(main_app) ? main_app.foreman_tasks_task_path(task.id) : foreman_tasks_task_path(task.id) %></td>
       <td><%= task.state %></td>
       <td><%= task.result %></td>
-      <td><%= _('%s ago') % time_ago_in_words(task.ended_at) %></td>
+      <td><%= task.ended_at ? _('%s ago') % time_ago_in_words(task.ended_at) : _('Paused') %></td>
```
adamruzicka: I'd probably go with @beav's suggestion and only display a blank field. Displaying `paused` twice in one row looks weird
screenshot_20170711_153624

tbrisker: I prefer not having an empty column, would "N/A" make sense? btw, is there any way to tell when the task was paused? that would be better then not showing anything.

beav: I was trying to figure that out as well but didn't see anything obvious. The best would be a way to grab the last updated time on the task but I don't know if thats possible. We would also need to change the "Ended" col to "Updated" on the table if we find out how to get this info.

I would vote 👍 to "N/A" if finding the updated timestamp is not possible.

adamruzicka: We don't keep this kind of information in tasks. We could reach over to dynflow and pull this information from there, but I'd rather not go that way. I'd say it's not *that* important information, so plain 'N/A' will have to suffice

tbrisker: The idea is that users will know if this is a new task that's erroring or an old one they already saw, i think that would make sense to show the information if we can.

adamruzicka: One could do something like this. @iNecas is there a more elegant way to do this?
```ruby
Time.at(task.execution_plan.execution_history.events.last.time)
```

it might need some timezone magic to display properly.

iNecas: I agree the `N/A` partially doesn't help with the ordering. I'm not keen on using Dynflow internal data here. What about just using `started_at`, that is always available?

beav: That is a good suggestion, I would be 👍 to `started_at` which avoids the paused/completed edge case and still presents useful info.

tbrisker:
Changed to start time.

beav:
looks good to me as well, very nice

thanks!

PR #267, opened 2017-07-13
Fixes #20300 - Support for aborting proxy tasks

adamruzicka:
requires https://github.com/Dynflow/dynflow/pull/240


iNecas:
Tested and works well: let's release a new dynflow version and raise a dependency here, before merging.

iNecas:
Rubocop doesn't seem to be that happy sa I am :)

iNecas:
Thanks @adamruzicka, I will do cherry-picks to 0.9.x branch so that we can use it in rex with foreman-1.15

PR #268, opened 2017-07-17
Fixes #20310 - handle missing option

iNecas:


lib/foreman_tasks/dynflow/daemon.rb
```diff
@@ -30,7 +30,7 @@ def run(foreman_root = Dir.pwd, options = {})
 
       ForemanTasks.dynflow.executor!
 
-      if options[:memory_limit] > 0
+      if options[:memory_limit] && options[:memory_limit] > 0
```
ehelms: You repeated the same thing here :)

iNecas: What a nitpicker! :D Updated

dLobatog:
Not sure if this makes sense to merge now - it's incorporated in https://github.com/Dynflow/dynflow/pull/236

iNecas:
Depends if @ehelms needs this for nightlies only, or if this should be backported do 1.15? If nightlies, I agree with having this fixed in dynflow as enough

ehelms:
Currently, I have only seen this break in nightlies.

ehelms:
However, this does upgrade the code to be more robust to ensure tight checking no matter what fix is in Dynflow (unless this entire chunk of code would be removed).

iNecas:
In dynflow is the very same test, closing for now

PR #271, opened 2017-07-25
Fixes #20352 - backup deleted tasks

mbacovsky:
This patch uses and extends https://github.com/Dynflow/dynflow/pull/241.
- adds backup of foreman tasks and ensures it lands in the same dir as the backup of related dynflow objects
- adds a way to configure the backups (the default dir is `./backup` for now though. Should it be set during packaging?)

```
cat >> /etc/foreman/plugins/foreman-tasks.yaml <<EOF

  :backup:
    :backup_deleted_plans: true
    :backup_dir: /var/lib/foreman/tasks-backup
```

### DIFF
--- a/lib/foreman_tasks/cleaner.rb
+++ b/lib/foreman_tasks/cleaner.rb
@@ -78,9 +78,10 @@ module ForemanTasks
         say "[noop] #{ForemanTasks::Task.search_for(full_filter).size} tasks would be deleted"
       else
         start_tracking_progress
+        backup_dir = ForemanTasks.dynflow.world.persistence.current_backup_dir
         while (chunk = ForemanTasks::Task.search_for(full_filter).limit(batch_size)).any?
-          delete_tasks(chunk)
-          delete_dynflow_plans(chunk)
+          delete_tasks(chunk, backup_dir)
+          delete_dynflow_plans(chunk, backup_dir)
           report_progress(chunk)
         end
       end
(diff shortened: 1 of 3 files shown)


lib/foreman_tasks/cleaner.rb
```diff
@@ -90,13 +93,29 @@ def tasks
       ForemanTasks::Task.search_for(full_filter).select('DISTINCT foreman_tasks_tasks.id, foreman_tasks_tasks.type, foreman_tasks_tasks.external_id')
     end
 
-    def delete_tasks(chunk)
-      ForemanTasks::Task.where(:id => chunk.map(&:id)).delete_all
+    def delete_tasks(chunk, backup_dir = ForemanTasks.dynflow.world.persistence.current_backup_dir)
```
adamruzicka: This breaks compatibility with remote execution's extensions for the cleaner, I'm not against the change, just making a note of it

iNecas: It seems to me having the `backup_dir` as instance variable configured in the initializer would help here, wouldn't it? It would also reduce some code duplicity.

mbacovsky: @iNecas makes sense as Cleaner instance has short life

mbacovsky:
Moved backup_dir to instance variable

lib/foreman_tasks/cleaner.rb
```diff
@@ -91,12 +95,28 @@ def tasks
...
+        csv << columns.to_csv unless appending
+        dataset.each do |row|
+          csv << row.attributes.values.to_csv
+        end
+      end
+      dataset
     end
 
-    def delete_dynflow_plans(chunk)
+    def delete_dynflow_plans(chunk, backup_dir = ForemanTasks.dynflow.world.persistence.current_backup_dir)
```
adamruzicka: Any reason for passing the `backup_dir` explicitly instead of using `@backup_dir`?

mbacovsky:
Updated - removed backup_dir also from delete_dynflow_plans, fixed rubocop.
The tests are failing due to missing related changes in Dynflow. It should wok fine once dynflow PR is merged. Should I uopdate the dependency on dynflow?

adamruzicka:
Tested with https://github.com/Dynflow/dynflow/pull/241 and it works as expected. Please bump the dependency on dynflow. Tests passed locally just fine, but we'll wait for Mr. Jenkins to confirm that.

iNecas:
@mbacovsky there are conflicts with the current maser, Please resolve

lib/foreman_tasks/dynflow/configuration.rb
```diff
@@ -6,6 +6,27 @@
 module ForemanTasks
   # Import all Dynflow configuration from Foreman, and add our own for Tasks
   class Dynflow::Configuration < ::Foreman::Dynflow::Configuration
+    def world_config
+      super.tap do |config|
+        config.backup_deleted_plans = backup_settings[:backup_deleted_plans]
+        config.backup_dir          = backup_settings[:backup_dir]
```
adamruzicka: Nitpick: needs one extra space to align the `=`

adamruzicka:
Looks good to me, let's wait what Jenkins has to say

adamruzicka:
@iNecas anything to change here or can we get it in?

iNecas:
Closing in favor of  #277

PR #272, opened 2017-07-26
Fixes #20417 - follow standard plugin assets structure

iNecas:
Requires https://github.com/theforeman/foreman/pull/4698


adamruzicka:
I'm failing to reproduce the issue with theforeman/foreman#4698 and everything seems to work both with and without this, but I'm not against merging it as a refactoring

iNecas:
Thakns @adamruzicka for review

PR #273, opened 2017-07-26
Refs #20230 - handle missing started_at in tasks widget

iNecas:
There are some corner-cases where the started_at doesn't have to be specified.


adamruzicka:
Looks good to me. @iNecas  do we care about commit message line length?

iNecas:
Nope :)

PR #275, opened 2017-07-31
Fixes #20443 - Update table pagination to use patternfly guidelines

adamruzicka:


ohadlevy:
I suggest to add the common tests (basic_pagination_per_page_test and basic_pagination_rendered_test) so this will ensure tests do not fail in the future.

test/controllers/tasks_controller_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'foreman_tasks_test_helper'
+
+module ForemanTasks
+  class TasksControllerTest < ActionController::TestCase
+
+    describe ForemanTasks::TasksController do
+      basic_pagination_per_page_test
```
ohadlevy: maybe worth adding  basic_index_test too?

app/views/foreman_tasks/tasks/index.html.erb
```diff
@@ -39,4 +39,4 @@ $(document).on('click', ".table-two-pane td.two-pane-link", function(e) {
   <% end %>
 </table>
 <%= page_entries_info @tasks %>
-<%= will_paginate @tasks %>
+<%= will_paginate_with_info @tasks %>
```
ohadlevy: also, I believe you can pass model: _('Tasks') to change the display (instead of foreman-tasks/tasks)

test/controllers/tasks_controller_test.rb
adamruzicka: I thought it required some more changes to the controller, but it didn't. Added that

app/views/foreman_tasks/tasks/index.html.erb
adamruzicka: Removed the `page_entries_info` altogether to closer match the look of foreman

ohadlevy:
in that case, can you remove it from recurring_logics/index.html.erb too?

adamruzicka:
Did the same for `recurring_logics`, couldn't include the `basic_pagination_rendered_test`, because one of the tests there relies on searching for the resource by name and recurring logics do not support it. That test would never pass.

iNecas:
Thanks @adamruzicka and @ohadlevy

PR #276, opened 2017-08-17
Fixes #20624 - Silence undeliverable messages sent by Ticker

adamruzicka:


iNecas:
New dynflow got out: please dump the dynflow version dependency.

adamruzicka:
Dependency bumped and the tests are green

iNecas:
Thanks @adamruzicka

iNecas:
I know we already merged this but still: do we even need this change vs. doing this directly in smart_proxy_dynlfow? Just to simplify the flow a bit. What was the intention behind doing also though foreman-tasks-core?

adamruzicka:
Well, we could do it directly in dynflow core, but I don't think dynflow core should know directly which actors from plugins might need silencing. This way we have an interface through which foreman tasks can add more things which need to be silenced, without having to change dynflow core.

PR #277, opened 2017-08-24
Fixes #20352 - Backup deleted tasks

adamruzicka:
includes https://github.com/theforeman/foreman-tasks/pull/271 and adds tests on top. I needed to break it up into more functions to ease writing tests.

### DIFF
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -9,6 +9,10 @@
 #     :action:
 #       :enabled: true
 
+  :backup:
+    :backup_deleted_tasks: true
+    :backup_dir: /var/lib/foreman/tasks-backup
+
 
 # Cleaning configuration: how long should the actions be kept before deleted
 # by `rake foreman_tasks:clean` task
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -9,10 +9,18 @@
 #     :action:
 #       :enabled: true
 
-  :backup:
-    :backup_deleted_tasks: true
-    :backup_dir: /var/lib/foreman/tasks-backup
-
+# Task backup configuration can be changed by uncommenting the backup
+# section
+#
+#   :backup:
+#
+# Whether to back up tasks when they are removed
+#
+#     :backup_deleted_tasks: true
+#
+# Where to put the tasks which were backed up
+#
+#     :backup_dir: /var/lib/foreman/tasks-backup
 
 # Cleaning configuration: how long should the actions be kept before deleted
 # by `rake foreman_tasks:clean` task
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -9,18 +9,18 @@
 #     :action:
 #       :enabled: true
 
-# Task backup configuration can be changed by uncommenting the backup
-# section
-#
-#   :backup:
-#
-# Whether to back up tasks when they are removed
-#
-#     :backup_deleted_tasks: true
-#
-# Where to put the tasks which were backed up
-#
-#     :backup_dir: /var/lib/foreman/tasks-backup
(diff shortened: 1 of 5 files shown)


iNecas:
Rubocop 🔴

lib/foreman_tasks/dynflow/configuration.rb
```diff
@@ -6,6 +6,27 @@
...
+        :backup_deleted_plans => true,
+        :backup_dir => default_backup_dir
+      }
+      settings = SETTINGS[:'foreman-tasks'] && SETTINGS[:'foreman-tasks'][:backup]
+      backup_options.merge!(settings) if settings
+      backup_options
+    end
+
+    def default_backup_dir
+      "./backup"
```
iNecas: As this value is mainly applicable in development, what about using `File.join(Rails.root, 'tmp', 'task-backup')` instead to not make mess in the root foreman directory and need to add it to `.gitignore`

adamruzicka: No need to add to .gitignore because the backups will end up in foreman/tmp, which is already ignored

config/foreman-tasks.yaml.example
```diff
@@ -9,6 +9,18 @@
...
+#
+#   :backup:
+#
+# Whether to back up tasks when they are removed
+#
+#     :backup_deleted_tasks: true
+#
+# Where to put the tasks which were backed up
+#
+#     :backup_dir: /var/lib/foreman/tasks-backup
```
iNecas: I think this was not commented on purpose, as it's being used in the packaging https://github.com/mbacovsky/foreman-packaging/blob/22d5cb766cf88646bdd72cf2bf433718c0e4c6f9/rubygem-foreman-tasks/rubygem-foreman-tasks.spec#L97

adamruzicka: Oh, I see

lib/foreman_tasks/tasks/cleanup.rake
```diff
@@ -9,6 +9,7 @@ namespace :foreman_tasks do
         * NOOP        : set to "true" if the task should not actuall perform the deletion
         * VERBOSE     : set to "true" for more verbose output
         * BATCH_SIZE  : the size of batches the tasks get processed in (1000 by default)
+        * TASK_BACKUP : the path where to place backup of cleaned tasks (set to empty to disable)
```
iNecas: Capturing the off-line conversation: we need to make the TASK_BACKUP field to be used when initializing the dynflow world to fully propagate to all areas. It should be true/false flag, rather than directory setup, to reduce the confusion around it

lib/foreman_tasks/dynflow/configuration.rb
```diff
@@ -14,19 +14,29 @@ def world_config
...
+      end
+      @backup_settings = backup_options
     end
 
     def default_backup_dir
       File.join(Rails.root, 'tmp', 'task-backup')
     end
 
+    def environment_override(env_var)
+      # Everything except 0, n, no, false is considered to be a truthy value
```
iNecas: Using `ENV['TASK_BACKUP']` directly here would be more readable, as it would encapsulate all we need to know about the influence of the env variable.

iNecas:
Worked as expected. Waiting for the jenkins to pass: the next thing to figure out is compression of older directories + cleanup of the backups after some time. Could we use logrotate for this purpose?

adamruzicka:
The tests are green.

I don't think logrotate can operate on directories directly. We would need to drop the "daily directory", put all the backups into one directory and rotate its contents.

iNecas:
there are conflicts from merging previous Pr, please resolve. Let's start tracking the deletion of backup in another issue and look into that after this gets merged

iNecas:
Thanks @adamruzicka and @mbacovsky

PR #278, opened 2017-08-29
Fixes #16922 - Clean old tasks automaticaly

adamruzicka:
This PR includes https://github.com/theforeman/foreman-tasks/pull/247. Based on discussion in original PR this now also allows to distinguish between green an red (and any other) tasks and set different cleanup intervals.

### DIFF
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -28,7 +28,7 @@
 #
 # Rules defined in this section by default don't operate
 # on tasks specified in the actions section. This behavior
-# can be overriden by setting the avoid_actions to false
+# can be overriden by setting the override_actions to true
    :rules:
      # Delete successful tasks after a month
      - :filter: result = success
@@ -36,4 +36,4 @@
      # Delete everything (any action, any state) after one year
      - :states: []
        :after: 1y
-       :avoid_actions: false
+       :override_actions: true
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -16,7 +16,7 @@
   :cleanup:
 #
 # the period after which to delete all the tasks (by default all tasks are not being deleted after some period)
-# will be deprecated (TODO: mention when) and the use of rules is recommended
+# will be deprecated in Foreman 1.18 and the use of rules is recommended.
 #
 #     :after: 30d
 #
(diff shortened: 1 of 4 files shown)


config/foreman-tasks.yaml.example
```diff
@@ -16,11 +16,24 @@
   :cleanup:
 #
 # the period after which to delete all the tasks (by default all tasks are not being deleted after some period)
+# will be deprecated (TODO: mention when) and the use of rules is recommended
```
iNecas: Whatever comes with Foreman 1.18?

lib/foreman_tasks/cleaner.rb
```diff
@@ -11,11 +11,17 @@ def self.run(options)
           end
         end
         if cleanup_settings[:after]
+          # TODO: Mention version in which this will be removed completely
+          ActiveSupport::Deprecation.warn('You are using legacy interface, please migrate from using :after to cleanup rules')
```
iNecas:
```ruby
Foreman::Deprecation.deprecation_warning("1.18", _(':after setting in tasks cleanup section is deprecated, use :after in :rules section to set the value. to cleanup rules'))
```

iNecas:
Let's rebase this and add some tests before we proceed

adamruzicka:
The test matrix is all green but the build was marked as failed. Let's try it again

[test]

config/foreman-tasks.yaml.example
```diff
@@ -13,14 +13,27 @@
...
+#
+# Rules defined in this section by default don't operate
+# on tasks specified in the actions section. This behavior
+# can be overriden by setting the override_actions to true
+   :rules:
+     # Delete successful tasks after a month
+     - :filter: result = success
+       :after: 30d
+     # Delete everything (any action, any state) after one year
+     - :states: []
```
iNecas: I know I just keep adding comments here, but this is the last one (hopefully:). It's not very clear that empty states means all while missing/nil means 'stopped'. What about introducing 'all' as recognized value that would replace the `[]` optoin we have now.

iNecas:
Looks good, waiting for tests to finish. There is one failure so far, but seems not to be related directly

adamruzicka:
The tests are done, with one (most likely) unrelated failure.

iNecas:
Thanks @adamruzicka and @mbacovsky

PR #279, opened 2017-09-13
Fixes #20914 - Connect to configured Foreman DB during export

mbacovsky:
foreman-debug extension was updated to use DB from Foreman configuration
to export foreman-tasks DB data. Before the patch connection to local
postgresql server and database 'foreman' was hard-coded.

### DIFF
--- a/extra/dynflow-debug.sh
+++ b/extra/dynflow-debug.sh
@@ -3,8 +3,18 @@
 # This file provides additional debug information for foreman-debug tool and is
 # symlinked as /usr/share/foreman/script/foreman-debug.d/60-dynflow_debug
 
-GET_DB_CONFIG_SCRIPT='db = YAML.load_file("/etc/foreman/database.yml")["production"];puts %(PGPASSWORD="#{db["password"]}" psql -U #{db["username"]} -h "#{db["host"]||"localhost"}" -p #{db["port"]||"5432"} -d #{db["database"]})'
-FOREMAN_PSQL_COMMAND=`ruby -ryaml -e "$GET_DB_CONFIG_SCRIPT"`
+FOREMAN_PSQL_COMMAND=$(ruby -ryaml <<-END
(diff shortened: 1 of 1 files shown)


extra/dynflow-debug.sh
```diff
@@ -3,8 +3,11 @@
 # This file provides additional debug information for foreman-debug tool and is
 # symlinked as /usr/share/foreman/script/foreman-debug.d/60-dynflow_debug
 
+GET_DB_CONFIG_SCRIPT='db = YAML.load_file("/etc/foreman/database.yml")["production"];puts %(PGPASSWORD="#{db["password"]}" psql -U #{db["username"]} -h "#{db["host"]||"localhost"}" -p #{db["port"]||"5432"} -d #{db["database"]})'
```
adamruzicka: Whoa, this is pretty wild. Any reason for not having it span multiple lines?
```shell
FOREMAN_PSQL_COMMAND=$(ruby -ryaml <<-END
db = YAML.load_file("/etc/foreman/database.yml")["production"]
puts [
  "PGPASSWORD='#{db['password']}'",
  'psql',
  '-U', "'#{db.fetch('username')}'",
  '-h', "'#{db['host'] || 'localhost'}'",
  '-p', "'#{(db['port'] || 5432).to_s}'",
  '-d', "'#{db.fetch('database')}'"
].join(' ')
END
)
```

mbacovsky:
@adamruzicka, good point, I was struggling with the right quoting. Thanks for the suggestion

adamruzicka:
As per discussion on IRC, we should check for database type and do the export only if the db is postgresql and skip it for any other.

mbacovsky:
@adamruzicka updated

adamruzicka:
Works as expected. @iNecas do we really need the tests to be green for this?

iNecas:
We don't

iNecas:
Thanks @mbacovsky and @adamruzicka

PR #280, opened 2017-09-14
Refs #20917 - Started field should show timestamp

snagoor:
Started filed should show timestamps rather than the time it was running. Feature request was more of the timestamp in the  "latest warning/error tasks" widget.

Also as a end user, I see only the latest 5 tasks in the widget, where timestamp would be more feasible rather than when it was started. See image below 

Latest Warning / Errors Tasks

### DIFF
--- a/app/views/foreman_tasks/tasks/dashboard/_latest_tasks_in_error_warning.html.erb
+++ b/app/views/foreman_tasks/tasks/dashboard/_latest_tasks_in_error_warning.html.erb
@@ -11,7 +11,7 @@
       <td class="ellipsis"><%= link_to task.humanized[:action], defined?(main_app) ? main_app.foreman_tasks_task_path(task.id) : foreman_tasks_task_path(task.id) %></td>
       <td><%= task.state %></td>
       <td><%= task.result %></td>
-      <td><%= task.started_at ? (_('%s ago') % time_ago_in_words(task.started_at)) : _('N/A') %></td>
+      <td><%= task.started_at ? (_(date_time_relative(task.started_at))) : _('N/A') %></td>
     </tr>
   <% end %>
 </table>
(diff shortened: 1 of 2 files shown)


adamruzicka:
> where timestamp would be more feasible rather than when it was started.

But even with you're patch it is still showing "when it was started", it is just more precise. Also the benefit of exact representation (On Oct 25 12:10) over approximate representation in words (20 days ago) is quite arguable.

iNecas:
Since we just merged https://github.com/theforeman/foreman/pull/4820, I think the new helpers should be used instead. It also show relative vs. absolute value in tooltip, which should help both sides of the argument to find the right value quickly. So I suggest starting using `date_time_relative` helper, which provides the absolute value in the tooltip, so that both information are available.

I agree with adam, that the relative information is more useful here to see if the info there is something that happened today/yesterday, or 60 days ago.

iNecas:
Please also use Refs, instead of Fixes, as the original issue was already fixed, or file different issue

snagoor:
@iNecas, I made the changes as per the new helper `date_time_relative`.  Can you have another look. Here is the sample image 

screenshot from 2017-09-15 15-20-17

snagoor:
Modified both `Latest Warning/Error Tasks` and `Task Status` widgets with new `date_time_relative` helper

iNecas:
Looks good. Waiting for stable builds + 1.16 RC builds so that we are sure the helpers are already there before merging this one.

iNecas:
I've checked and the helpers are now officially in 1.16. This one is good to go

iNecas:
Thanks @snagoor !

ares:
@iNecas I think this should have bumped the required Foreman version https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L44

also this will need a change in packaging when new version is built

PR #284, opened 2017-09-21
Fixes #18422 - Expose pool_size in config file

adamruzicka:
Or should this go directly into core, now that core uses dynflow as well? Should this also be configurable for smart_proxy_dynflow_core?

### DIFF
--- a/config/foreman-tasks.yaml.example
+++ b/config/foreman-tasks.yaml.example
@@ -9,6 +9,11 @@
 #     :action:
 #       :enabled: true
 
+#
+# Configure how many workers should be used to handle incoming requests.
+#
+#   :pool_size: 5
+
   # Task backup configuration can be changed by altering the values in
   # the backup section
   #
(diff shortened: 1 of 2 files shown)


iNecas:
Rebase needed. This leads me to the idea to expose the number of workers via https://github.com/Dynflow/dynflow/pull/251, which might be pretty interesting information, especially as people will start tweaking more the values.

iNecas:
Tested and works as expected. Thanks @adamruzicka

iNecas:
One question though: I have a feeling that this is something that should go to the foreman core under `:dynflow:` key in `settings.yml`

adamruzicka:
Would it make sense to have it in both places? With the setting in foreman-tasks overriding the value from foreman for foreman-tasks service?

iNecas:
It would make sense for older versions, but in that case I would probably go with cherry-picking to older versions only, but not having them in nightly. So for now, I would go for core only setting

adamruzicka:
Opened https://github.com/theforeman/foreman/pull/4897

adamruzicka:
Closing in favor of theforeman/foreman#4897

PR #287, opened 2017-09-22
Fixes #21082 - make the owner search join tables uniq

iNecas:


iNecas:
I've also filed a refactor issue for extracting the owner association from locks, but this is outside of scope of this PR http://projects.theforeman.org/issues/21083

app/models/foreman_tasks/task.rb
```diff
@@ -134,26 +134,28 @@ def self.search_by_owner(key, operator, value)
       return { :conditions => '0 = 1' } if value == 'current_user' && User.current.nil?
 
       key = 'owners.login' if key == 'user'
+      # using uniq suffix to avoid colisions when searching by two different owners via ScopedSearch
+      uniq_suffix = SecureRandom.hex
```
adamruzicka: Isn't this an overkill? `SecureRandom.hex` gives us 32 character long string (16^32 different values). Using `SecureRandom.hex(2)` would give us 4 characters long string (16^4 different values), which still should be good enough (and probably a bit faster) with reasonably low conflict chance.

iNecas: I've set it to 3, just in case :)

adamruzicka:
[test]

adamruzicka:
Test failure seems unrelated

PR #289, opened 2017-10-03
Fixes #21173 - rails 5 safe class reloading

jlsherrill:


lib/foreman_tasks.rb
```diff
@@ -20,16 +20,27 @@ def self.trigger(action, *args, &block)
...
+            end),
+            (on ::Dynflow::World::Triggered.call(execution_plan_id: ~any, future: ~any) do |id, finished|
+              finished.wait if async == false
+              ForemanTasks::Task::DynflowTask.where(:external_id => id).first!
+            end)
+    end
+  end
+
+  def self.rails_safe_trigger_task
+    if SETTINGS[:rails] == '4.2'
```
iNecas: Checking on `Rails::VERSION::MAJOR` would be better, as it would not depend on the rails setting being present.

lib/foreman_tasks/engine.rb
```diff
@@ -39,6 +39,12 @@ class Engine < ::Rails::Engine
       Foreman::Gettext::Support.add_text_domain locale_domain, locale_dir
     end
 
+    initializer 'foreman_tasks.rails_executor_middleware' do
+      Actions::Base.class_eval do
+        middleware.use ::Actions::Middleware::RailsExecutorWrap if SETTINGS[:rails] != '4.2'
```
iNecas: Any reason to not put it directly to https://github.com/theforeman/foreman-tasks/blob/212ec9a066187a9e3bfebeb9a20283596ba585d0/app/lib/actions/base.rb#L2 ?

jlsherrill: For some reason i thought that class was in dynflow, i'm not really sure why i thought that

iNecas:
Added some description comment to the middleware, just to keep the intention documented and merged as 25f4b482bf, Thanks @jlsherrill !

iNecas:
Released in 0.10.5

PR #294, opened 2017-10-26
Fixes #20998 - Allow filtering by action column

adamruzicka:


db/migrate/20171026082635_add_task_action.rb
```diff
@@ -0,0 +1,14 @@
+class AddTaskAction < ActiveRecord::Migration
+  def up
+    add_column :foreman_tasks_tasks, :action, :string
+
+    ::ForemanTasks::Task.where(:action => nil).find_each do |task|
+      task.action = task.format_input
```
iNecas: This can be pretty expensive operation, that would slow down the whole migration a lot. I wonder, if we should do it here? Perhaps a rake task to calculate this, if anyone wanted that, but in general, I'm not sure many people would want that for older tasks: if the did, they could run the rake task.

adamruzicka: The way the UI is now, without actually computing the changes all the old tasks would have blank action column in the tasks index. We could fallback to generating this stuff on demand when the user displays the task, but it might break the search.

iNecas: :+1: for fallback

adamruzicka:
The data is now not migrated when running the database migration, there is a rake task for doing that. When the user accesses the WebUI, we check for the data in DB and fall back to generating it on demand. This however breaks the search because only the in-db data is searchable. For this reason a `foreman_tasks:generate_task_actions`, which populates the action column is included. This is not needed for tasks started after this db migration, because they regenerate the column as they change states

lib/foreman_tasks/tasks/generate_task_actions.rake
```diff
@@ -0,0 +1,18 @@
...
+# Run "foreman-rake foreman_tasks:generate_task_actions" to generate missing values.
+
+namespace :foreman_tasks do
+  desc 'Generate missing values for action column in foreman_tasks_tasks table.'
+  task :generate_task_actions => :environment do
+    scope = ::ForemanTasks::Task.where(:action => nil)
+
+    puts "Generating action for #{scope.count} tasks."
+    scope.find_each do |task|
+      task.action = task.format_input
```
iNecas: Some info about the progress would be nice. Let's say printing `done/total', every 100 records or something like that, to give the user some idea on when to expect it will finish.

app/models/foreman_tasks/task.rb
```diff
@@ -36,6 +36,7 @@ def check_permissions_after_save
     end
 
     scoped_search :on => :id, :complete_value => false
+    scoped_search :on => :action, :complete_value => false
```
ShimShtein: I see we are adding yet another implicit field to search. IMHO it would be better to be more explicit, it will make the query simpler and run faster. You can do it easily by adding `:only_explicit => true`

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -19,6 +19,7 @@ def update_from_dynflow(data)
                                 end
                               end
       self.label          ||= main_action && main_action.class.name
+      self.action = format_input(true)
```
ShimShtein: Wouldn't it be better to put action calculation to `:after_validate` callback? This way we will be sure it's updated each time we save the model.

app/models/foreman_tasks/task.rb
iNecas: For this particular, I would probably leave this in the implicit search TBH

app/models/foreman_tasks/task/dynflow_task.rb
iNecas: Would work for me. Also: would it work if we did `self.action ||= `, to save some calcutlation time on each save?

ShimShtein: > would it work if we did self.action ||=, to save some calcutlation time on each save?

It depends if `:name` and `:input` could be changed. For clarity I would suggest using condition on the callback:
``` ruby
after_validation :set_action_field :if => Proc.new { |task| task.name_changed? || task.input_changed? }
```

db/migrate/20171026082635_add_task_action.rb
```diff
@@ -0,0 +1,9 @@
+class AddTaskAction < ActiveRecord::Migration
```
ShimShtein: Could you please generate the migration using our plugin specific task? See our [wiki](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Add-new-migration) for reference.

TL;DR: Please rename the migration to `20171026082635_add_task_action.foreman-tasks.rb`. It will be the first step to dynamic add and remove plugins.

lib/foreman_tasks/tasks/generate_task_actions.rake
```diff
@@ -0,0 +1,41 @@
...
+        @processed = 0
+        puts message % { :count => count } unless message.nil?
+      end
+
+      def progress(count)
+        @processed += count
+      end
+
+      def report
+        puts "Processed #{@processed}/#{@count} tasks"
```
ShimShtein: This needs i18n IMHO...

lib/foreman_tasks/tasks/generate_task_actions.rake
```diff
@@ -0,0 +1,41 @@
+#
+# generate_task_actions.rake is a data migration task to properly fill the missing
+# values into the foreman_tasks_tasks table.
+#
+# Run "foreman-rake foreman_tasks:generate_task_actions" to generate missing values.
+
+namespace :foreman_tasks do
+  desc 'Generate missing values for action column in foreman_tasks_tasks table.'
```
ShimShtein: This needs i18n I believe

app/helpers/foreman_tasks/tasks_helper.rb
```diff
@@ -2,16 +2,7 @@ module ForemanTasks
...
-      parts << task.get_humanized(:name) if include_action
-      parts << Array(task.get_humanized(:input)).map do |part|
-        if part.is_a? Array
-          part[1][:text]
-        else
-          part.to_s
-        end
-      end.join('; ')
-      parts.join(' ')
+      task.action || task.format_input(include_action)
```
ShimShtein: Helper shouldn't know how `action` field is generated. I would prefer leaving this calculation to the model:
``` ruby
# In task.rb:
class Task < ActiveRecord::Base
  def action
    super || format_input
  end
end
```

app/models/foreman_tasks/task.rb
```diff
@@ -202,6 +203,19 @@ def sub_tasks_counts
       result.symbolize_keys
     end
 
+    def format_input(include_action = false)
```
ShimShtein: I think we should get rid of `include_action` parameter. It will cause inconsistent output in the `format_task_input` helper, depending on the fact was the task migrated or created by `dynflow_task`.

lib/foreman_tasks/tasks/generate_task_actions.rake
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      def report
+        puts "Processed #{@processed}/#{@count} tasks"
+      end
+    end
+
+    scope = ::ForemanTasks::Task.where(:action => nil)
+    count = scope.count
+    reporter = ProgressReporter.new count, 'Generating action for %{count} tasks.'
```
ShimShtein: This needs i18n too...

app/models/foreman_tasks/task/dynflow_task.rb
adamruzicka: The thing here is that `name` and `input` are not attributes of the model so we're not getting all the `*_changed?` goodies from rails. Not sure about `name`, but `input` can definitely change during the run of the task so `||=` is probably not a good idea. Any other idea than implementing `*_changed?` ourselves?

ShimShtein: If we can't track changes to `name` and `input`, we will not be able to avoid recalculation. We don't have to write field tracking code from scratch, since we already have it in our models from [ActiveMode::Dirty](http://api.rubyonrails.org/classes/ActiveModel/Dirty.html). Anyway, the field will be omitted from SQL UPDATE statement, if the value remains the same (tested it on another ActiveRecord). I don't think the calculation is that expensive that we should try and optimize it. IMHO, we can set the field each save without any worries.

lib/foreman_tasks/tasks/generate_task_actions.rake
adamruzicka: There is no i18n context defined when the rake tasks are being defined. Added translations to all the other places

ShimShtein:
@adamruzicka, This needs a rebase. Also what about i18n of the rake task description?

adamruzicka:
Rebase is in the works, the rake task's description is being defined outside any i18n context so it probably can't be translated easily. Also none of rake tasks in foreman-tasks has translated description

db/migrate/20171026082635_add_task_action.foreman_tasks.rb
```diff
@@ -0,0 +1,9 @@
+class AddTaskAction < ActiveRecord::Migration
```
iNecas: Needs to be updated to `Rails 5` syntax.

lib/foreman_tasks/tasks/generate_task_actions.rake
```diff
@@ -0,0 +1,41 @@
...
+      def progress(count)
+        @processed += count
+      end
+
+      def report
+        puts _('Processed %{processed}/%{count} tasks') % { :processed => @processed, :count => @count }
+      end
+    end
+
+    scope = ::ForemanTasks::Task.where(:action => nil)
```
iNecas: What about ordering by `started_at desc`? This way, when somebody runs this and cuts it in the middle (because she doesn't want to wait until the script to finish), she still can search in the most recent/relevant tasks.

iNecas:
Thanks @adamruzicka

PR #297, opened 2017-11-15
Fixes #21792 - Hide secrets from proxy action

adamruzicka:


iNecas:
Please:

* create a separate issue for this, so that we can map it to proper tasks release
* add some test to https://github.com/theforeman/foreman-tasks/blob/master/test/unit/actions/proxy_action_test.rb to cover the functionality

adamruzicka:
[test]

adamruzicka:
[test]

adamruzicka:
[test]

iNecas:
Test failures seem to be related: it seems like you should be calling this only once, probably at the most outer context of the test.

```ruby
    before do
      Support::DummyProxyAction.reset
    end
```

iNecas:
Tested and works well. Thanks @adamruzicka

PR #298, opened 2017-11-15
Fixes #23017 - Rex resilience

adamruzicka:
requires https://github.com/theforeman/smart_proxy_dynflow/pull/40


adamruzicka:
Test failure seems unrelated

iNecas:
[test]

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -0,0 +1,73 @@
...
+      def notify(event, actions)
+        actions.each do |action|
+          action.world.event action.execution_plan_id,
+                             action.run_step_id,
+                             event
+        end
+      end
+
+      def delegated_actions
+        action.sub_plans('state' => %w[running])
```
iNecas: This can be potentially expensive and memory-hungry method, loading all the actions data for a job with 10000s of hosts. I'm thinking of an alternative approach, where we would keep the track of association between local action and remote one in a separate table, where we would track the relation between local step and remote task. We could also use this relation to group the requests to the proxy when creating the remote actions.

In short, instead of calling to proxy on every host, we would just mark that the remote action is ready to be triggered at the remote host, and a periodic procedure (that could be done as part of the parent task) would collect the info from all actions that would be ready to be triggered remotely and trigger it in bulk.

I think we could do this in two phases, leaving the bulk triggering as a separate PR, but we could do the work in the PR ready for moving this direction.

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -0,0 +1,73 @@
+module Actions
+  module Middleware
+    class WatchDelegatedProxySubTasks < ::Dynflow::Middleware
+      class CheckOnProxyActions; end
+      POLL_INTERVAL = 30
```
iNecas: This might be quite aggressive interval, especially with large scale. Once in 5 minutes could be quite good, perhaps exposing this through settings, just in case.

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -2,13 +2,13 @@ module Actions
   module Middleware
     class WatchDelegatedProxySubTasks < ::Dynflow::Middleware
       class CheckOnProxyActions; end
-      POLL_INTERVAL = 30
+      POLL_INTERVAL = 10
```
iNecas: Was it intentional to shorten the interval? I would actually expect to prolong this a bit, perhaps even something like once per 5 minutes could be fine.

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -0,0 +1,65 @@
...
+      private
+
+      def set_clock
+        action.world.clock.ping action.send(:suspended_action),
+                                POLL_INTERVAL,
+                                CheckOnProxyActions
+      end
+
+      def check_triggered
+        tasks = remote_tasks.triggered.group_by(&:proxy_url)
```
iNecas: I wonder how this works with 10000s of tasks? Shouldn't we use some iteration though batches to spread the tasks over some "smaller' amounts? Maybe it's fine, but from the previous experience, it's good to keep thinking about this option.

app/lib/actions/proxy_action.rb
```diff
@@ -95,12 +111,21 @@ def on_resume
     # @override to put custom logic on event handling
     def on_data(data)
       output[:proxy_output] = data
-      wipe_secrets!
     end
 
-    def wipe_secrets!
+    # Removes the :secrets key from the action's input and output and saves the action
+    def wipe_secrets!()
```
iNecas: Why the `()` for empty arguments?

app/lib/actions/proxy_action.rb
```diff
@@ -156,6 +181,10 @@ def set_timeout!
       metadata[:timeout] = time.to_s
     end
 
+    def process_timeout
```
iNecas: Why are we doing this?

app/lib/actions/proxy_action.rb
```diff
@@ -5,6 +5,9 @@ class ProxyAction < Base
 
     middleware.use ::Actions::Middleware::HideSecrets
 
+    execution_plan_hooks.use :clean_remote_task, :on => :stopped
```
iNecas: We should bump dependency on on dynflow to 0.8.37 to ensure we have the hooks in dynflow

app/lib/actions/proxy_action.rb
```diff
@@ -39,18 +45,28 @@ def run(event = nil)
...
+      @remote_task ||= ForemanTasks::RemoteTask.find_by(:execution_plan_id => execution_plan_id, :step_id => run_step_id)
+    end
+
     def trigger_proxy_task
       suspend do |_suspended_action|
         set_timeout! unless timeout_set?
         response = proxy.trigger_task(proxy_action_name,
                                       input.merge(:callback => { :task_id => task.id,
                                                                  :step_id => run_step_id }))
+        ::ForemanTasks::RemoteTask.new(:remote_task_id => response['task_id'], :execution_plan_id => execution_plan_id,
```
iNecas: Can't be used `create!` directly?

iNecas:
There is conflict at the moment. Also, please create actually an issue for foreman-tasks: the foreman bot's getting kind of spammy, which makes it harder to figure out what's actually going on.

app/lib/actions/proxy_action.rb
```diff
@@ -95,12 +111,21 @@ def on_resume
...
-    def wipe_secrets!
+    # Removes the :secrets key from the action's input and output and saves the action
+    def wipe_secrets!()
       input.delete(:secrets)
       output.delete(:secrets)
+      world.persistence.save_action(execution_plan_id, self)
+    end
+
+    def on_proxy_action_missing
+      error! _('Proxy task gone missing from the smart proxy')
```
iNecas: I guess we should use this method also in `check_task_status`, in case the remote task is missing. The reason:

1. I schedule the job
2. restart the smart proxy dynflow before the connection timeout happens 
3. during the connection timeout, the task fails with 'Error loading data from proxy: RestClient::NotFound - 404 Not Found', instead of saying that "'Proxy task gone missing from the smart proxy'".

I wonder if the `foreman_tasks_proxy_action_start_timeout` actually still makes sense: it would free up som unnecessary events we are emoting at the moment

iNecas:
In the output of the task failed at timeout, I see the 404, but don't see the reason, why that happened, even though in the task details page, I see `Proxy task gone missing from the smart proxy` in errors. That's because `Error loading data from proxy: RestClient::NotFound - 404 Not Found` is still in output. It would be good to get this into the output, so that the users has better insights into what's going on and info about the "task gone missing" is better than 404 IMO

rex-resiliency-output

app/lib/actions/proxy_action.rb
adamruzicka: I dropped the `foreman_tasks_proxy_action_start_timeout` completely

app/lib/actions/proxy_action.rb
adamruzicka: Done

iNecas:
One last thing pending is regarding the output message in the rex terminal.

iNecas:
:policeman:  doesn't seem to be happy

lib/foreman_tasks/cleaner.rb
```diff
@@ -115,6 +116,10 @@ def delete_tasks(chunk)
       tasks.delete_all
     end
 
+    def delete_remote_tasks(chunk)
+      ForemanTasls::RemoteTask.where(:execution_plan_id => chunk.map(&:external_id)).delete_all
```
iNecas: Test failures related, probably related to this tpyo :)

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -0,0 +1,72 @@
...
+      class CheckOnProxyActions; end
+      # Poll the proxy every 10 minutes
+      POLL_INTERVAL = 600
+      BATCH_SIZE = 1000
+
+      def run(event = nil)
+        if event.nil?
+          set_clock
+        elsif event == CheckOnProxyActions
+          check_triggered
```
iNecas: In case `check_triggered` raises expection (such as in case the proxy itself is down, which is something I've tested), the next `CheckOnProxyActions` in not triggered, and we would not check for the presence once the proxy is available again. We should handle this case

iNecas: more importantly, the unhandled exception would actually lead the whole task to fail

app/lib/actions/proxy_action.rb
```diff
@@ -146,22 +168,15 @@ def metadata=(thing)
...
       # Fails if the plan is not finished within 60 seconds from the first task trigger attempt on the smart proxy
       # If the triggering fails, it retries 3 more times with 15 second delays
       { :retry_interval => Setting['foreman_tasks_proxy_action_retry_interval'] || 15,
-        :retry_count    => Setting['foreman_tasks_proxy_action_retry_count'] || 4,
-        :timeout        => Setting['foreman_tasks_proxy_action_start_timeout'] || 60 }
+        :retry_count    => Setting['foreman_tasks_proxy_action_retry_count'] || 4 }
+    end
+
+    def clean_remote_task(*_args)
+      remote_task.destroy!
```
iNecas: I don't know what exactly lead to this, but I got into situation, where this failed with 

```
Failed to run hook 'clean_remote_task' for action 'Actions::ProxyAction'
undefined method `destroy!' for nil:NilClass (NoMethodError)
 | /home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/lib/actions/proxy_action.rb:179:in `clean_remote_task'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan/hooks.rb:68:in `block (2 levels) in run'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan/hooks.rb:66:in `each'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan/hooks.rb:66:in `block in run'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/executors.rb:11:in `run_user_code'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan/hooks.rb:65:in `run'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan.rb:145:in `block in run_hooks'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan.rb:144:in `each'
 | /home/inecas/Projects/ws/foreman-rex/dynflow/lib/dynflow/execution_plan.rb:144:in `run_hooks'
```

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
adamruzicka: I thought that would be handled here https://github.com/theforeman/foreman-tasks/pull/298/files#diff-1316c50efc730290525bdcb5bf816560R66 , but maybe another exception is raised

iNecas: Please give it a try on your setup: there is still some chance there would be something weird on my setup: The error was one of `Errno::ECONNREFUSED, Errno::EHOSTUNREACH, Errno::ENETUNREACH` and despite the fact that it seems it should inherit from `StandardError`, it's not being caught it seems, but I welcome additional try of someone else.

iNecas:
re-running the tests http://ci.theforeman.org/job/test_plugin_pull_request/5017/

db/migrate/20180207150921_add_remote_tasks.foreman_tasks.rb
```diff
@@ -0,0 +1,17 @@
...
+  def change
+    create_table :foreman_tasks_remote_tasks do |t|
+      t.string :execution_plan_id, :null => false
+      t.integer :step_id, :null => false
+      t.string :state, :null => false, :default => 'new'
+
+      t.string :proxy_url, :null => false
+      t.string :remote_task_id
+
+      t.index :execution_plan_id
```
tbrisker: This index isn't needed, all DBs know how to use partial indices so the second index will also allow using just `:execution_plan_id`.

adamruzicka:
I'd say the test failures are unrelated, I've seen those happen randomly on various db-ruby version combinations

iNecas:
Thanks @adamruzicka and @tbrisker

PR #299, opened 2017-11-21
Fixes #21714 - always use CurrentUser middleware

ares:


adamruzicka:
Capturing bits of discussion here:

All and all good idea, but could cause some issues when introduced.

Let's take this scenario: User applies this patch and restarts foreman-tasks service.
What happens is the tasks which weren't originally using the middleware and were running before the restart fail on resume with `Couldn't find User without an ID (ActiveRecord::RecordNotFound)` when trying to restore the current user.

To work around this we could make the middleware first check if there is a `:current_user_id`key in the action's input. On all following executions it would be set properly.

adamruzicka:
@ares what's the status here?

ares:
being busy :-( is there an example of how middleware can access action input? that would help me

adamruzicka:
just by using `action.input`, see https://github.com/theforeman/foreman-tasks/blob/master/app/lib/actions/middleware/keep_current_user.rb#L38

ares:
thanks @adamruzicka, I guess this should do then?

iNecas:
[test]

adamruzicka:
[test]

adamruzicka:
Thanks @ares !

PR #300, opened 2017-11-22
Fixes #21734 - Remove alias_method_chain

xprazak2:


app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -7,8 +7,6 @@ module ActionTriggering
         after_create :plan_hook_action
         after_update :plan_hook_action
         after_destroy :plan_hook_action
```
adamruzicka: Will this block get evaluated when the module gets prepended instead of getting included? If not, this needs to be converted to something like 
```ruby
def self.prepended(base)
  base.after_create :plan_hook_action
  base.after_update :plan_hook_action
  base.after_destroy :plan_hook_action
end
```

xprazak2:
I made changes as suggested

app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -3,12 +3,10 @@ module Concerns
     module ActionTriggering
       extend ActiveSupport::Concern
 
-      included do
-        after_create :plan_hook_action
-        after_update :plan_hook_action
-        after_destroy :plan_hook_action
-
-        alias_method_chain :save, :dynflow_task_wrap
+      def prepended(base)
```
iNecas: I don't think this works as expected: missing `self.prepended`. Try out this.

```ruby
module Foo
  extend ActiveSupport::Concern

  def prepended(base)
     puts "hello from foo"
   end
end

class Bar
   prepend Foo
end
```

app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -3,12 +3,10 @@ module Concerns
     module ActionTriggering
       extend ActiveSupport::Concern
 
-      included do
```
iNecas: Since this is being included from other plugins, I think we should at least raise some expalanationary exception here, warning the users that they should use `prepend` instead of `include`

xprazak2:
I made changes as suggested

mmoll:
This is fully compatible with Rails 5.0, so this could go in right now, pending review (probably a good idea to create a 0.10-stable branch before that).

iNecas:
I guess this will need a rebase in order the tests to work

ehelms:
@iNecas This appears to be green, so what are the tests it's waiting on?

mmoll:
after the master test went green a few hours ago, this PR test got triggered 👍

iNecas:
We were waiting mainly for me finishing with :christmas_tree: Marry Christmas, foreman-tasks-0.11.0 just got pushed to rubygems :)

PR #303, opened 2017-12-13
Revert "Fixes #18673 - Remove initialization code in favor of Dynflow"

iNecas:
This partially reverts commit ecb8c5b39f6281c9c29ff3ea677b3d43002bbea2.

The only thing I've reverted was removing the init scripts themselves
and updated them to use the code, that was moved to foreman/dynflow

### DIFF
--- a/bin/dynflow-executor
+++ b/bin/dynflow-executor
@@ -64,7 +64,7 @@ end
 if $PROGRAM_NAME == __FILE__
   parser = ArgvParser.new(ARGV, $PROGRAM_NAME)
 
-  Dir.chdir(options[:foreman_root])
+  Dir.chdir(parser.options[:foreman_root])
   app_file = File.expand_path('./config/application', parser.options[:foreman_root])
   require app_file
 
--- a/bin/dynflow-executor
+++ b/bin/dynflow-executor
@@ -1,0 +1,72 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+
+class ArgvParser
+  attr_reader :options, :command
+
+  def initialize(argv, file)
+    @options = { foreman_root: Dir.pwd }
+
+    opts = OptionParser.new do |opts|
+      opts.banner = banner(file)
+
+      opts.on('-h', '--help', 'Show this message') do
+        puts opts
+        exit 1
+      end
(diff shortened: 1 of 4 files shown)


bin/foreman-tasks
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
+
+foreman_root = '/usr/share/foreman'
+require File.expand_path('./config/application', foreman_root)
+Dynflow::Rails::Daemon.new.run_background(parser.command, parser.options)
```
adamruzicka: `undefined local variable or method 'parser' for main:Object (NameError)`

iNecas: So much for pushing in hurry… updated

bin/dynflow-executor
```diff
@@ -0,0 +1,71 @@
...
+
+BANNER
+    banner
+  end
+end
+
+# run the script if it's executed explicitly
+if $PROGRAM_NAME == __FILE__
+  parser = ArgvParser.new(ARGV, $PROGRAM_NAME)
+
```
adamruzicka: When requiring foreman's application file it loads additional file relative to the current working directory. We should probably change working directory to `foreman_root` here, otherwise having an option to specify it would make no sense.

iNecas: The thing is, we don't know here, where the foreman actually should be installed: therefore we leave this up to the user of this script. Since this was working this way before, I would prefer just keep it this way, as the main goal here is just to fix the state in 1.16. Any further improvements should go to 1.17 core repo anyway.

iNecas: Ok, updated to make sure we are on the expected directory

bin/dynflow-executor
```diff
@@ -0,0 +1,72 @@
...
+  Dir.chdir(options[:foreman_root])
```
adamruzicka: `undefined local variable or method 'options' for main:Object (NameError)`, I believe it should be `parser.options`

adamruzicka:
It doesn't make sense to wait for the tests, merging.

adamruzicka:
Thanks @iNecas !

PR #307, opened 2017-12-21
pin foreman-tasks-core to matching version

mmoll:


mmoll:
should fix the 0.9.x tests: http://ci.theforeman.org/job/test_plugin_pull_request/4432/

iNecas:
Thanks @mmoll

PR #308, opened 2017-12-29
Fixes #22097 - Adapt to Rails 5 redirect_back

dLobatog:
redirect_to :back is a NoMethodError in the current version of Rails
used by Foreman nightly.


mmoll:
@iNecas ☝️ that would also need a merge now

adamruzicka:
Thanks @dLobatog !

PR #310, opened 2018-01-08
Fixes #22177 - do not set RAILS_RELATIVE_URL_ROOT

jlsherrill:


jlsherrill:
I know this is being changed going forward, and already removed as part of https://github.com/theforeman/foreman-packaging/pull/1549#pullrequestreview-66996485

but it would be nice to fix this in 1.16

adamruzicka:
Not opposed to the change, just curious what supported platform on 1.16 still uses SysV?

iNecas:
Thanks @jlsherrill

iNecas:
Will cherry-pick this to 0.10.x and release a new version, we need to get other updates to 1.16 as well

PR #312, opened 2018-01-19
Fixes #22336 - use `distinct` in has_many definition

iNecas:
Since Rails 5.1, it seems the semantic is a bit different. Also added
tests for this code path.


adamruzicka:
👮 is not happy

iNecas:
Updated

adamruzicka:
Thanks @iNecas !

PR #313, opened 2018-01-28
Fixes #22438 - Remove KeepParams

tbrisker:
KeepParams was created as a workaround for an issue in strong
params prior to Rails 5.1. It is no longer required.


iNecas:
👮 is 👿

iNecas:
Tested and works well, waiting for Jenkins.

tbrisker:
looks like failures was rubocop, pushed again with fix to that

PR #314, opened 2018-02-06
Fixes #22523 - add hover effect to tables

ares:


app/views/foreman_tasks/task_groups/recurring_logic_task_groups/_recurring_logic_task_group.html.erb
```diff
@@ -1,5 +1,5 @@
 <% recurring_logic = task_group.recurring_logic -%>
-<table class='table table-condensed'>
+<table class='<%= table_css_classes('table-condensed') %>'>
```
adamruzicka: I'm not sure if it makes sense here. The table is used just to get an aligned list of key-value pairs. Maybe using a  table for it wasn't the best choice in the first place, but... Anyway, if we need the tables to be consistent everywhere, I won't block this.

image

iNecas: Agreed it doesn't make sense here.

ares: we have the same in host detail page in propertis/nics tabs, why would you prefer to be inconsistent here?

adamruzicka: I somehow missed it being in properties/nics table as well. It doesn't make much sense to me there either, but guess we should stay consistent with core.

PR #317, opened 2018-03-09
Fixes #22838 - avoid double-initialization of Dynflow runtime

iNecas:
Since Foreman 1.17, we're already doing it in core. Repeating this in
here was causing race-conditions around ActiveRecord connection.


adamruzicka:
@iNecas, @ehelms thank you both!

PR #320, opened 2018-03-16
Fixes #22915 - Make the cleaner clean orphaned dynflow plans

adamruzicka:
TODO:
- [ ] i18n support

### DIFF
--- a/lib/foreman_tasks/cleaner.rb
+++ b/lib/foreman_tasks/cleaner.rb
@@ -96,7 +96,7 @@ module ForemanTasks
     # Delete the filtered tasks, including the dynflow execution plans
     def delete
       message = "deleting all tasks matching filter #{full_filter}"
-      with_noop(ForemanTasks::Task.search_for(full_filter), 'task', message) do |source, name|
+      with_noop(ForemanTasks::Task.search_for(full_filter), 'tasks matching filter', message) do |source, name|
         with_batches(source, name) do |chunk|
           delete_tasks chunk
           delete_dynflow_plans chunk
@@ -143,7 +143,7 @@ module ForemanTasks
     end
 
     def delete_orphaned_dynflow_tasks
-      with_noop(orphaned_dynflow_tasks, 'orphaned execution plan') do |source, name|
+      with_noop(orphaned_dynflow_tasks, 'orphaned execution plans') do |source, name|
         with_batches(source, name) do |chunk|
           delete_dynflow_plans_by_uuid chunk.select_map(:uuid)
         end
@@ -154,7 +154,7 @@ module ForemanTasks
     def with_noop(source, name, noop_message = nil)
       if noop
         say '[noop] ' + noop_message if noop_message
-        say "[noop] #{source.count} #{name}(s) would be deleted"
+        say "[noop] #{source.count} #{name} would be deleted"
       else
         yield source, name
       end
@@ -163,7 +163,7 @@ module ForemanTasks
     def with_batches(source, name)
       count = source.count
       if count.zero?
-        say("No #{name}s found, skipping.")
+        say("No #{name} found, skipping.")
         return
       end
       start_tracking_progress(name, count)
@@ -189,7 +189,7 @@ module ForemanTasks
     end
 
     def start_tracking_progress(name, total = tasks.size)
-      say "About to remove #{total} #{name}(s)"
+      say "About to remove #{total} #{name}"
       @current = 0
       @total = total
       say "#{@current}/#{@total}", false if verbose
@@ -201,7 +201,7 @@ module ForemanTasks
     end
 
     def report_done(name)
-      say "Deleted #{@current} #{name}(s)"
+      say "Deleted #{@current} #{name}"
     end
 
     def say(message, log = true)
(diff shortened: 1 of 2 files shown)


test/unit/cleaner_test.rb
```diff
@@ -18,6 +19,8 @@ class TasksTest < ActiveSupport::TestCase
                          end,
                          FactoryBot.create(:dynflow_task, :product_create_task)]
       cleaner.expects(:tasks_to_csv)
+      # Hack to make the tests pass due to ActiveRecord shenanigans
+      cleaner.expects(:delete_orphaned_dynflow_tasks)
```
adamruzicka: `ActiveRecord` doesn't actually save things to the db in tests, so the tests fail when accessing AR data using sequel.

lib/foreman_tasks/cleaner.rb
```diff
@@ -149,21 +182,26 @@ def prepare_filter
...
-        @total = tasks.size
+        @total = total
         say "#{@current}/#{@total}", false
       end
     end
 
     def report_progress(chunk)
       if verbose
-        @current += chunk.size
+        @current += chunk.count
```
iNecas: When testing, I was seeing

```
About to remove 3 orphaned execution plan(s)
Deleted  orphaned execution plan(s)
```
It seems the `@current` was not set in non-berbose mode. I guess, we should track the `@current` always, and only prevent the actual printing in the `verbose` mode.

lib/foreman_tasks/cleaner.rb
```diff
@@ -138,8 +135,44 @@ def with_backup_file(backup_dir, file_name)
...
+      if noop
+        say '[noop] ' + noop_message if noop_message
+        say "[noop] #{source.count} #{name}(s) would be deleted"
+      else
+        yield source, name
+      end
+    end
+
+    def with_batches(source, name)
+      start_tracking_progress(name, source.count)
```
iNecas: Currently, when having an empty set, I get:

```
About to remove 0 orphaned execution plan(s)
Deleted  orphaned execution plan(s)
```

I guess, in this case, we should just mention that no orphaned execution plans were found on go on.

iNecas:
Thanks for coming up with this one. Few comments from testing, but looks promising in general.

There are some cleaner tests failures, that will probably be related.

lib/foreman_tasks/cleaner.rb
adamruzicka: Hope this is better
```
2018-03-16T14:54:25 [I|for|] No tasks found, skipping.                                                                                                                                                                                      
2018-03-16T14:54:25 [I|for|] No orphaned execution plans found, skipping.                                                                                                                                                                   
```

adamruzicka:
[test]

iNecas:
Running tests again http://ci.theforeman.org/job/test_plugin_pull_request/4796/

lib/foreman_tasks/cleaner.rb
```diff
@@ -138,8 +135,49 @@ def with_backup_file(backup_dir, file_name)
...
+        say "[noop] #{source.count} #{name}(s) would be deleted"
+      else
+        yield source, name
+      end
+    end
+
+    def with_batches(source, name)
+      count = source.count
+      if count.zero?
+        say("No #{name}s found, skipping.")
```
iNecas: Passing the arguments as plural + not having to use the additional `s` would be probably a bit more flexible. When I was testing, I got message:

```
No tasks found, skipping
```

It was not clear what tasks those were. Having `name` something like `tasks matching filter` would be more self-describing.

lib/foreman_tasks/cleaner.rb
```diff
@@ -149,19 +187,20 @@ def prepare_filter
...
-      if verbose
-        @current += chunk.size
-        say "#{@current}/#{@total}", false
-      end
+      @current += chunk.count
+      say "#{@current}/#{@total}", false if verbose
+    end
+
+    def report_done(name)
+      say "Deleted #{@current} #{name}(s)"
```
iNecas: With orphaned tasks, I was getting `Deleted 0 orphaned execution plans` even though there has been reported (and deleted` 1 orhphaned execution plan to be deleted

adamruzicka:
Both comments should be addressed

iNecas:
Tests on sqlite still seem to not be happy http://ci.theforeman.org/job/test_plugin_pull_request/4796/, ACK otherwise

test/unit/cleaner_test.rb
```diff
@@ -5,6 +5,8 @@ class TasksTest < ActiveSupport::TestCase
     # To stop dynflow from backing up actions, execution_plans and steps
     ForemanTasks.dynflow.world.persistence.adapter.stubs(:backup_to_csv)
     ForemanTasks::Cleaner.any_instance.stubs(:say) # Make the tests silent
+    # Hack to make the tests pass due to ActiveRecord shenanigans
```
iNecas: I didn't know `shenanigans` is actually a word. Definitely a nice one :)

iNecas:
Thanks @adamruzicka

PR #323, opened 2018-03-21
Fixes #22958 - Do not fail in finalize of if there is no callback

adamruzicka:


lib/foreman_tasks_core/shareable_action.rb
```diff
@@ -15,5 +15,11 @@ def plan(input)
         plan_action(SmartProxyDynflowCore::Callback::Action, callback, planned_action.output)
       end
     end
+
+    private
+
+    def on_proxy?
```
iNecas: We could probably use this on line 14 as well, right?

lib/foreman_tasks_core/runner/action.rb
```diff
@@ -24,7 +24,8 @@ def run(event = nil)
 
       def finalize
         # To mark the task as a whole as failed
-        error! 'Script execution failed' if failed_run?
+        require 'pry'; binding.pry
```
iNecas: Hello `pry`, nice to meet you :) Not that I don't like you, but please, leave this PR before I call :policeman: :)

iNecas:
I guess this was the reason why @bastilian saw the `finalize` method not working for him, right?

bastilian:
I've been testing this and see odd behavior. When a job is run it starts as it should, but the host jobs stay in the running status (at least in the UI) and the task stays "pending". When I restart Foreman though the jobs do appear as "failed".

bastilian:
This also seems to break Ansible jobs, which appear to not get any status reported.

lib/foreman_tasks_core/shareable_action.rb
adamruzicka: Done

adamruzicka:
[test]

iNecas:
Running tests at http://ci.theforeman.org/job/test_plugin_pull_request/5013/

PR #324, opened 2018-03-21
Fixes #22969: Allow tasks to use Org.current

akofink:
Basically copied verbatim from https://github.com/theforeman/foreman-tasks/blob/master/app/lib/actions/middleware/keep_current_user.rb

### DIFF
--- a/app/lib/actions/middleware/keep_current_taxonomies.rb
+++ b/app/lib/actions/middleware/keep_current_taxonomies.rb
@@ -1,0 +1,64 @@
+#
+# Copyright 2014 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public
+# License as published by the Free Software Foundation; either version
+# 2 of the License (GPLv2) or (at your option) any later version.
+# There is NO WARRANTY for this software, express or implied,
+# including the implied warranties of MERCHANTABILITY,
+# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
+# have received a copy of GPLv2 along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+
+module Actions
+  module Middleware
+    class KeepCurrentTaxonomies < Dynflow::Middleware
+      def delay(*args)
+        pass(*args).tap { store_current_taxonomies }
+      end
+
+      def plan(*args)
+        with_current_taxonomies do
+          pass(*args).tap { store_current_taxonomies }
+        end
+      end
+
+      def run(*args)
+        restore_current_taxonomies { pass(*args) }
+      end
+
+      def finalize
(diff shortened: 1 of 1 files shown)


app/lib/actions/middleware/keep_current_organization.rb
```diff
@@ -0,0 +1,56 @@
...
+# 2 of the License (GPLv2) or (at your option) any later version.
+# There is NO WARRANTY for this software, express or implied,
+# including the implied warranties of MERCHANTABILITY,
+# NON-INFRINGEMENT, or FITNESS FOR A PARTICULAR PURPOSE. You should
+# have received a copy of GPLv2 along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+
+module Actions
+  module Middleware
+    class KeepCurrentOrganization < Dynflow::Middleware
```
iNecas: What about using `KeepCurrentTaxonomies` and storing both org and location in the context?

iNecas:
Maybe this request will be a bit bold, but would you mind adding some test case (inspiration can be taken from https://github.com/theforeman/foreman-tasks/blob/master/test/unit/actions/proxy_action_test.rb) to actually cover the behavior? Bonus points for adding that also for `KeepCurrentUser`, which I guess would be quite simple to do, once this is done. And sorry for not enforcing this in the past, I realize it would make the life easier for folks now.

akofink:
Thanks for checking me. :) How could I submit a PR without tests! :see_no_evil:

akofink:
This PR is blocking some other work in Katello. Would you mind if I add middleware tests in a separate PR?

iNecas:
Ok, setting recurring reminder to ping you to add the tests later :) Merged as 8e9e7daaac397a4a6dca3c91a046926dc6fbae42,  foreman-tasks 0.12.2 pushed to rubygems

mmoll:
```
app/lib/actions/middleware/keep_current_taxonomies.rb:46:11: C: Naming/PredicateName: Rename has_current_taxonomies? to current_taxonomies?.
    def has_current_taxonomies?
        ^^^^^^^^^^^^^^^^^^^^^^^
app/lib/actions/middleware/keep_current_taxonomies.rb:48:9: C: Layout/MultilineOperationIndentation: Use 2 (not 0) spaces for indenting an expression spanning multiple lines.
    (Location.current || action.input[:current_location_id].nil?)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

akofink:
Huh - I guess we should run rubocop on PRs in this repo.

iNecas:
Lessons learned, let's address that in https://github.com/theforeman/foreman-tasks/pull/325

PR #330, opened 2018-04-12
Fixes #23244 - Migrate two-pane to breadcrumb-bar

sharvit:


adamruzicka:
Could something be done about this?
image

sharvit:
@adamruzicka I could not reproduce it locally with chrome or firefox.
Can you think about any differences between our steps or environments?

Thanks!

app/views/foreman_tasks/tasks/show.html.erb
```diff
@@ -1,5 +1,20 @@
...
+  items: [
+    {
+      caption: _('Tasks'),
+      url: url_for(foreman_tasks_tasks_path)
+    },
+    {
+      caption: truncate(format_task_input(@task), :length => 50)
+    }
+  ],
+  name_field: 'label',
```
tstrachota: @sharvit could you use `humanized.action` instead of `label`?
The camel-case action names in the switcher list are not very readable. It's inconsistent with what we display in the breadcrumbs too.

sharvit: There are two ways we can achieve that:

1. Expand the core logic so we can pass `name_field: 'humanized.action'` and nested export it from the object.

2. Expand the plugin so it will render a new field `title?` at the root level of the json.  :+1: 

Any thoughts? @tstrachota @iNecas

tstrachota: I tend to prefer updating the breadcrumbs logic. Similar problem can appear elsewhere too.

Maybe more flexible alternative to parsing a path to the field could be using a mapping function. Something like this:
```js
nameField: record => record.humanized.action
```
But I'd still keep the possibility to use a string for the simple cases.

sharvit: @tstrachota I don't think it's trivial to use a ruby-mapper inside javascript. Or put a js-mapper inside a ruby hash.

I will implement a nested extraction at the core and will update this PR soon.

sharvit:
@tstrachota I updated the `name_field` to be `humanized.action` and created a PR in core to support it.
https://github.com/theforeman/foreman/pull/5534

iNecas:
So I guess this is waiting for the Foreman PR to be merged first, right?

tstrachota:
The core dependency has been merged. It works well and the switcher items have now much more readable names.

I found one trouble not directly related to this PR, but making the ux of the switcher worse: When a running task is displayed the page keeps refreshing in some interval. That made it very difficult to use the switcher, because it was constantly being closed before I managed to click on an item. One solution could be not to reload the whole page but load it async and replace just the relevant part of dom.

sharvit:
Nice catch @tstrachota :+1:  It does make the experience worth :-1: 

Do you think it should actually block this PR?

tstrachota:
I have mixed feelings about it. I makes the switcher nearly unusable for running tasks. On the other hand it's just the switcher that becomes out of the game. Users can still use the breadcrumbs to navigate back and select another task. We know that we want this code in eventually. So I would say don't block this PR and fix the other issue separately but with priority.
@iNecas what do you think?

tstrachota:
Some additional info:
- the code that reloads the whole page is here: https://github.com/theforeman/foreman-tasks/blob/master/app/views/foreman_tasks/tasks/show.html.erb#L18
- it should be replaced with code similar to our old two-pane https://github.com/theforeman/foreman/blob/develop/app/assets/javascripts/two-pane.js#L92-L110 (obviously some changes are necessary)
- it should be enough just to replace the `.task-details` div with the updated version returned from the server https://github.com/theforeman/foreman-tasks/blob/master/app/views/foreman_tasks/tasks/show.html.erb#L91

iNecas:
I think that its a regression and it should be figured out before merging

app/views/foreman_tasks/tasks/index.html.erb
```diff
@@ -2,20 +2,7 @@
 <% title_actions csv_link, help_button %>
 <% stylesheet 'foreman_tasks/tasks' %>
 
-<script>
-
-var currentTwoPaneTask;
-
-$(document).on('click', ".table-two-pane td.two-pane-link", function(e) {
-  currentTwoPaneTask = $(this).find("a");
```
tstrachota: I found out that this var is used in the show template. With removing it, this https://github.com/theforeman/foreman-tasks/blob/master/app/views/foreman_tasks/tasks/show.html.erb#L10-L17 becomes dead code.

app/views/foreman_tasks/tasks/show.html.erb
```diff
@@ -1,94 +1,117 @@
...
+    }
+  ],
+  name_field: 'action',
+  resource_url: foreman_tasks_api_tasks_path,
+  switcher_item_url: foreman_tasks_task_path(:id => ':id')
+) %>
+
+
+<div class="task-details">
+  <script>
```
sharvit: It's showing such a big diff but I just moved the script into `.task-details` and updated `taskProgressReloader.reload`.

app/views/foreman_tasks/api/tasks/show.json.rabl
```diff
@@ -1,5 +1,5 @@
 object @task if @task
 
-attributes :id, :label, :pending
+attributes :id, :label, :pending, :action
```
sharvit: I followed the `task_label` from:
https://github.com/theforeman/foreman-tasks/blob/master/app/views/foreman_tasks/tasks/_details.html.erb#L92

It seems the end result is the action field so the new strategy is to add it to the json result.

sharvit:
@tstrachota great review :+1: 

I followed your strategy and it works well. Can you have another look?

app/views/foreman_tasks/api/tasks/show.json.rabl
tstrachota: Just to have complete picture, @adamruzicka what is the difference between the `action` and `humanized.action` fields, please?

adamruzicka: Each action can define a human readable name, this is `humanized.action` and is defined on an action class as `#humanized_name`.  Just `action` is generated from various parts of `humanized`.

For example for `Actions::RemoteExecution::RunHostJob` action:

```ruby
pry(main)> task.humanized[:action]
=> "Remote action:"
pry(main)> task.humanized
=> {:action=>"Remote action:", :input=>"Run true on stub-100", :output=>"»Exit status: 0«", :errors=>[]}
pry(main)> task.action
=> "Remote action: Run true on stub-100"
```

app/views/foreman_tasks/tasks/show.html.erb
sharvit: Sorry i forgot to mention it.
I replaced the call to `two_pane_open` with an `ajax` call,
so instead doing full reload, i'm getting from the server only the `div.class-details` and replace the current one by the new one.
We need to have the script inside the `div.class-details` so it will get reloaded when the div replaced.
It mostly needed to reapply the event handlers.

tstrachota:
Reloading works as a charm now, good job @sharvit !

I found out that the pagination is always disabled in the switcher. It's because of similar issue as we have with the hosts api in core. `subtotal` always equals `per_page`. See https://github.com/theforeman/foreman-tasks/blob/master/app/controllers/foreman_tasks/api/tasks_controller.rb#L138 for details. Hopefully it won't be difficult to fix so I'd vote for fixing it as part of this PR (I'm sorry for not spotting it earlier).

app/views/foreman_tasks/api/tasks/show.json.rabl
tstrachota: Got it, thanks.

adamruzicka:
@sharvit What's the status here?

sharvit:
@adamruzicka There is just one small issue @tstrachota found https://github.com/theforeman/foreman-tasks/pull/330#issuecomment-392771567

I will probably find some time for it on the next few days.

sharvit:
@tstrachota @adamruzicka, I fixed the pagination issue, you can see the diff:
https://github.com/theforeman/foreman-tasks/pull/330/files#diff-76e220fa2f3fb075d1aab086606ce317

Can you have another look please?

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -118,7 +118,9 @@ def bulk_resume
       end
       def index
```
adamruzicka: When you're doing changes to the index, wouldn't it make sense to use the [index layout](https://github.com/theforeman/foreman/blob/develop/app/views/api/v2/layouts/index_layout.json.erb)?

iNecas: I would be ok with leaving this for separate refactoring issue.

PR #331, opened 2018-04-12
Fixes #23186 - ensure we don't get stuck in interactive mode

iNecas:


iNecas:
@ekohl mind giving this patch a try?

iNecas:
What about now?

lib/foreman_tasks_core/runner/command_runner.rb
```diff
@@ -23,18 +23,25 @@ def refresh
...
 
       def close_io
-        @command_out.close if @command_out && !@command_out.closed?
-        @command_out = nil
-
-        @command_in.close if @command_in && !@command_in.closed?
-        @command_in = nil
+        [@command_out, @command_in, @input_stream].each do |stream|
+          stream.close if stream && !stream.closed?
+          stream = nil
```
adamruzicka: This modifies the local binding, not the instance variable

iNecas: Nice catch, updated

adamruzicka:
Code looks good, will try tomorrow

iNecas:
Bump

adamruzicka:
Sorry, I take that back, I made a mistake when reproducing the issue. But I noticed an interesting thing

```
# pgrep -f ansible-playbook
30445

# ls -l /proc/30445/fd | grep null
lr-x------. 1 aruzicka aruzicka 64 Apr 24 11:31 0 -> /dev/null
```

When I trigger the bug, the ansible-playbook process has `/dev/null` as its STDIN and yet it still asks for the confirmation on the console

ekohl:
> So even though we close the STDIN for the ansible-playbook process, it forks into ssh which apparently just grabs the terminal

So even using a PTY in the first place is incorrect and we should be using a regular process to run in? From https://serverfault.com/questions/593399/what-is-the-benefit-of-not-allocating-a-terminal-in-ssh/593419#593419
> If a PTY is allocated, applications can detect this and know that it's safe to prompt the user for additional input without breaking things. There are many programs that will skip the step of prompting the user for input if there is no terminal present, and that's a good thing. It would cause scripts to hang unnecessarily otherwise.

iNecas:
The reason why we went with PTY in the first place to get the output in the same format (formatting and color). Without PTY, the output is not that readable.

adamruzicka:
We may want to carry this over to smart_proxy_dynflow eventually. Closing for now

PR #332, opened 2018-04-20
Fixes #23348 - Set label field properly

adamruzicka:
`self.label` invoked the method so `||=` never happend and the label was always calculated on demand


iNecas:
Running the tests manually http://ci.theforeman.org/job/test_plugin_pull_request/4997/

adamruzicka:
The tests are green

iNecas:
Thanks @adamruzicka . Merging now

PR #333, opened 2018-04-22
Fixes #23355 - Replace alias_method_chain with prepend

tbrisker:
This was most likely missed becuase the extention module is only
included when `FOREMAN_TASKS_MONKEYS` env variable is set to true.
The module has been updated to use the new module#prepend syntax.


iNecas:
Thanks @tbrisker

PR #336, opened 2018-05-08
Fix dynflow-executor task in README

dLobatog:
The task name has changed since it was moved to core


iNecas:
Thanks. Btw. be aware of this issue, we have a path ready for testing in dynflow: without it, the executor from rake would not work properly in develpment. https://github.com/Dynflow/dynflow/pull/283

iNecas:
Test failures unrelated

PR #337, opened 2018-05-14
Package name change?

tek0011:
I believe the package name has changed per yum search and http://yum.theforeman.org/plugins/1.17/el7/x86_64 to:
tfm-rubygem-foreman-tasks


iNecas:
You're right. Thanks for the update!

PR #338, opened 2018-05-24
readme: add compatibility chart

timogoebel:
... I hope, I got the versions right. Please double check.


ares:
:heart:

adamruzicka:
In the yum plugin repos for 1.16 we have tasks-0.9.1 and tasks-0.10.0 which doesn't really match `~> 0.10.0`. Do we need to capture that in the table somehow?

timogoebel:
> Do we need to capture that in the table somehow?

I guess that's actually a leftover from how we handle packaging. We should be on the save side with the current version. 👼

iNecas:
Thanks @timogoebel!

PR #339, opened 2018-05-26
Fixes #22958 - Do not fail in finalize of if there is no callback

mmoll:
(cherry picked from commit 329b0a2683bca49f74afbd062cd03999e34109f4)

### DIFF
--- a/lib/foreman_tasks_core/runner/action.rb
+++ b/lib/foreman_tasks_core/runner/action.rb
@@ -24,7 +24,7 @@ module ForemanTasksCore
 
       def finalize
         # To mark the task as a whole as failed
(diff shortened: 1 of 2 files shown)


mmoll:
no idea what's going in wit hthe tests there... however, I hope that this fixes the ABORTed builds we're seeing on 0.11.x still

iNecas:
I think that will be due to the fact, that the tests were run against develop, while we need 1.17-stable. Trying to run manually here http://ci.theforeman.org/view/All/job/test_plugin_matrix/5366/

mmoll:
thanks, the one test did hang again, so it was not this commit...

PR #342, opened 2018-06-05
[WIP] Fixes #XXXXX - Associate tasks with group on start

jlsherrill:


adamruzicka:
The task should associate the recurring logic's task group when it gets `#delay`ed (see [the recurring logic middleware](https://github.com/theforeman/foreman-tasks/blob/master/app/lib/actions/middleware/recurring_logic.rb#L13)). This was done to have a uniform way of creating the association between the task and the recurring logic on the first and all subsequent repetitions

jlsherrill:
Thanks for the explanation @adamruzicka i wasn't including that middleware at that time.  Even afterwards it wasn't working, but after switching dev envs its all working now!  I'll keep an eye on it.

PR #344, opened 2018-06-07
Fixes #23585 - allow future start date for recurring tasks

jlsherrill:


iNecas:
Thanks @jlsherrill and @adamruzicka

PR #349, opened 2018-06-20
Refs #23837: incl `update` in actn_trggr methods

akofink:
`update_attributes` is an alias for `update` in Rails 5. We no longer support Rails 4, so we should use the new update methods.

### DIFF
--- a/app/models/foreman_tasks/concerns/action_triggering.rb
+++ b/app/models/foreman_tasks/concerns/action_triggering.rb
@@ -35,13 +35,19 @@ module ForemanTasks
         dynflow_task_wrap(:destroy) { super }
       end
 
-      def update_attributes(*args)
+      # In order to use host.<attribute>_changed?, we must assign_attributes to
+      # the host record for these update and update! methods.
+      def update(*args)
+        assign_attributes(*args)
         dynflow_task_wrap(:save) { super(*args) }
       end
+      alias update_attributes update
 
-      def update_attributes!(*args)
+      def update!(*args)
+        assign_attributes(*args)
         dynflow_task_wrap(:save) { super(*args) }
       end
+      alias update_attributes! update!
 
       # this can be used when HostActionSubject is used for orchestration but you want to avoid
       # triggering more tasks by ActiveRecord callbacks within run/finalize phase of your task


adamruzicka:
Nope, we don't support Rails 4 anymore

akofink:
Updated to remove Rails 4 update methods.

app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -35,11 +35,11 @@ def destroy
         dynflow_task_wrap(:destroy) { super }
       end
 
-      def update_attributes(*args)
+      def update(*args)
```
iNecas: I think we should still do aliasing here:

```
alias_method :update_attributes, :update
```

It seems the `update_attributes` is not actually deprecated in Rails and calling `obj.update_attributes` doesn't trigger this method.

akofink: hrm... I would have thought https://github.com/rails/rails/blob/375a4143cf5caeb6159b338be824903edfd62836/activerecord/lib/active_record/persistence.rb#L432 would take care of this. :man_shrugging:

akofink:
Updated with alias_methods

app/models/foreman_tasks/concerns/action_triggering.rb
iNecas: That's how ruby works unfortunately. Seems like rubocop prefers `alias` over `alias_method` (don't ask me why :)

akofink:
Let's please wait to merge this. Just in case we have more changes related to https://github.com/Katello/katello/pull/7427

iNecas:
Ok, let me know once I can hit the button :)

app/models/foreman_tasks/concerns/action_triggering.rb
```diff
@@ -35,13 +35,17 @@ def destroy
...
+        self.attributes = args[0]
```
iNecas: Why? It's not intended here to do anything else than the `dynflow_task_wrap`. Unless it's really related to the functionality of the `ActionTriggering`, it should be somewhere else.

akofink: This does fix [#23837](https://projects.theforeman.org/issues/23837), and here's why: we are using `host.<attribute>_changed?` [here](https://github.com/Katello/katello/blob/master/app/models/katello/host/subscription_facet.rb#L253) to check whether we need to trigger the `update_action`. Those methods do not work if you use `host.update(attributes)` because the attributes aren't updated in memory on the host, so the method always returns false. Instead, we must use `host.attributes = attributes; host.save`.

This controller works without this line: [app/controllers/api/v2/hosts_controller.rb#L161](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L161)
This one doesn't: [app/controllers/hosts_controller.rb#L122](https://github.com/theforeman/foreman/blob/develop/app/controllers/hosts_controller.rb#L122)

This is **one** solution, but it seems pretty hacky to me. I'd welcome other suggestions. :)

iNecas: What about doing this in https://github.com/theforeman/foreman/blob/develop/app/controllers/hosts_controller.rb#L122? It seems like a Rails bug, but we should update the usage of the `@host.update` in Foreman, if the `host.<attribute>_changed?` doesn't work with the `update` method.

```ruby
@host.attributes = attributes
if @host.save
  #...
```

akofink: I'm fine with changing it in Foreman. What's one more PR for this issue? ;)

akofink: Though, if it's not a bug in Rails, we may want to include some more documentation with this ActionTriggering module to inform people that it must be used with the two-step save. Putting the attributes on the host here in this repo removes that usage caveat.

iNecas: Oh, I understand the issue now: I agree this is actually an issue here then.

I would prefer avoiding calling args[0] in favor of usign *args with `assign_attributes`
```
assign_attributes(*args)
```

Also, please explanatory comment why we are doing this in the first place, for future us still being able to understand why.

akofink:
Updated! Let me know if the comment is OK.

iNecas:
One thing: the rubocop is complaining about the redundant `self`

akofink:
OK! This is ready for merge. The Katello PR is failing just as it should without this change in place.

http://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/2099/pipeline/#step-88-log-125

iNecas:
Thanks @akofink, foreman-tasks-0.13.3 has just been pushed  to rubygems

akofink:
Thanks for your help, @iNecas! Packaging PR for nightly: https://github.com/theforeman/foreman-packaging/pull/2711

PR #350, opened 2018-06-26
Fixes #23888 - support enable/disable recurring logic

jlsherrill:


iNecas:
So far looks good: have not tested yet. Mind covering this with some unit tests?

jlsherrill:
```
having frozen tasks marked in the UI
```
Is the best way to do this be to go through its task groups to lookup its recurring logic?  It seems like a task could have multiple (although i'm not sure thats possible).  Is it safe to assume one?

adamruzicka:
Yeah, just looking at the delayed plan should be enough

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -65,6 +65,14 @@ def execution_plan(silence_exception = true)
       nil
     end
 
+    def frozen
+      delayed_plan.try(:frozen)
+    end
+
+    def delayed_plan
+      ForemanTasks.dynflow.world.persistence.load_delayed_plan(external_id)
```
adamruzicka: How about returning nil early if the task is not in scheduled state so we won't hit the db 20 times each time a user opens the task index just to see that there are no delayed plans for the tasks?

app/helpers/foreman_tasks/foreman_tasks_helper.rb
```diff
@@ -52,14 +55,16 @@ def duration_in_words_span(start, finish)
...
+        buttons << link_to(N_('Cancel'), cancel_foreman_tasks_recurring_logic_path(recurring_logic), :method => :post, :class => '') unless recurring_logic.done?
       end
-      button_group buttons
+      action_buttons buttons
     end
 
     def recurring_logic_next_occurrence(recurring_logic)
       default = '-'
-      return default if %w[finished cancelled].include? recurring_logic.state
+      return default if recurring_logic.done? || !recurring_logic.enabled?
```
adamruzicka: Why not use `recurring_logic.disabled?`

app/controllers/foreman_tasks/recurring_logics_controller.rb
```diff
@@ -1,13 +1,23 @@
...
     def show; end
 
+    def enable
+      @recurring_logic.update!(:enabled => true)
+      redirect_to :action => :index
+    end
+
+    def disable
+      @recurring_logic.update!(:enabled => false)
+      redirect_to :action => :index
```
adamruzicka: Clicking disable/enable when looking at recurring logic details redirects you to the index. I think the it should stay on the same page, but on the other hand cancel behaves the same way. Guess we can be consistent first and fix this behavior in another PR

jlsherrill:
@adamruzicka fixed the nitpicks and added an error when enabling a cancelled task.

jlsherrill:
Ah right, good call, updated!

adamruzicka:
Dynflow 1.1.0 with the needed changes was just released, please bump the dependency here

jlsherrill:
[test foreman-tasks]

jlsherrill:
These failures look unrelated?   I can reproduce locally but even on master.

iNecas:
[test foreman-tasks] the failures should be fixed now with https://github.com/theforeman/foreman/pull/5823

jlsherrill:
@iNecas Thanks! its :green_heart:

iNecas:
@adamruzicka no other comments from me

adamruzicka:
Thanks @jlsherrill !

PR #352, opened 2018-07-30
Fixes 24479 - fix csv tasks export

iNecas:


app/models/foreman_tasks/task.rb
```diff
@@ -85,8 +85,26 @@ def execution_type
       delayed? ? N_('Delayed') : N_('Immediate')
     end
 
+    def get_humanized(method)
```
iNecas: Needed to define this for tests to work properly: we've already seen this issue in Katello, so I though it's better to fix it right way.

test/controllers/tasks_controller_test.rb
```diff
@@ -24,6 +24,14 @@ def in_taxonomy_scope(organization, location = nil)
         Organization.current = Location.current = nil
       end
 
+      it 'supports csv export' do
```
iNecas: This should have been done from the beginning of the feature: we would have found the issue much faster than without it.

app/models/foreman_tasks/task.rb
```diff
@@ -85,8 +85,26 @@ def execution_type
...
+      attr = case method
+             when :humanized_name
+               :action
+             when :humanized_input
+               :input
+             when :humanized_output
+               :output
+             else
+               nil
```
adamruzicka: `case` will implicitly return `nil` if none of the branches match. That's what rubocop complains about

iNecas:
Updated

adamruzicka:
Test failure seems unrelated, ACK

adamruzicka:
Thanks @iNecas !

PR #353, opened 2018-07-31
Fixes #24493 - fix double creation of resources on host.update

iNecas:
The issue is the `assign_attributes` can have already some side-effects
for sub-resources (such as parameters), and calling `super` causes
double-creation of the resources.

Changing super to `save` makes sure that we avoid this double creation.


iNecas:
The tests are green

adamruzicka:
Thanks @iNecas !

PR #354, opened 2018-07-31
remove test_attributes

mmoll:
1.18.x doesn't support these


mmoll:
this is against 0.13.x

iNecas:
Thanks @mmoll!

PR #355, opened 2018-08-11
Fixes #23355 - Replace alias_method_chain with prepend

bjozet:
This was most likely missed becuase the extention module is only
included when FOREMAN_TASKS_MONKEYS env variable is set to true.
The module has been updated to use the new module#prepend syntax.

(cherry picked from commit 226d4523ce1fce95d33f75b243b7998d67c1952b)


tbrisker:
ok to test

iNecas:
Thanks @bjozet

iNecas:
The tests fail as we don't have currently option in Jenkins that I know of to run tests against older releases, but I'm confident this will not cause any test failures there.

PR #357, opened 2018-09-06
Fixes #24830 - Improve search by user

adamruzicka:


iNecas:
Combination of `~ *` with `.id` (such as `user.id ~ skoro*`) fails with:

```
ActiveRecord::StatementInvalid in ForemanTasks::Tasks#index
PG::UndefinedFunction: ERROR:  operator does not exist: integer ~~* unknown
LINE 4: ...ERE ((foreman_tasks_locks_ownerff71de.resource_id ILIKE 'sko...
                                                             ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
: SELECT  DISTINCT "foreman_tasks_tasks".* FROM "foreman_tasks_tasks" INNER JOIN foreman_tasks_locks AS foreman_tasks_locks_ownerff71de
                 ON (foreman_tasks_locks_ownerff71de.task_id = foreman_tasks_tasks.id AND
                     foreman_tasks_locks_ownerff71de.resource_type = 'User' AND
                     foreman_tasks_locks_ownerff71de.name = 'task_owner') WHERE ((foreman_tasks_locks_ownerff71de.resource_id ILIKE 'skoro%')) ORDER BY "foreman_tasks_tasks"."started_at" DESC NULLS LAST  LIMIT $1 OFFSET $2
```

adamruzicka:
This is interesting, when I use `owner.id ~ 4` I get an validation error, if I use `user.id ~ 4` which should behave the same way, the action is allowed to proceed and crashes with the error you mentioned. I'll look into adding more validations.

screenshot_20180907_105813

iNecas:
Test failures related

test/unit/task_test.rb
```diff
@@ -21,6 +21,19 @@ class TasksTest < ActiveSupport::TestCase
     test 'can search the tasks by current_user in combination with implicit search' do
       assert_equal [@task_one], ForemanTasks::Task.search_for("owner.id = current_user AND #{@task_one.label}")
     end
+
+    test 'can search the tasks by user' do
```
iNecas: Since we have a specific brach in the logic for the `id`, it would be good to add some test around searching based on id. No other comments: tested and it worked well.

adamruzicka: Added a test for that

iNecas:
Thanks @adamruzicka

PR #361, opened 2018-09-12
Fixes #24909 - Mark proxy tasks as failed only if confirmed

adamruzicka:
Before this changed we marked tasks as failed if we failed to check on
their counterparts on the smart proxy. With this change, tasks are
marked as failed only if the check succeeds.


iNecas:
Tested by:

1. triggering the job
2. killing the proxy as it was running (to simulate the host being unreachable)
3. the tasks were marked as running until I've started the process again, so that the proxy returned 200 again - this time the tasks were marked as failed, as they were actually lost during the restart.

Thanks @adamruzicka

PR #363, opened 2018-09-24
Fixes #25017 - Support all operators for search tasks by user

adamruzicka:

PR #364, opened 2018-09-27
Fixes #21083  - Link tasks to users explicitly

adamruzicka:


adamruzicka:
This will probably deprecate https://github.com/theforeman/foreman-tasks/pull/363

iNecas:
Test failures related

app/models/foreman_tasks/lock.rb
```diff
@@ -120,7 +120,7 @@ def link?(resource, uuid)
 
       # Records the information about the user that triggered the task
       def owner!(user, uuid)
-        build_owner(user, uuid).save!
```
iNecas: Any reason to keep this around https://github.com/theforeman/foreman-tasks/blob/4aa960623bfb19dff930b96efcf23bab7019a90b/lib/foreman_tasks/dynflow/persistence.rb#L39 ?

adamruzicka: I can't think of any, just an oversight on my side

adamruzicka:
test failures seem unrelated

test/factories/task_factory.rb
```diff
@@ -6,11 +6,11 @@
     result 'success'
 
     transient do
-      set_owner nil
+      lock_owner nil
     end
 
     after(:create) do |task, evaluator|
-      ForemanTasks::Lock.owner!(evaluator.set_owner, task.id) if evaluator.set_owner
+      ForemanTasks::Lock.owner!(evaluator.lock_owner, task.id) if evaluator.lock_owner
```
iNecas: Is the lock_owner used anywhere? If not, shouldn't we just remove it?

adamruzicka: I briefly grep'd through the code and some plugins and couldn't find it (or the original variant) being uesd anywhere. Dropped

app/models/foreman_tasks/lock.rb
```diff
@@ -120,7 +120,7 @@ def link?(resource, uuid)
...
+        # TODO: Log some deprecation warning?
```
iNecas: Can we drop this method all together with other unused code or still the todo is valid?

db/migrate/20180927120509_add_user_id.foreman_tasks.rb
```diff
@@ -0,0 +1,40 @@
...
+        user_locks.where(:resource_id => owner_id).delete_all
+      end
+    end
+  end
+
+  def down
+    User.as_anonymous_admin do
+      ForemanTasks::Task.select(:user_id).distinct.pluck(:user_id).compact.each do |user_id|
+        ForemanTasks::Task.where(:user_id => user_id).select(:id).find_in_batches do |group|
+          group.each { |task| create_lock user_id, task } # TODO: Bulk insert would be nice
```
iNecas: Given the probability of the down scenario, I'm ok with removing this TODO, as it's probably not worth doing it.

iNecas:
I have not tested yet, but code-wise looks good. I would like to wait until the branch date of Foreman 1.20: it's still relatively big change and I would be more comfortable having this in nightlies for a while before we go GA with it.

adamruzicka:
Agreed, let's put this on hold for a while

adamruzicka:
[test foreman-tasks]

ares:
:clap:

adamruzicka:
The tests are green, let's put this on hold until all the 1.20 candidates are dealt with

ares:
needs rebase

iNecas:
Release the kraken!

PR #365, opened 2018-09-30
Fixes #25076 - fix importing facts

ares:


ares:
actually I can keep the action input key, ignore me :-)

PR #366, opened 2018-09-30
Fixes #25078 - fixes importing facts

ares:


iNecas:
Given we default to `Host::Managed` in Foreman codebase as well (https://github.com/theforeman/foreman/blob/526e302fc69d8d1db706aa3dcf67766161aaf0d4/app/controllers/api/v2/hosts_controller.rb#L374), I'm ok with this change

iNecas:
Thanks @ares

PR #368, opened 2018-10-03
Fixes #25109 - add timeout for sync_task

iNecas:


lib/foreman_tasks.rb
```diff
@@ -27,8 +27,15 @@ def self.trigger_task(async, action, *args, &block)
...
-              ForemanTasks::Task::DynflowTask.where(:external_id => id).first!
+              unless async
+                timeout = Setting['foreman_tasks_sync_task_timeout']
+                finished.wait(timeout)
+                task = ForemanTasks::Task::DynflowTask.where(:external_id => id).first
+                if task.nil? || task.pending?
+                  raise TimeoutError.new("The time waiting for task #{task.try(:id)} to finish exceeded the 'foreman_tasks_sync_task_timeout' (#{timeout}s)")
+                end
+                task
+              end
```
adamruzicka: If I'm reading this right, this will return `nil` if `async` is `true`

iNecas: good catch, updated

iNecas:
Updated

adamruzicka:
Thanks @iNecas !

PR #369, opened 2018-10-09
Fixes #25160 - Trigger tasks on the smart proxy in batches

adamruzicka:
- [x] requires https://github.com/theforeman/smart_proxy_dynflow/pull/55 on the SPDC side


app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -7,14 +7,14 @@ class CheckOnProxyActions; end
...
-          set_clock
-        elsif event == CheckOnProxyActions
+        if event == CheckOnProxyActions
           check_triggered
           set_clock
           action.send(:suspend)
         end
         pass event
+      ensure
+        batch_trigger_pending if event.nil? || event.is_a?(Dynflow::Action::WithBulkSubPlans::PlanNextBatch)
```
adamruzicka: I'll probably try to find a better place for this

lib/foreman_tasks_core/runner/parent_action.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      # { task_id => { :action_class => Klass, :input => input } }
+      def plan(callback, input_hash)
+        results = input_hash.reduce({}) do |acc, (key, value)|
+          # This is ugly, but we need the IDs
+          result = trigger(::Dynflow::Utils.constantize(value['action_class']),
+                           value['action_input'].merge(:callback_host => callback))
+          acc.merge(key => format_result(result))
+        end
+        plan_self :result => results
```
iNecas: let's use `()` here

lib/foreman_tasks_core/runner/parent_action.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def initiate
+        ping suspended_action
+        wait_for_sub_plans sub_plans
```
iNecas: and let's use `()` here :)

iNecas:
Was not doing full review just yet (except some nitpicking about parenthesis:) but tried to understand the concept and I like the direction it goes.

I wonder if we should keep both batch and non-batch ways there… but probably would be good to have some measurements first before we drop the non-batch solution.

Also, since we have some time, would be good to invest some of it to cover this with more tests. I seems we don't have any coverage of remote tasks in tests right now: would be good to fill in this gap together with this improvement.

iNecas:
How much close to working state this is, from 0 (doesn't work and breaks your computer) to 10 (everything tested and works without any other work needed)?

adamruzicka:
The happy path in this part is pretty solid, other paths not so much. At least better error handling is needed and test coverage would be nice, overall right now I'd give this 4/10.

I have other, yet unpublished, changes which add scaffolding for the next stage, where the parent can run a single process for all the target hosts and distribute outputs to the child (per-host) actions.  This brings the overall readiness to about 2.5-ish.

Also because things are getting more and more complicated I'd like to spend some time on documentation for this

lib/foreman_tasks_core/runner/action/parent.rb
```diff
@@ -0,0 +1,68 @@
...
+          end
+          { :targets => result }
+        end
+
+        def use_group_runner?
+          !group_runner_action_class.nil?
+        end
+
+        def group_runner_action_class
+          nil
```
iNecas: I think we were talking about some register, that would return the group runner by feature, weren't we?

lib/foreman_tasks_core/runner/action/parent.rb
```diff
@@ -0,0 +1,68 @@
...
+                :input => input,
+                :run_step_id => 2,
+                :execution_plan_id => targets[key][:task_id]
+            })
+          end
+          { :targets => result }
+        end
+
+        def use_group_runner?
+          !group_runner_action_class.nil?
```
iNecas: I think we talked about possibility to still allow the non-group-runner version for the next few releases, mainly to provide the backward compatibility and ability to compare the behavior of the implementations

lib/foreman_tasks_core/task_launcher/abstract.rb
```diff
@@ -0,0 +1,42 @@
...
+        if result.triggered?
+          { :result => 'success', :task_id => result.execution_plan_id }
+        else
+          plan = world.persistence.load_execution_plan(result.id)
+          { :result => 'error', :errors => plan.errors }
+        end
+      end
+
+      def action_class(input)
+        require 'pry'; binding.pry
```
iNecas: Hello pry, nice to meet you :)

lib/foreman_tasks_core/runner/parent.rb
```diff
@@ -0,0 +1,52 @@
...
+        super()
+      end
+
+      def generate_updates
+        @outputs.reduce({}) do |acc, (key, value)|
+          if value.empty? && @exit_status.nil?
+            acc
+          else
+            @outputs[key] = ForemanTasksCore::ContinuousOutput.new
+            key = host_action(key) unless key == :control
```
iNecas: This reads kind of hard, as there is no explanation what the `:control` key means, and it's kind of strange that we mix the `:control` key, with some `Dynflow::Action::Suspended` objects. It would be nice if you could think of possible refactoring to help with readability.

iNecas:
I went though the code again and left some comments. In general, it would be good to leave some explanatory docs at the key places to help understanding the flow of the code.

app/lib/actions/middleware/remote_task_triggering.rb
```diff
@@ -1,6 +1,8 @@
 module Actions
   module Middleware
     class RemoteTaskTriggering < ::Dynflow::Middleware
+      BATCH_SIZE = 1000
```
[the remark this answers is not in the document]
adamruzicka: This is just a "how many records should we load from the db at once", I don't think users need fine grained control over this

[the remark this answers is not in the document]
iNecas: Ok, as long as it doesn't influnce the batches we sent to the proxy, I'm good with this

adamruzicka:
now requires also https://github.com/Dynflow/dynflow/pull/313

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -149,23 +149,39 @@ def index
...
+      param :callbacks, Array, :of => param_group(:callback)
       def callback
-        task = ForemanTasks::Task::DynflowTask.find(params[:callback][:task_id])
-        ForemanTasks.dynflow.world.event(task.external_id,
-                                         params[:callback][:step_id].to_i,
-                                         # We need to call .to_unsafe_h to unwrap the hash from ActionController::Parameters
-                                         ::Actions::ProxyAction::CallbackData.new(params[:data].to_unsafe_h))
+        callbacks = params.key?(:callback) ? Array(params) : params[:callbacks]
+        callbacks.each do |payload|
+          process_callback(payload[:callback][:task_id], payload[:callback][:step_id].to_i, payload[:data].to_unsafe_h)
```
iNecas: Perhaps worth optimizing the `N` database task queries, and since we're only interested into `external_id` maybe even to fetch just those data, to minimiaze the amount of data we load here, since we're working in batches here

iNecas:
Wasn't this change supposed to still be compatible with older proxies? I see this, when trying against the proxy without the support for batch processing:

```
Runner error: NoMethodError - undefined method `<<' for #<Dynflow::Utils::IndifferentHash:0x0055e9ba64d580>
```

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -149,23 +149,39 @@ def index
...
         render :json => { :message => 'processing' }.to_json
       end
 
       private
 
+      def process_callback(task_id, step_id, data)
+        task = ForemanTasks::Task::DynflowTask.find(task_id)
+        ForemanTasks.dynflow.world.event(task.external_id,
+                                         step_id,
+                                         # We need to call .to_unsafe_h to unwrap the hash from ActionController::Parameters
```
iNecas: Is this comment still valid? I don't see the usage of `to_unsafe_h` here

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -7,13 +7,12 @@ class CheckOnProxyActions; end
...
       def run(event = nil)
-        if event.nil?
-          set_clock
-        elsif event == CheckOnProxyActions
+        if event == CheckOnProxyActions
           check_triggered
           set_clock
           action.send(:suspend)
         end
+        set_clock
```
iNecas: Shouldn't we still set the clock just at the first run (with event.nil?) and then only on `CheckOnProxyActions`. Why is this change needed?

app/controllers/foreman_tasks/api/tasks_controller.rb
adamruzicka: It is being called [here](https://github.com/theforeman/foreman-tasks/pull/369/files#diff-76e220fa2f3fb075d1aab086606ce317R170), I should move it there

iNecas:
Btw. some basic measuring (with 2600 hosts) shown about 15% speed improvement: I guess with remote proxies it would be even more significant.

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -68,6 +64,13 @@ def poll_proxy_tasks(url, tasks)
         action.action_logger.warn(_('Failed to check on tasks on proxy at %{url}: %{exception}') % { :url => url, :exception => e.message })
         []
       end
+
+      def in_remote_task_batches(scope)
+        # Sort by proxy_url so there are less requests per proxy
+        scope.order(:proxy_url, :id).find_in_batches(:batch_size => BATCH_SIZE) do |batch|
+          yield batch
```
iNecas: This fails with:

```

LocalJumpError: no block given (yield)
/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb:71:in `block in in_remote_task_batches'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/batches.rb:136:in `block in find_in_batches'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/batches.rb:238:in `block in in_batches'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/batches.rb:222:in `loop'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/batches.rb:222:in `in_batches'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/activerecord-5.2.1/lib/active_record/relation/batches.rb:135:in `find_in_batches'
/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb:70:in `in_remote_task_batches'
/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb:29:in `check_triggered'
/home/inecas/Projects/ws/foreman-rex/foreman-tasks/app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb:13:in `run'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware/stack.rb:22:in `call'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware/stack.rb:26:in `pass'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware.rb:18:in `pass'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware.rb:31:in `run'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware/stack.rb:22:in `call'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware/stack.rb:26:in `pass'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware.rb:18:in `pass'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware.rb:31:in `run'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware/stack.rb:22:in `call'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware/stack.rb:26:in `pass'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/dynflow-1.2.0/lib/dynflow/middleware.rb:18:in `pass'
```

iNecas:
Rebased version available in https://github.com/theforeman/foreman-tasks/pull/385

PR #370, opened 2018-10-22
Fixes #25268 - do not timeout when task is paused

jlsherrill:


jlsherrill:
This is making the assumption that if a task is paused, its due to an error and a timeout shouldn't be raised.

iNecas:
[test foreman-tasks]

iNecas:
Thanks @jlsherrill, test failures unrelated

PR #371, opened 2018-10-23
Fixes #25282 - Create a task for each target even if missing 

adamruzicka:
We always assumed we will get all targets from the DB. If some of
the targets were missing in the DB, we didn't create the task for
it and it lead to task counters going out of sync.

With this patch we create tasks for missing targets, but pass nil
instead of the target to it.

### DIFF
--- a/test/unit/actions/bulk_action_test.rb
+++ b/test/unit/actions/bulk_action_test.rb
@@ -1,0 +1,44 @@
+require 'foreman_tasks_test_helper'
+
+module ForemanTasks
+  class BulkActionTest < ActiveSupport::TestCase
+    before do
+      User.current = User.where(:login => 'apiadmin').first
+    end
+
+    Target = Struct.new(:id)
+
+    class ParentAction < Actions::BulkAction; end
+
+    class ChildAction < Actions::EntryAction
+      def plan(target)
+        target.id
+      end
+    end
+
+    describe Actions::BulkAction do
+      let(:targets) { (1..5).map { |i| Target.new i } }
+      let(:task) do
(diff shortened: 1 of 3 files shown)


adamruzicka:
requires https://github.com/Dynflow/dynflow/pull/299 to fix the issue fully

adamruzicka:
[test foreman-tasks]

test/unit/actions/bulk_action_test.rb
```diff
@@ -0,0 +1,46 @@
...
+      def plan(target)
+        target.id
+      end
+    end
+
+    describe Actions::BulkAction do
+      let(:targets) { (1..5).map { |i| Target.new i } }
+      let(:task) do
+        triggered = ForemanTasks.trigger(ParentAction, ChildAction, targets)
+        triggered.finished.wait(2)
```
iNecas: I think we should raise an error here when exceeding the timeout (the wait returns false in that case): When I tried running the test without dynflow changes, the test was still passing, but the task actually never finished, which I noticed by longer time until the tests actually finished (as there is the termination timeout waiting for events to finish).

iNecas:
Dynflow 1.1.5, including the dependent change, got released. After bumping the version requirement here, plus fixing the one comment I had, this would be good to go.

test/unit/actions/bulk_action_test.rb
adamruzicka: Good point, fixed

adamruzicka:
This PR is :green_apple:

iNecas:
Thanks @adamruzicka

PR #372, opened 2018-11-02
Fix rubocop issues

adamruzicka:
I'm not very fond of enforcing `Rails/RefuteMethods` cop, but let's see how the vote in https://github.com/theforeman/foreman/pull/6194 turns out and we can either keep or drop the last commit accordingly.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -75,6 +75,9 @@ Layout/IndentHeredoc:
     - '*.gemspec'
     - bin/*
 
+Layout/EmptyLineAfterGuardClause:
+  Enabled: false
+
 Style/RescueStandardError:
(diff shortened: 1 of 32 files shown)


adamruzicka:
[test foreman-tasks]

iNecas:
Can we rebase and revert https://github.com/theforeman/foreman-tasks/pull/372/commits/70991e95d044687db17716c574a056b695b508ee similarly as we did in rex?

adamruzicka:
Rebased, reverted and disabled the Layout/EmptyLineAfterGuardClause cop

iNecas:
Thanks @adamruzicka I was able to reproduce the test failure issues, so it's not related to this PR

PR #375, opened 2018-11-09
Fixes #25427 - Make task autocomplete hit the right endpoint

adamruzicka:


iNecas:
Tested and works well. Not sure what Jenkins doesn't like, but I'm pretty confident it's nothing related to this PR. Thanks @adamruzicka

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -96,6 +96,10 @@ def resource_class
       ForemanTasks::Task
     end
 
+    def auto_complete_controller_name
```
iNecas: This code has actually been added in https://github.com/theforeman/foreman-tasks/pull/375 to workaround regression introduced in https://github.com/theforeman/foreman/pull/5771. It seems that it has not fixed the issue properly.

In order for this fix to work, we need also https://github.com/theforeman/foreman/pull/6460.

After that is merged, we can just remove the `auto_complete_controller_name` method complete from here, so that it would work as before https://github.com/theforeman/foreman-tasks/pull/375.

PR #376, opened 2018-11-12
Fixes #25434 - default debug export time is a month

lzap:
It's creating 30+ GiB CSV files on one of customers production installation, this renders foremand-debug unusable "by default".


iNecas:
I'm not sure about the limit, as there are `n` steps per task, so for 1000 tasks, we would get just portion of steps. Wherefore, if we wanted to some meaningful limiting, it should be by the started_at time. So we could search for 1000 (or higher number) of steps/actions/tasks, and take the highest timestamp of those to limit the amout of tasks.
For start, I think it could be useful just limit `DYNFLOW_EXPORT_MONTHS` to 1

adamruzicka:
Sounds good to me, also could you please rebase?

lzap:
Done.

extra/dynflow-debug.sh
```diff
@@ -31,10 +31,10 @@ add_files /var/log/foreman/dynflow_executor*.output*
...
+    export_csv "select dynflow_execution_plans.* from foreman_tasks_tasks join dynflow_execution_plans on (foreman_tasks_tasks.external_id = dynflow_execution_plans.uuid) where foreman_tasks_tasks.started_at > 'now'::timestamp - '${DYNFLOW_EXPORT_MONTHS:-1} months'::interval limit ${DYNFLOW_EXPORT_LIMIT:-100000}" "$DIR/dynflow_execution_plans.csv"
+    export_csv "select dynflow_actions.* from foreman_tasks_tasks join dynflow_actions on (foreman_tasks_tasks.external_id = dynflow_actions.execution_plan_uuid) where foreman_tasks_tasks.started_at > 'now'::timestamp - '${DYNFLOW_EXPORT_MONTHS:-1} months'::interval limit ${DYNFLOW_EXPORT_LIMIT:-100000}" "$DIR/dynflow_actions.csv"
+    export_csv "select dynflow_steps.* from foreman_tasks_tasks join dynflow_steps on (foreman_tasks_tasks.external_id = dynflow_steps.execution_plan_uuid) where foreman_tasks_tasks.started_at > 'now'::timestamp - '${DYNFLOW_EXPORT_MONTHS:-1} months'::interval limit ${DYNFLOW_EXPORT_LIMIT:-100000}" "$DIR/dynflow_steps.csv"
+    export_csv "select * from dynflow_schema_info limit ${DYNFLOW_EXPORT_LIMIT:-100000}" "$DIR/dynflow_schema_info.csv"
```
adamruzicka: The `dynflow_schema_info` table should always be just one row, I don't think we need the limit here.

PR #378, opened 2018-12-04
Fix intermittent task controller failures

iNecas:
By waiting for specific event to happen

### DIFF
--- a/test/controllers/api/tasks_controller_test.rb
+++ b/test/controllers/api/tasks_controller_test.rb
@@ -30,20 +30,35 @@ module ForemanTasks
       describe 'GET /api/tasks/summary' do
         class DummyTestSummaryAction < Support::DummyDynflowAction
           # a dummy test action that do nothing
+          def plan(suspend = false)
+            plan_self :suspend => suspend
+          end
+
           def run
             # a dummy run method
+            if input[:suspend] && output[:suspended].nil?
+              output[:suspended] = true
+              suspend
+            end
           end
         end
 
         test_attributes :pid => 'bdcab413-a25d-4fe1-9db4-b50b5c31ebce'
         it 'get tasks summary' do
-          ForemanTasks.trigger(DummyTestSummaryAction)
+          triggered = ForemanTasks.trigger(DummyTestSummaryAction, true)
+          wait_for { ForemanTasks::Task.find_by(external_id: triggered.id).state == 'running' }
+          wait_for do
+            w = ForemanTasks.dynflow.world
+            w.persistence.load_step(triggered.id, 2, w).state == :suspended
+          end
           get :summary
           assert_response :success
           response = JSON.parse(@response.body)
           assert_kind_of Array, response
           assert_not response.empty?
           assert_kind_of Hash, response[0]
+          ForemanTasks.dynflow.world.event(triggered.id, 2, nil)
+          triggered.finished.wait
         end
       end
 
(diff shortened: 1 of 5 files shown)


test/controllers/api/tasks_controller_test.rb
```diff
@@ -37,7 +37,8 @@ def run
 
         test_attributes :pid => 'bdcab413-a25d-4fe1-9db4-b50b5c31ebce'
         it 'get tasks summary' do
-          ForemanTasks.trigger(DummyTestSummaryAction)
+          triggered = ForemanTasks.trigger(DummyTestSummaryAction)
+          wait_for { ForemanTasks::Task.find_by(external_id: triggered.id).state = 'stopped' }
```
adamruzicka: Shouldn't we compare for equality instead of assigning?

iNecas: Thank you: you've just saved several hours of my life hunting for the problem :)

test/controllers/api/tasks_controller_test.rb
```diff
@@ -30,20 +30,35 @@ class Api::TasksControllerTest < ActionController::TestCase
...
         end
 
         test_attributes :pid => 'bdcab413-a25d-4fe1-9db4-b50b5c31ebce'
         it 'get tasks summary' do
-          ForemanTasks.trigger(DummyTestSummaryAction)
+          triggered = ForemanTasks.trigger(DummyTestSummaryAction, true)
+          wait_for { ForemanTasks::Task.find_by(external_id: triggered.id).state == 'running' }
+          wait_for do
+            w = ForemanTasks.dynflow.world
+            w.persistence.load_step(triggered.id, 2, w).state == :suspended
```
iNecas: @adamruzicka I've added additional waiting just to ensure we're in suspended state, so that we make sure that the event gets to the right recipient later.

adamruzicka:
Tests on mysql seem to be failing with which is mostly 

```
2018-12-04T17:15:09 [W|dyn|] Error on on_execution_plan_save event
 | ActiveRecord::ValueTooLong: Mysql2::Error: Data too long for column 'action' at row 1: UPDATE `foreman_tasks_tasks` SET `state` = 'running', `action` = 'Dummy proxy action {\"foo\"=>\"bar\",\n \"connection_options\"=>{\"retry_interval\"=>15, \"retry_count\"=>4},\n \"proxy_url\"=>\"proxy.example.com\",\n \"proxy_action_name\"=>\"Proxy::DummyAction\",\n \"current_user_id\"=>886836129,\n \"current_organization_id\"=>nil,\n \"current_location_id\"=>nil}' WHERE `foreman_tasks_tasks`.`id` = 'd9976b7f-3909-4555-abf2-b0567a61784b'

```

redefining `#task_input` to return just `''` in `::Actions::ProxyAction` seems to fix the issue.

iNecas:
I've fixed the issue by using `''` as implicit `humanized_{input,output}` for `Action::Base`. Most of root task actions use `Actions::EntryAction` that uses own implementation of the methods, so I think it's better to not try to humanize something we don't know how to do.

I've also fixed the issue causing the tests to fail in November.

iNecas:
I will link issues to commits that deserve link (mainly the one updating the humanized_input) tomorrow: I'm curious to see how the tests are doing now

adamruzicka:
Last remaining failure. It looks like we may try to create a `Time` object where minutes equal 60, if the test is run at `*:59`

https://github.com/theforeman/foreman-tasks/blob/master/test/unit/recurring_logic_test.rb#L97

adamruzicka:
> AuditMailerTest.test_0006_Audit template change should not crash

Guess we can say we're finally green

iNecas:
Ok, I've rebased, squashed things that could be squashed and linked commits to issues that have actually some user impact. Should be ready to go after the tests hopefully finally pass.

adamruzicka:
Thanks @iNecas !

PR #379, opened 2018-12-07
Fixes #25608 - Rework locking

adamruzicka:


adamruzicka:
Rebased

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
```diff
@@ -0,0 +1,28 @@
...
+    # rubocop:disable Rails/CreateTableWithTimestamps
+    create_table :foreman_tasks_links do |t|
+      task_id_options = { :index => true, :null => true }
+      if on_postgresql?
+        t.uuid :task_id, task_id_options
+      else
+        t.string :task_id, task_id_options
+      end
+      # TODO: Figure out if we need this
+      # t.string :name, index: true, null: false
```
iNecas: Seems like something to decide

db/migrate/20181206131436_drop_all_locks.foreman_tasks.rb
```diff
@@ -0,0 +1,11 @@
+class DropAllLocks < ActiveRecord::Migration[5.0]
+  BATCH_SIZE = 10_000
+
+  def up
+    ForemanTasks::Lock.limit(BATCH_SIZE).delete_all while ForemanTasks::Lock.any?
```
iNecas: Shouldn't this be limited to stopped tasks only?

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.string :resource_type
+      t.integer :resource_id
+    end
+    # rubocop:enable Rails/CreateTableWithTimestamps
+
+    add_index :foreman_tasks_links, [:resource_type, :resource_id]
+    add_index :foreman_tasks_links, [:task_id, :resource_type, :resource_id],
```
iNecas: I wonder if we need the unique constraint here - just thinking if it doesn't influence negatively the link creation process once there are many of those?

adamruzicka: My expectation was a task can be linked to a resource only once and having it enforced by the database seemed like the right thing to do. If we decided to go the other way and check if a link exists in ruby, then we'd still need this index to be able to find the record quickly. At that point, there is probably no difference between an index and an index with unique constraint. Or we could just not check it at all and live with occasional duplicates

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
adamruzicka: I can't think of a reason why we should keep this around, I'd drop it.

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
iNecas: That's true. One options might be to cache the task object on `Thread.current` in https://github.com/theforeman/foreman-tasks/blob/master/app/lib/actions/base.rb#L9, in which case all the plannings would know which resources are already linked, and the deduplication could be achieved in memory. I'm ok though to have that handled as separate issue, while thinking more about it first.

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
iNecas: :+1:

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
adamruzicka: Ok, I'll leave it as it is for now.

db/migrate/20181206131627_make_locks_exclusive.foreman_tasks.rb
```diff
@@ -0,0 +1,14 @@
+class MakeLocksExclusive < ActiveRecord::Migration[5.0]
+  def up
+    change_table :foreman_tasks_locks do |t|
+      t.remove :exclusive
+      t.remove :name
+      t.remove_index :name => 'index_foreman_tasks_locks_on_resource_type_and_resource_id'
+      t.index [:resource_type, :resource_id], :unique => true
```
iNecas: Dunno why it's not happening in jenkins, but with postgres on my setup, I'm getting `PG::UniqueViolation: ERROR:  could not create unique index "index_foreman_tasks_locks_on_resource_type_and_resource_id"`. When looking at some production setup, I'm seeing:

```
# in psql
\d foreman_tasks_locks
Indexes:
    "foreman_tasks_locks_pkey" PRIMARY KEY, btree (id)
    "index_foreman_tasks_locks_name_resource_type_resource_id" btree (name, resource_type, resource_id)
    "index_foreman_tasks_locks_on_exclusive" btree (exclusive)
    "index_foreman_tasks_locks_on_name" btree (name)
    "index_foreman_tasks_locks_on_resource_type" btree (resource_type)
    "index_foreman_tasks_locks_on_resource_type_and_resource_id" btree (resource_type, resource_id)
    "index_foreman_tasks_locks_on_task_id" btree (task_id)
```

adamruzicka: Any `********** NO FILE **********` migrations in the output of `bundle exec rake db:migrate:status`?

iNecas: Oh, I see, it was not due to the conflicting index (I've missed you've deleted the previous on just one line before), but there have still been some duplicates: I need to check whether I haven't run some of the migrations before during testing, that would be skipped now.

adamruzicka: I'll need to work on the `drop_old_locks` migration a bit more.

db/migrate/20181206131436_drop_old_locks.foreman_tasks.rb
```diff
@@ -0,0 +1,24 @@
...
+    scope.limit(BATCH_SIZE).delete_all while scope.any?
+
+    # For each group of locks, where each lock has the same task_id, resource_type and resource_id
+    # return the highest id within the group, if there is more than 1 lock in the group
+    scope = ForemanTasks::Lock.select("MAX(id) as id")
+                              .group(:task_id, :resource_type, :resource_id)
+                              .having("count(*) > 1")
+
+    # Make sure there is at most one lock per task and resource
+    ForemanTasks::Lock.where(:id => scope.limit(BATCH_SIZE)).delete_all while scope.any?
```
adamruzicka: A single iteration removing `10000` locks takes around `0.9` second on my machine. Given most locks should be already pruned by [L9](https://github.com/theforeman/foreman-tasks/pull/379/files#diff-3223d086987c91ff687cb8b245e4c45aR9), I'd say this is acceptable.

iNecas: :+1:

adamruzicka:
[test foreman-tasks]

adamruzicka:
Should be ready for review

test/controllers/tasks_controller_test.rb
```diff
@@ -48,8 +48,8 @@ def in_taxonomy_scope(organization, location = nil)
             @locations = [0, 0].map { FactoryBot.create(:location) }
             @tasks = [0, 0].map { FactoryBot.create(:some_task) }
             @tasks.zip(@organizations, @locations).each do |task, org, loc|
-              Lock.link!(org, task.id)
-              Lock.link!(loc, task.id)
+              Link.link!(org, task)
```
adamruzicka: Should we maybe provide a compatibility layer which would contain `Lock.link!` and would print a deprecation warning?

ares: sounds reasonable to me

ares:
@ezr-ondrej could you please take a look at this one once you'll have a bit of time? Perhaps something to discuss when you meet @adamruzicka face to face again

app/views/foreman_tasks/api/tasks/details.json.rabl
```diff
@@ -9,10 +9,13 @@ node(:failed_steps) { @task.input_output_failed_steps }
 node(:running_steps) { @task.input_output_running_steps }
 node(:help) { troubleshooting_info_text }
 node(:has_sub_tasks) { @task.sub_tasks.any? }
+node(:allowDangerousActions) { Setting['dynflow_allow_dangerous_actions'] }
```
ezr-ondrej, line 12: Why do we use `camelCase` here?

ezr-ondrej (changes requested):
Looks quite nice overall :+1:
As discussed offline, I'd like the migrations to pass if migrating down, for the plugin to be uninstallable.
To the unique locking, I've got bad news it was this one: https://github.com/theforeman/foreman/pull/7772#discussion_r450059828 and that covered my use-case, but it won't cover the tasks interlocking. We should still wrap the exception class for the user to know what have happend, right?

app/views/foreman_tasks/api/tasks/details.json.rabl
adamruzicka: This is a bad rebase, it doesn't seem to be used anywhere.

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
```diff
@@ -0,0 +1,26 @@
...
+              :unique => true, :name => 'foreman_tasks_links_unique_index'
+  end
+
+  private
+
+  def on_postgresql?
+    ActiveRecord::Base.connection.adapter_name == 'PostgreSQL'
```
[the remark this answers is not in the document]
adamruzicka: Rubocop has another idea
```
db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb:24:5: C: Performance/Casecmp: Use ActiveRecord::Base.connection.adapter_name.casecmp('postgresql').zero? instead of ActiveRecord::Base.connection.adapter_name.downcase == 'postgresql'.
    ActiveRecord::Base.connection.adapter_name.downcase == 'postgresql'
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

app/lib/actions/helpers/lock.rb
```diff
@@ -4,19 +4,24 @@ module Lock
...
+        else
+          ::ForemanTasks::Lock.exclusive!(resource, task)
+        end
       end
 
       # @see Lock.lock!
-      def lock!(resource, *lock_names)
-        phase! Dynflow::Action::Plan
-        ::ForemanTasks::Lock.lock!(resource, task.id, *lock_names.flatten)
+      def lock!(resource, *_lock_names)
```
ezr-ondrej, line 17: Do I read it correctly we are renaming this to `exclusive_lock!` to be more verbose about the behaviour?
Should we deprecate this method?

ezr-ondrej (commented):
I've found few more nits going through the code, but apart of that, LGTM.
Let me know about these, I'll find a way to test it in the mean time :+1:

db/migrate/20181206123910_create_foreman_tasks_links.foreman_tasks.rb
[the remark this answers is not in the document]
ezr-ondrej: I'm good with that :shrug:

db/migrate/20181206124952_migrate_non_exclusive_locks_to_links.foreman_tasks.rb
```diff
@@ -0,0 +1,14 @@
+class MigrateNonExclusiveLocksToLinks < ActiveRecord::Migration[5.0]
+  def up
+    execute <<-SQL
+      INSERT INTO foreman_tasks_links(task_id, resource_type, resource_id)
+      SELECT DISTINCT locks.task_id, locks.resource_type, locks.resource_id
+      FROM foreman_tasks_locks AS locks
+      LEFT JOIN foreman_tasks_links AS links
+        ON links.task_id = locks.task_id
+          AND links.resource_type = locks.resource_type
+          AND links.resource_id = locks.resource_id
```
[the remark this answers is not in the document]
adamruzicka: I think in the old model it was possible to have a task with multiple non-exclusive locks on a single resource, but the links table has an unique index on the `[:task_id, :resource_type, :resource_id]` triplet. It shouldn't really happen but I'd rather stay on the safe side than have a migration which requires the users to do a manual cleanup

app/lib/actions/helpers/lock.rb
adamruzicka: The rename was done because of the change in arity. Guess deprecation warning would be a good idea

test/unit/lock_test.rb
```diff
@@ -2,7 +2,7 @@
 require 'foreman_tasks_test_helper'
 
 module ForemanTasks
-  class LockTest < ::ActiveSupport::TestCase
+  class LockingTest < ::ActiveSupport::TestCase
```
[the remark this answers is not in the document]
adamruzicka: Wonder what I did here, both files are identical apart from changes I did this week to one of them :confused:

app/lib/actions/helpers/lock.rb
adamruzicka: Marked as deprecated and to be removed in foreman 2.3, hope this gives enough time for people to adjust to this change.

ezr-ondrej: 2.3 is around a corner :thinking: shouldn't we enable one version in production?

adamruzicka: Well, my thought was that 2.2 is not even out yet, but I wouldn't mind bumping it to 2.4

ezr-ondrej: I'd be for bumping to 2.4, because this goes to production with foreman 2.3, right? we won't be backporting it to 2.2, even though it still in RC, right? :shrug:

adamruzicka: Right, 2.4 it is

ezr-ondrej:
Code make sense, but I've only find Links in Katello, would you know about any usage of exclusive locking we have, I can use for testing, or do I need to DIY? :thinking:

adamruzicka:
I think a lot of Katello stuff uses exclusive locking, basically anything related to content syncing is using it iirc

adamruzicka:
@ezr-ondrej did you have time to take this for a spin?

ezr-ondrej (approved):
I've been running it a while now on my luna, I went again over the code and I found no issues :+1: 
Let's get this in :+1:

ezr-ondrej:
I can't merge though

ares:
Thanks @aruzicka and @ezr-ondrej, merging!

PR #380, opened 2018-12-10
Fixes #25655 - Recurring Logic skips the first schedule before starting

sjha4:
In the method start_after we call next_occurence_time which calls @parser.next(start_time) when starting even if the start time is valid for the cron and we could very well start at the start-time itself. 
For example
cron = "* * * * *"
start_time = 12:35:00
We can start at 12:35 instead of starting at 12:36.

### DIFF
--- a/app/models/foreman_tasks/recurring_logic.rb
+++ b/app/models/foreman_tasks/recurring_logic.rb
@@ -83,6 +83,9 @@ module ForemanTasks
 
     def next_occurrence_time(time = Time.zone.now)
       @parser ||= CronParser.new(cron_line, Time.zone)
+      # @parser.next(start_time) is not inclusive of the start_time hence stepping back one run to include checking start_time for the first run.
+      before_next = @parser.next(@parser.last(time))
+      return before_next if before_next >= time && tasks.count == 0
       @parser.next(time)
     end
 
--- a/test/unit/recurring_logic_test.rb
+++ b/test/unit/recurring_logic_test.rb
@@ -25,9 +25,9 @@ class RecurringLogicsTest < ActiveSupport::TestCase
       minute = 0
       reference_time = Time.utc(year, month, day, hour, minute)
       parser = ForemanTasks::RecurringLogic.new_from_cronline('* * * * *')
-      parser.next_occurrence_time(reference_time).must_equal Time.utc(year, month, day, hour, minute + 1)
+      parser.next_occurrence_time(reference_time).must_equal Time.utc(year, month, day, hour, minute)
       parser = ForemanTasks::RecurringLogic.new_from_cronline('*/2 * * * *')
-      parser.next_occurrence_time(reference_time).must_equal Time.utc(year, month, day, hour, minute + 2)
+      parser.next_occurrence_time(reference_time).must_equal Time.utc(year, month, day, hour, minute)
       parser = ForemanTasks::RecurringLogic.new_from_cronline('*/2 18,19 * * *')
       parser.next_occurrence_time(reference_time).must_equal Time.utc(year, month, day, 18)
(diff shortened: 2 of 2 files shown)


beav:
@sjha4 without this PR, if I had a cron for `0 * * * *` and I entered it at 11:45, would the cron not fire at 12:00?

adamruzicka:
@beav If I got that right if you entered a cron for `0 * * * *` and configured it to start at 12:00, it wouldn't fire until 13:00

adamruzicka:
Depends whether you set the start at property, without it it should work fine

beav:
test was successful. repro steps:

* create an hourly sync plan, set "start time" to something (I usually used ~5 min in the future)
* note that "next sync" one hour *past* the start time
* wait 6 minutes, note that sync plan didn't execute

with patch:

* create an hourly sync plan, set "start time" to something (again, ~5 min for my test)
* note that "next sync" is now correct
* wait 6 minutes, sync plan executes

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -83,6 +83,7 @@ def cancel
 
     def next_occurrence_time(time = Time.zone.now)
       @parser ||= CronParser.new(cron_line, Time.zone)
+      return @parser.next(@parser.last(time)) if @parser.next(@parser.last(time)) >= time && tasks.count == 0
```
beav: this logic works and is correct, but is a little hard to follow. What data is `@parser.next(@parser.last(time))` capturing? is it the next run, but backdated one run?

sjha4: So with a start_at 12:46 if we have a cron "46 * * * *" , @parser.next(12:46) returns 13:46 while parser.last returns 11:46 and parser.next(11:46) returns 12:46 which is when we want the task to run..

beav: Can you add a comment saying something like "CronParser's `next` is not inclusive of the current time, so we step backwards one run to calculate the next run"? My concern is that if I came across this line in the future, I would not know what it was doing at all.

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -83,6 +83,8 @@ def cancel
 
     def next_occurrence_time(time = Time.zone.now)
       @parser ||= CronParser.new(cron_line, Time.zone)
+      # @parser.next(start_time) is not inclusive of the start_time hence stepping back one run to include checking start_time for the first run.
```
beav: :+1:

sjha4:
@adamruzicka : I am not sure if the tests ran ok..

sjha4:
@adamruzicka : Does this look good to go to you?

app/models/foreman_tasks/recurring_logic.rb
adamruzicka: One nitpick, it would be nice if you could avoid generating the time twice. Something like this might work
```ruby
before_next = @parser.next(@parser.last(time))
return before_next if before_next >= time && task.count.zero?
@parser.next(time)
```

sjha4:
@adamruzicka : I updated the code as per your reviews..I am not sure if the test suite failure is related.

sjha4:
Thanks @adamruzicka ! This needs to land with some version of katello 3.10..I am not sure what the dates would be but probably will need a foreman-tasks release when merged?

beav:
@sjha4 do you mean like a 3.10.1?

sjha4:
@beav: Yes..I thought 3.10.1 was out but turns out it isn't :smile: ..so that is the target release.

adamruzicka:
Thanks @sjha4 !

PR #383, opened 2018-12-19
 Fixes #25736 - Pass id of request which triggered a task into the task

adamruzicka:


iNecas:
Tested and works well, except covering exceptions logging, which could be solved by this https://github.com/adamruzicka/foreman-tasks/pull/2

iNecas:
Dynflow 1.2.1 has been released, let's bump the version dependency

app/lib/actions/recurring_action.rb
```diff
@@ -10,11 +10,15 @@ def self.included(base)
     # Hook to be called when a repetition needs to be triggered. This either happens when the plan goes into planned state
     #   or when it fails.
     def trigger_repeat(execution_plan)
+      request_id = ::Logging.mdc['request']
+      ::Logging.mdc['request'] = nil
```
iNecas: After discussion with @ponrjek, we think the best way to support the recurring jobs would be to assign it a new (but non-empty) request id, so that we can group the related tasks from every iteration together (otherwise, we can do it just with the first iteration)

iNecas:
Two last comments before merging this.

app/lib/actions/recurring_action.rb
adamruzicka: Should I also log that?

iNecas: I canl handle the logging part in separate PR

test/unit/actions/recurring_action_test.rb
```diff
@@ -82,6 +82,25 @@ def run
...
+          ::Logging.mdc['request'] = expected_id
+
+          delay_options = recurring_logic.generate_delay_options
+          task = ForemanTasks.delay HookedAction, delay_options, true, args.last
+          task.input[:current_request_id].must_equal expected_id
+
+          # Perform the planning (trigger repeat)
+          task.execution_plan.delay_record.plan
+          repetition = recurring_logic.tasks.find { |t| t.id != task.id }
+          repetition.input[:current_request_id].must_be :nil?
```
iNecas: This needs to be updated as well

iNecas:
Merged as 12b53f2, thanks @adamruzicka

PR #385, opened 2019-01-31
Fixes #25160 - Trigger tasks on the smart proxy in batches

iNecas:
Continues from https://github.com/theforeman/foreman-tasks/pull/369


iNecas:
Rebased in https://github.com/theforeman/foreman-tasks/pull/385, let's continue there

lib/foreman_tasks_core/task_launcher/batch.rb
```diff
@@ -7,53 +7,14 @@ class ParentAction < ::Dynflow::Action
       # { task_id => { :action_class => Klass, :input => input } }
       def plan(launcher, input_hash)
         launcher.launch_children(self, input_hash)
-        sequence do
```
iNecas: @adamruzicka I've removed this part of https://github.com/theforeman/foreman-tasks/pull/385/commits/d2af822a59439f848748e72aebc6c97224040870#diff-854a78e251be71cf44b31b98e342ee22R10, as it was changing the end2end behaviour of the ssh provider (so no hosts would be marked as finished until the last host from the batch finishes) and could look like a regression.

iNecas: Also, it was not working when there were some failed hosts, as the task failed with `Stop` rescue_scenario

test/core/unit/task_launcher_test.rb
```diff
@@ -11,7 +11,6 @@ class TaskLauncherTest < ActiveSupport::TestCase
         let(:launcher_input) { { 'action_class' => Support::DummyDynflowAction.to_s, 'action_input' => input } }
         let(:input) { { :do => :something } }
         let(:expected_result) { input.merge(:callback_host => {}) }
-        let(:expected_batch_result) { expected_result.merge('callback' => nil) }
```
iNecas: @adamruzicka I guess since we're sending the results from each action vs. the parent one, we don't expect the callback to be nil, right?

iNecas:
Test failures unrelated. @adamruzicka anything to add?

adamruzicka:
Nothing comes to mind, let's get this in.

iNecas:
@adamruzicka squashed - ready to the compatiblity changes so that we can update rex without breaking ansible until ready

adamruzicka:
The tests are greener than they were, looks like the current failures are transient

iNecas:
Thanks @adamruzicka, squashed the last two commits and merged. Ready to rebase the other PR against this

PR #386, opened 2019-01-31
Add features needed for ansible-runner

adamruzicka:
Based on:
- [x] ~https://github.com/theforeman/foreman-tasks/pull/369~
- [x] https://github.com/theforeman/foreman-tasks/pull/385


adamruzicka:
Rebased on top of https://github.com/theforeman/foreman-tasks/pull/385

lib/foreman_tasks_core/task_launcher/abstract.rb
```diff
@@ -0,0 +1,42 @@
...
+      def action_class(input)
+        options[:action_class_override] || ::Dynflow::Utils.constantize(input['action_class'])
+      end
+
+      def with_callback(input)
+        input.merge(:callback_host => callback)
+      end
+
+      def trigger(parent, klass, *input)
+        world.trigger { world.plan_with_caller(parent, klass, *input) }
```
iNecas: this was renamded to `plan_with_options` in new dynflow, see https://github.com/Dynflow/dynflow/commit/e4c0f9cd16a2d3acd5ae8495cd76649a9ad1c06e

iNecas: Right format should be `world.trigger { world.plan_with_options(caller_action: parent, action_class: klass, args: input) }`

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -149,23 +149,41 @@ def index
...
-                                         params[:callback][:step_id].to_i,
-                                         # We need to call .to_unsafe_h to unwrap the hash from ActionController::Parameters
-                                         ::Actions::ProxyAction::CallbackData.new(params[:data].to_unsafe_h, :request_id => ::Logging.mdc['request']))
+        callbacks = params.key?(:callback) ? Array(params) : params[:callbacks]
+        ids = callbacks.map { |payload| payload[:callback][:task_id] }
+        external_map = Hash[*ForemanTasks::Task.where(:id => ids).pluck(:id, :external_id).flatten]
+        callbacks.each do |payload|
+          # We need to call .to_unsafe_h to unwrap the hash from ActionController::Parameters
+          callback = payload[:callback]
+          process_callback(external_map[callback[:task_id]], callback[:step_id].to_i, payload[:data].to_unsafe_h, :request_id => ::Logging.mdc['request'])
```
iNecas: There were new callback added that need to be updated here https://github.com/theforeman/foreman-tasks/commit/d70a2916057e348694062ef476149a71914fd52e#diff-76e220fa2f3fb075d1aab086606ce317R163

iNecas: It seems this PR has the right version https://github.com/theforeman/foreman-tasks/pull/385/files#diff-76e220fa2f3fb075d1aab086606ce317R181

adamruzicka:
Rebased, the changes here should now work with:
- [x] REX master + ansible master
- [x] REX launcher + ansible master
- [x] REX launcher + ansible ansible-runner

adamruzicka:
Live output does not work when using a group runner, that will need to be reworked.

iNecas:
rubocop is not happy

iNecas:
Rebased against latest master + added whitespace fix to make rubocop happy, let's see what else might break

lib/foreman_tasks_core/task_launcher/batch.rb
```diff
@@ -27,11 +27,31 @@ def launch!(input)
...
         input_hash.each do |task_id, input|
-          launcher = Single.new(world, callback, :parent => parent)
-          launcher.launch!(input)
+          launcher = child_launcher(parent)
+          launcher.launch!(transform_input(input))
           results[task_id] = launcher.results
         end
       end
+
+      def prepare_batch(input_hash)
```
iNecas: Is there some better name for this method: took me a while to figure out what it's supposed to do.

I think it would help the the launched not having to know about the callbacks, and just provide a method `planned_task_ids` that would return the task ids that were successfully planned, and use it in `GroupParentAction` to prepare the callback input

lib/foreman_tasks_core/task_launcher/group.rb
```diff
@@ -0,0 +1,122 @@
...
+
+module ForemanTasksCore
+  module TaskLauncher
+    class AbstractGroup < Batch
+      class GroupParentAction < ForemanTasksCore::TaskLauncher::ParentAction
+        def plan(launcher, input_hash)
+          launcher.launch_children(self, input_hash)
+          sequence do
+            results = plan_self
+            plan_action BatchCallback, launcher.prepare_batch(input_hash), results.output[:results]
```
iNecas: As suggested in https://github.com/theforeman/foreman-tasks/pull/386/files#r267295169, the responsiblity for formating the callback input might be better in this action. Something like:

```ruby
prepare_callback_input(input_hash, launcher.planned_task_ids)
# ...
def prepare_callback_input(input_hash, task_ids)
  task_ids.reduce({}) do |hash, task_id|
    hash.update(task_id => input_hash[task_id]['action_input']['callback'])
  end
end

```

lib/foreman_tasks_core/task_launcher/group.rb
```diff
@@ -0,0 +1,122 @@
...
+
+        def run(event = nil)
+          super unless event == Dynflow::Action::Skip
+        end
+
+        def check_for_errors!(optional = true)
+          super unless optional
+        end
+
+        def on_finish
```
iNecas: Where is this method called from?

lib/foreman_tasks_core/task_launcher/group.rb
```diff
@@ -0,0 +1,122 @@
...
+            acc.merge(cur.execution_plan_id => cur.output)
+          end
+        end
+
+        def finalize
+          output.delete(:results)
+          check_for_errors!
+        end
+
+        def initiate
```
iNecas: nitpick: I would expect `initate` method to in in code before `finalize`

lib/foreman_tasks_core/task_launcher/group.rb
```diff
@@ -0,0 +1,122 @@
...
+        private
+
+        def format_payload(input_hash, results)
+          input_hash.map do |task_id, callback|
+            { :callback => callback, :data => results[task_id] }
+          end
+        end
+      end
+
+      class Dummy < ::ForemanTasksCore::Runner::Action
```
iNecas: Naming is hard, but `Dummy` is quite confusing when looking at the dynflow console. What about calling it `ChildAction`, and `GroupParentAction` would be just `ParentAction`

lib/foreman_tasks_core/task_launcher/abstract.rb
```diff
@@ -13,6 +13,8 @@ def launch!(_input)
         raise NotImplementedError
       end
 
+      def self.input_format; end
```
iNecas: What's purpose of this method?

lib/foreman_tasks_core/task_launcher/group.rb
```diff
@@ -0,0 +1,122 @@
+require 'foreman_tasks_core/runner'
+
+module ForemanTasksCore
+  module TaskLauncher
+    class AbstractGroup < Batch
```
iNecas: This is pretty complex class that would benefit from some documentation comment, explaining:

1. what's the responsibility of each sub-class in here (`GroupParentAction`, `Dummy`. `GroupRunner`, `BatchCallback`)
2. what's the interaction between the classes
3. what methods are expected to be overridden

lib/foreman_tasks_core/task_launcher/group.rb
adamruzicka: https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/action/with_sub_plans.rb#L137

lib/foreman_tasks_core/task_launcher/abstract.rb
adamruzicka: I used it as a reminder of how the input for the runner will look like (see [here](https://github.com/theforeman/foreman-tasks/pull/386/files#diff-186c5dc3b3701e52c768b38bea0c1c2cR5)), guess all these could be removed

iNecas:
Rubocop has not been utterly happy: pushed a commit to fix it to see the other test results.

lib/foreman_tasks_core/task_launcher/batch.rb
```diff
@@ -1,37 +1,37 @@
...
-      end
-
-      def rescue_strategy
-        Dynflow::Action::Rescue::Fail
-      end
-    end
-
     class Batch < Abstract
       def launch!(input)
         trigger(nil, ParentAction, self, input)
```
iNecas: Seems like missed rename

iNecas: The tests are failing now for the very same reason.

iNecas:
Squashed and merged as 518aa5abc6ad2e6000bde5a34406b8f8c1c594e6, released in foreman-tasks-core 0.3.2

PR #387, opened 2019-02-01
fixes #25958 - incorrect value for auto_complete_controller_name

alanwevans:
- [x] [required change on Foreman side](https://github.com/theforeman/foreman/pull/6460)

/foreman_tasks/tasks should be foreman_tasks_tasks
Fixes https://projects.theforeman.org/issues/25958

### DIFF
--- a/app/controllers/foreman_tasks/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/tasks_controller.rb
@@ -98,7 +98,7 @@ module ForemanTasks
     end
 
     def auto_complete_controller_name
-      '/foreman_tasks/tasks'
+      'foreman_tasks_tasks'
     end
 
     private


app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -98,7 +98,7 @@ def resource_class
     end
 
     def auto_complete_controller_name
-      '/foreman_tasks/tasks'
```
iNecas: This code has actually been added in #375 to workaround regression introduced in theforeman/foreman#5771. It seems that it has not fixed the issue properly.

In order for this fix to work, we need also theforeman/foreman#6460.

After that is merged, we can just remove the auto_complete_controller_name method complete from here, so that it would work as before #375.

iNecas:
Thanks alan for reporting and fixing the issue. Unfortunately, we also need a change in foreman to make this work. After that is done, we don't need `auto_complete_controller_name` method at all.

app/controllers/foreman_tasks/tasks_controller.rb
iNecas: The foreman PR has just been merge. Mind updating the PR to just remove this method completely? That should fix the whole problem.

iNecas:
Closing in favor of https://github.com/theforeman/foreman-tasks/pull/388/files :  currently the tasks are broken in nightlies: we should be moving quickly to unblock other developers

PR #389, opened 2019-02-08
Fixes #26014 - don't use Rails truncate in breadcrumb

iNecas:
The truncation does additional html escaping that doesn't work well
will JS's one. Since https://projects.theforeman.org/issues/25503, we
don't need to do explicit truncation in breadrumbs anymore, so we can
just get rid of it.


iNecas:
[test foreman-tasks]

PR #390, opened 2019-02-20
Fixes #26073 - Adding react infrastructure

Laviro:


sharvit:
**TODO:**
- [ ] Use babel configurations as core uses
- [ ] Use eslint configurations as core uses
- [ ] Use travis configurations as core uses
- [ ] Use Jest configurations as core uses
- [ ] Create hello-world react-component
- [ ] Test the hello-world component with Jest
- [ ] Add storybook with the hello-world component

webpack/global_test_setup.js
```diff
@@ -0,0 +1,6 @@
+// runs before each test to make sure console.error output will
+// fail a test (i.e. default PropType missing). Check the error
+// output and traceback for actual error.
+global.console.error = (error) => {
```
sharvit: I don't sure we actually need this file, in core it just works fine

webpack/.eslintrc
```diff
@@ -0,0 +1,35 @@
+{
```
sharvit: Better to use the core configurations that use pf-react-eslint

webpack/global_test_setup.js
Laviro: thanks, removed it and instead using only `test_setup.js`

webpack/.eslintrc
Laviro: using `patternfly-react-eslint` now

sharvit:
Based on this core pr https://github.com/theforeman/foreman/pull/6515

app/controllers/foreman_tasks/react_controller.rb
```diff
@@ -0,0 +1,11 @@
+module ForemanTasks
+  class ReactController < ::ApplicationController
```
ohadlevy: why is that needed?

package.json
```diff
@@ -0,0 +1,176 @@
...
+    "stylelint-config-standard": "^18.0.0",
+    "uglifyjs-webpack-plugin": "^1.2.2",
+    "url-loader": "^1.0.1",
+    "webpack": "^3.4.1",
+    "webpack-bundle-analyzer": "^2.13.1",
+    "webpack-dev-server": "^2.5.1",
+    "webpack-stats-plugin": "^0.1.5"
+  },
+  "dependencies": {
+    "@novnc/novnc": "^1.0.0",
```
ohadlevy: please don't copy all of the packages from foreman as they are not all needed here ^

Laviro: thanks, sure.. it's just that many things broke when I removed some of them.. so i wanted to make sure everything works before.

Laviro:
Storybook is working now:
selection_018

Laviro:
if you'll go to the `/foreman_tasks/ex_tasks` (experimental page), you will see:
selection_020

app/controllers/foreman_tasks/react_controller.rb
sharvit: We want to implement it with `react-router`

Laviro:
@sharvit thanks for all of your commits in this PR !!

app/controllers/foreman_tasks/react_controller.rb
ohadlevy: ack, lets finalize the way its supposed to be used in foreman first? ( i would assume as a plugin you would reuse that infra)

sharvit: agree, imo foreman should provide `registerRoutes` so we won't need to use turbolinks when moving between foreman-routes to plugin-routes.

Laviro:
@sharvit Thanks for your help on this, couldn't do this without you

iNecas:
Awesomesauce. I need to test it + figure out how to skip the auth tests, but hopefully could get in by tomorrow.

app/controllers/foreman_tasks/react_controller.rb
waldenraines: @sharvit how much work would it be to do the above?

app/controllers/foreman_tasks/react_controller.rb
```diff
@@ -1,11 +1,17 @@
 module ForemanTasks
   class ReactController < ::ApplicationController
-    skip_before_action :authorize
```
iNecas: Added authorization, to make tests green, but it also probably make sense to not expose it without having tasks permissions (although the permissions will be covered by API)

iNecas:
Travis has been enabled for this repo

Laviro:
ESlint is working now,, 
I saw how Katello managed this issue.

iNecas:
Awesome. I've pushed one more fix regarding permissions + rebased against master because there were some tests failing due to other issue we've fixed in the mean time. Once tests are green, this will go in.

lib/foreman_tasks/engine.rb
```diff
@@ -57,7 +57,9 @@ class Engine < ::Rails::Engine
 
         security_block :foreman_tasks do |_map|
           permission :view_foreman_tasks, { :'foreman_tasks/tasks' => [:auto_complete_search, :sub_tasks, :index, :show],
-                                            :'foreman_tasks/api/tasks' => [:bulk_search, :show, :index, :summary] }, :resource_type => ForemanTasks::Task.name
+                                            :'foreman_tasks/react' => [:index],
+                                            :'foreman_tasks/api/tasks' => [:bulk_search, :show, :index, :summary],
```
adamruzicka: Rubocop is not happy about changes in this file

iNecas: Should be fixed now

iNecas:
All almost green (test failure unrelated). Thanks @Laviro @sharvit @ohadlevy

iNecas:
Squashed and merged in a040206

PR #391, opened 2019-02-28
Refs #22602 - remove non-uuid value dynflow-related id

iNecas:
Tests started failing after merging
https://github.com/Dynflow/dynflow/pull/319


adamruzicka:
Thanks @iNecas !

PR #392, opened 2019-03-04
Sample actions for tasks development

iNecas:
A set of actions with pre-defined behaviour for easier simulation of
various tasks states: these actions are development only.

### DIFF
--- a/app/lib/actions/try.rb
+++ b/app/lib/actions/try.rb
@@ -1,0 +1,51 @@
+module Actions
+  module Try
+    # A set of actions to try different scenarios with tasks in development.
+    #
+    # Usage:
+    # from rails console, run:
+    #
+    #    # Successful task (ending in stopped state/success result)
+    #    ForemanTasks.sync_task(Actions::Try::Success)
+    #
+    #    # Failing task task (ending in paused state/error result)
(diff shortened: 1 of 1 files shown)


app/lib/actions/try.rb
```diff
@@ -0,0 +1,51 @@
...
+    #
+    #    # Successful task (ending in stopped state/success result)
+    #    ForemanTasks.sync_task(Actions::Try::Success)
+    #
+    #    # Failing task task (ending in paused state/error result)
+    #    ForemanTasks.sync_task(Actions::Try::Pause)
+    #
+    #    # Task that fails but resumes (ending in stopped state/warning result)
+    #    ForemanTasks.sync_task(Actions::Try::Skip)
+    unless Rails.env.production?
```
adamruzicka: I'd move this one level higher so we won't end up with empty `Actions::Try` module in production environments

adamruzicka:
@ezr-ondrej opinions?

adamruzicka:
This has been open for ages and while it could be useful in some situations, it is not really necessary to have this in. With that being said I'll go ahead and close this, rather than leaving it hanging again.

PR #396, opened 2019-03-13
Fixes #26352 - ActiveJob task arguments loading

ezr-ondrej:
ActiveJob arguments loading is broken by: https://github.com/Dynflow/dynflow/pull/317

### DIFF
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -83,7 +83,7 @@ module ForemanTasks
     end
 
     def input
-      return execution_plan_action.input['job_arguments'] if active_job?
+      return active_job_data['arguments'] if active_job?
       main_action.task_input if main_action.respond_to?(:task_input)
     end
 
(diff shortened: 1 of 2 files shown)


app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -118,7 +118,8 @@ def main_action
                else
                  execution_plan_action.input
                end
-        @main_action = active_job_action(args['job_class'], args['job_arguments'])
+        args = args['job_data'] || args.slice('job_class')
```
iNecas: For the tasks created before this change, shouldn't this be:

```ruby
args = args['job_data'].try { |data| data['arguments'] } || args['job_arguments']
@main_action = active_job_action(args['job_class'], args['arguments'])
```

ezr-ondrej: Yes, you are right :+1: thanks.

ezr-ondrej:
I believe now we are backwards compatible :)

app/models/foreman_tasks/task/dynflow_task.rb
iNecas: One small ask: please add a comment that we need the fallback for `job_arguments` for fobs from Dynflow `<=1.2.1`, for the future version of us that would wonder why we're doing this here :)

iNecas:
There is one more occurrence of `job_arguments` in https://github.com/theforeman/foreman-tasks/blob/1c853d37cd1a7a66ea07b57280046070dbcc5041/app/models/foreman_tasks/task/dynflow_task.rb#L86, would be good to fix that one as well while touching it.

ezr-ondrej:
I have pulled the logic into its method, just left one comment, where I am unsure.

iNecas:
Thanks @ezr-ondrej

PR #397, opened 2019-03-15
Refs #25160 - fix disabling of batch triggering

iNecas:
also make the default for batch size a bit smaller: the planning
of a job on host can sometimes be quite long-running process, and
the larger the batch size is, the longer we would wait until something
would start happening.


adamruzicka:
Thanks @iNecas !

PR #399, opened 2019-03-18
Fixes #26406 - Adds tasks dashboard state

sharvit:


sharvit:
Thanks @waldenraines, rebased

iNecas:
With some smaller widths: the layout of the cards gets broken, see:

task-dashbaord-width

iNecas:
Also, occasionally, I'm getting:

```
Uncaught TypeError: Cannot read property 'selectAll' of null
    at c3.js:4562
c3.js:9223 Uncaught TypeError: Cannot read property 'data_types' of null
    at ChartInternal.c3_chart_internal_fn.isPieType (c3.js:9223)
    at ChartInternal.c3_chart_internal_fn.isArcType (c3.js:9234)
    at ChartInternal.c3_chart_internal_fn.getArc (c3.js:4473)
    at c3.js:4816
    at SVGPathElement.<anonymous> (d3.js:8775)
    at Object.tick [as c] (d3.js:8956)
    at d3_timer_mark (d3.js:2166)
    at d3_timer_step (d3.js:2147)
```
especially after clicking part of some chart

iNecas:
The backend part is here: https://github.com/theforeman/foreman-tasks/pull/407

I did some updates in the parameter naming, namely:

* time -> time_horizon
* mode -> time_mode
* time_mode/last -> time_mode/recent

iNecas:
@sharvit could we merge the `mode = ` and `time = h24` in something like `state updated within last 24 hours` / `state updated before 24 hours`

sharvit:
> Also, occasionally, I'm getting:
> 
> ```
> Uncaught TypeError: Cannot read property 'selectAll' of null
>     at c3.js:4562
> c3.js:9223 Uncaught TypeError: Cannot read property 'data_types' of null
>     at ChartInternal.c3_chart_internal_fn.isPieType (c3.js:9223)
>     at ChartInternal.c3_chart_internal_fn.isArcType (c3.js:9234)
>     at ChartInternal.c3_chart_internal_fn.getArc (c3.js:4473)
>     at c3.js:4816
>     at SVGPathElement.<anonymous> (d3.js:8775)
>     at Object.tick [as c] (d3.js:8956)
>     at d3_timer_mark (d3.js:2166)
>     at d3_timer_step (d3.js:2147)
> ```
> especially after clicking part of some chart

Fixed :+1:

sharvit:
> With some smaller widths: the layout of the cards gets broken, see:
> 
> task-dashbaord-width

Any suggestion for what to do for smaller layouts?
atm it will change from a horizontal list of cards to a vertical list. (which is not ideal as well?)
We can fix the charts by removing the right legends when the screen is small but we will still get into issues with the stopped card table.

What would you think about not showing the dashboard when the screen size is not big enough?

iNecas:
I would find it acceptable for the cards and donuts to be fixed size, and we should just decide if we have enough space to put it in one row, or wrap them in multiple columns. I don't like the solution, where we would not show it for smaller widths: I'm personally using one monitor in portrait position, and I send negative energy to anyone that doesn't take `1080` width as something to count on when designing the site :)

iNecas:
What about having 2x2 layout for smaller widths?

sharvit:
Rebased and integrate with #407

sharvit:
> What about having 2x2 layout for smaller widths?

Added 2x2 layout for medium screen sizes

sharvit:
There are many missing tests to write before it can be merged. I will confirm once it will be covered well.

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/ScheduledTasksCard/ScheduledTasksCard.js
```diff
@@ -32,7 +32,12 @@ const ScheduledTasksCard = ({
     >
       <Card.Title onClick={onClick}>{__('Scheduled')}</Card.Title>
       <Card.Body>
-        <div className="scheduled-data" onClick={onClick}>
+        <div
+          className={classNames('scheduled-data', {
+            'not-focused': query.state && query.state !== SCHEDULED,
```
Laviro: I guess we don't need double checking for `query.state`, `query.state !== SCHEDULED` would be enough

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
```diff
@@ -46,45 +49,54 @@ const StoppedTasksCard = ({
...
-                      mode: LAST,
-                      time,
-                    })
-                  }
-                >
-                  {last}
-                </td>
-              </tr>
-            ))}
+            {Object.entries(data).map(([result, { total, last }]) => {
```
Laviro: could we move it into an helper file instead?

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardActions.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+};
+
+const startRequest = () => ({
+  type: FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_REQUEST,
+});
+
+const requestSuccess = data => ({
+  type: FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_SUCCESS,
+  payload: data,
```
Laviro: suggestion: destruct the data object so it would be easier to know what passes in the payload.

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardConstants.js
```diff
@@ -24,8 +24,71 @@ export const TASKS_DASHBOARD_AVAILABLE_TIMES = {
...
+  'FOREMAN_TASKS_DASHBOARD_UPDATE_QUERY';
+
+export const FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_REQUEST =
+  'FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_REQUEST';
+export const FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_SUCCESS =
+  'FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_SUCCESS';
+export const FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_FAILURE =
+  'FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_FAILURE';
+
+export const TASKS_SUMMARY_ZERO = {
```
Laviro: maybe rename it to `INITIAL` ?

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/ScheduledTasksCard/ScheduledTasksCard.js
sharvit: We do, it should be `not-focused` if there is something in `query.state` but it is not `SCHEDULED`.

sharvit: If `query.state` is empty, we should not add `not-focused`

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardActions.js
sharvit: not sure I understand, can you explain please?

Laviro: For example: 
```
const requestSuccess = ( { running, paused, stopped, scheduled } ) => ({
  type: FOREMAN_TASKS_DASHBOARD_FETCH_TASKS_SUMMARY_SUCCESS,
  payload: {
    running, 
    paused, 
    stopped, 
    scheduled
  },
```

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/ScheduledTasksCard/ScheduledTasksCard.js
Laviro: I think we are talking on the same thing, 
empty query state is not equal to `SCHEDULED` which means it will return false,
I am just suggesting writing: ` 'not-focused': query.state !== SCHEDULED,` instead.

sharvit: We are not, `not-focused` should be `false` when `query.state` is empty and `undefined !== SCHEDULED` will return `true`.

sharvit: In other words, `not-focused` shot be true only when `query.state` **has a value** and this value is different from `SCHEDULED`.
In other words, `not-focused` mean **another card is focused**

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
```diff
@@ -46,45 +49,54 @@ const StoppedTasksCard = ({
...
+                        mode: LAST,
+                        time,
+                      })
+                    }
+                  >
+                    {last}
+                  </td>
+                </tr>
+              );
+            })}
```
Laviro: could the data already load when the page refreshes, instead of starting with 0 and than update?
if it is a problem for now, I am fine with adding it in a follow up PR :)

Apr 8, 2019 2_59 PM

sharvit: Yes it makes sense but I would prefer to implement SPA and get this feature automatically with a future PR instead implementing it for a none SPA and then remove it when we introduce SPA.

sharvit:
> UI tests are failling currently. Also, the legend has disappeared for some reason:
> 
> tasks-missing-legend

Test failures will be fixed soon with a new commit that adds more tests coverage. (then it can be merged from my side)

About the legends, I changed it so they should be hidden when one of the values are 0.
I can change it so they will be hidden only when both of the values are set to 0. That would make more sense in term of UX? @terezanovotna

iNecas:
Merged as ef164ce467f28eb90e7a43a7347269c66ea08254, thanks @MariaAga, @sharvit and @waldenraines

PR #400, opened 2019-03-18
Fixes #26399 - add summary endpoint to support tasks dashboard

iNecas:
To support showing new tasks dashboard, we need the following endpoint:

```
GET /foreman_tasks/tasks/summary/24
{ running:   { recent: 3, total: 6 },
  paused:    { recent: 1, total: 3 },
  stopped:   { recent: 3, total: 7,
               by_result: {
                success: { recent: 2, total: 4 },
                warning: { recent: 1, total: 2 },
                error:   { recent: 0, total: 1 },
               }}
  scheduled: { recent: 0, total: 3 }}
```

### DIFF
--- a/test/unit/summarizer_test.rb
+++ b/test/unit/summarizer_test.rb
@@ -52,12 +52,8 @@ class SummarizerTest < ActiveSupport::TestCase
             task.update(state_updated_at: nil)
           end
         end
-        recent_tasks = []
-        status_summary[:recent].times do |i|
-          task = tasks[i]
-          task.update(state_updated_at: Time.now.utc)
-          recent_tasks << task
-        end
+        recent_tasks = tasks.take(status_summary[:recent])
+        recent_tasks.each { |t| t.update(state_updated_at: Time.now.utc) }
 
         untouched_recent_tasks = recent_tasks
         untouched_old_tasks = tasks - recent_tasks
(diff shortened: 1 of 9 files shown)


iNecas:
@MariaAga here's the endpoint to provide the dashboard data. The only thing changed from the original proposal is the results for stopped state to be nested behind `by_result` attribute

iNecas:
Test failures unrelated

test/unit/summarizer_test.rb
```diff
@@ -0,0 +1,79 @@
...
+                                    result: 'pending').tap do |task|
+            task.update(state_updated_at: nil)
+          end
+        end
+        recent_tasks = []
+        status_summary[:recent].times do |i|
+          task = tasks[i]
+          task.update(state_updated_at: Time.now.utc)
+          recent_tasks << task
+        end
```
adamruzicka:
```ruby
recent_tasks = tasks.take(status_summary[:recent])
recent_tasks.each { |t| t.update(state_updated_at: Time.now.utc)
```
Wouldn't something like this be a bit cleaner? Something like this could be done several times in this file

iNecas: Nice suggestion. Updated

iNecas:
Comments should be addressed

PR #404, opened 2019-03-28
Refs #26406 - Update card tasks to use query and updateQuery

sharvit:
Move the responsibility of reading and updating the `query` to the cards components.

Update the components:
- TasksDonutChart
- TasksDonutCard
- RunningTasksCard
- PausedTasksCard

[storybook](https://superb-boot.surge.sh)


iNecas:
Rebased and merged

PR #408, opened 2019-04-04
Refs #22602 - Add implicit casting from VARCHAR to UUID on postgres

adamruzicka:
Before this patch foreman_tasks:cleanup rake task would fail when trying
to delete orphaned execution plans.

The error was:
```
Sequel::DatabaseError: PG::UndefinedFunction: ERROR: operator does not exist:
uuid = character varying LINE 1: ...unt" FROM "dynflow_execution_plans"
WHERE ("uuid" NOT IN (SE...
```


iNecas:
Tested and works well. Thanks @adamruzicka

PR #409, opened 2019-04-05
Fixes #26544 - task export to contain task uuid

pmoravec:
to let uniquely identify individual tasks in the export

Signed-off-by: Pavel Moravec <pmoravec@redhat.com>


adamruzicka:
Thanks @pmoravec !

PR #411, opened 2019-04-08
Fixes #26560 - paused tasks notification and troubleshooting info

iNecas:
This path adds notifications and troubleshooting info for pauses tasks

Troubleshooting info
--------------------

A new `TroubleshootingHelpGenerator` service class has been added to
provide troubleshooting info for paused tasks. By default, it provides
generic information explaining the situation and link to troubleshooting
documentation (in case Setting[:foreman_tasks_troubleshooting_url] is
set.

The troubleshooting_url can contain macros for foreman version and
normalized task label (the label without non-alhpanumeric characters).
For example, for the following url setting:

      https://tfm.org/%{version}/tasks_troubleshooting.html#%{label}

and task 'Plugin::MyTask', the generated url would be:

      https://tfm.org/1.22/tasks_troubleshooting.html#PluginMyTask

The setting is currently not set (which means the troubleshooting link to
currently generated. We will add the setting once the page is ready in
docs.

It's also possible to extend the troubleshooting info and links on
per-action basis, by defining specific action method (see
`test/support/dummy_dynflow_action.rb` for example on how to define it.

Admin notifications
-------------------

Every time there is paused task, an admin gets a notification about
current amount of paused tasks with links to:

1. list of paused tasks
2. troubleshooting links from `TroubleshootingHelpGenerator`
  - the label for the troubleshooting url is filled with `''` in this
  case.

In case new paused task is added, the previous notification is removed
and an update notification is generated, so that the user's notification
is not overloaded with data.

Owner notification
------------------

In case the task is triggered by some user (the host-related tasks are
out of scope), the owner should the notification about the single task
being paused, with links to

1. details of the task
2. troubleshooting info for the task

How to test
-----------

1. set the `foreman_tasks_troubleshooting_url` config to some url
(as example might be
https://access.redhat.com/solutions/satellite6-tasks)
2. make some tasks to get paused
3. look at the notification drawer
4. loot the the details of the task

### DIFF
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -148,6 +148,22 @@ module ForemanTasks
       end
     end
 
+    def notification_recipients_ids
+      owner_ids
+    end
+
+    def build_notifications
+      notifications = []
+      if state == 'paused'
+        owner = self.owner
+        if owner && !owner.hidden?
+          notifications << UINotifications::Tasks::TaskPausedOwner.new(self)
+        end
+        notifications << UINotifications::Tasks::TaskPausedAdmin.new(self)
+      end
+      notifications
+    end
+
     def self.search_by_generic_resource(key, operator, value)
       key = 'resource_type' if key.blank?
       key_name = connection.quote_column_name(key.sub(/^.*\./, ''))
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -148,13 +148,14 @@ module ForemanTasks
       end
     end
 
+    # used by Foreman notifications framework
     def notification_recipients_ids
       owner_ids
     end
 
     def build_notifications
       notifications = []
-      if state == 'paused'
+      if paused?
         owner = self.owner
         if owner && !owner.hidden?
           notifications << UINotifications::Tasks::TaskPausedOwner.new(self)
--- a/test/foreman_tasks_test_helper.rb
+++ b/test/foreman_tasks_test_helper.rb
@@ -7,12 +7,15 @@ require_relative './support/dummy_task_group'
 require_relative './support/history_tasks_builder'
 
 require 'dynflow/testing'
+require 'foreman_tasks/test_helpers'
 
 FactoryBot.definition_file_paths = ["#{ForemanTasks::Engine.root}/test/factories"]
 FactoryBot.find_definitions
 
 ForemanTasks.dynflow.require!
 ForemanTasks.dynflow.config.disable_active_record_actions = true
+
+ForemanTasks::TestHelpers.use_in_memory_sqlite!
 
 # waits for the passed block to return non-nil value and reiterates it while getting false
 # (till some reasonable timeout). Useful for forcing the tests for some event to occur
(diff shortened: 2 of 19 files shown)


app/models/foreman_tasks/task.rb
```diff
@@ -148,6 +148,22 @@ def self_and_parents
       end
     end
 
+    def notification_recipients_ids
```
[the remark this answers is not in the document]
iNecas: This is a notifcations thing https://github.com/theforeman/foreman/blob/745ff85013db740b093a4c8a86e4ef3285f383f1/app/models/notification.rb#L47

adamruzicka:
Needs a rebase

iNecas:
Rebased: Regarding the notification on sub-action, I lean towards right now limiting us on the main action only: the troubleshooting docs can talk more about failing the specific sub-action in the text, but we're mainly focused on how to get the paused task resolved ~ the main action. I will try to fix that.

app/models/foreman_tasks/task.rb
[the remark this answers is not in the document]
iNecas: Added explanatory comment in the code

iNecas:
@adamruzicka all the comments should be addressed

iNecas:
Also, after offline discussion, I've removed the trimming of `::` from the label for the anchor purposes

adamruzicka:
Still APJ

iNecas:
Seems like test failures are related for some weird reason

test/foreman_tasks_test_helper.rb
```diff
@@ -7,13 +7,16 @@
...
 require 'dynflow/testing'
+require 'foreman_tasks/test_helpers'
 
 FactoryBot.definition_file_paths = ["#{ForemanTasks::Engine.root}/test/factories"]
 FactoryBot.find_definitions
 
 ForemanTasks.dynflow.require!
 ForemanTasks.dynflow.config.disable_active_record_actions = true
 
+ForemanTasks::TestHelpers.use_in_memory_sqlite!
```
iNecas: Seems like using the same db for standard test world and the in-thread executor one was causing issues: this should make the things more isolated.

iNecas:
The tests seems to be much happier now. Should by ready for final review

test/unit/actions/action_with_sub_plans_test.rb
```diff
@@ -35,7 +35,7 @@ def run; end
         user = FactoryBot.create(:user)
         triggered = ForemanTasks.trigger(ParentAction, user)
         raise triggered.error if triggered.respond_to?(:error)
-        triggered.finished.wait(2)
+        triggered.finished.wait(30)
```
adamruzicka: Does this need to stay increased when the root cause was resolved?

adamruzicka:
[test foreman_tasks]

adamruzicka:
[test foreman-tasks]

test/unit/actions/action_with_sub_plans_test.rb
iNecas: I would leave it there

iNecas:
All green now

iNecas:
@adamruzicka anything else to do here or ok to merge?

adamruzicka:
Nope, looks good to me

adamruzicka:
Thanks @iNecas !

PR #414, opened 2019-04-21
Fixes #26653 - cards-grid doesnt supports all screen sizes

MariaAga:
The issue is that the cards aren't shown as they should when viewport width is 1200-1500.
Please note that in the storybook the grid is rendered inside an iframe, which causes the viewport to be only the grid container, but in foreman the viewport is the whole page.
So in foreman, when the screen is 1300px this is the view:
image
but in the storybook:
image

Chose to break point of 1530px because on 1518px this is the stopped table :
1518pxCARD
and in 1530px the icons are shown as intended:
image

### DIFF
--- a/webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/TasksCardsGrid.stories.js
+++ b/webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/TasksCardsGrid.stories.js
@@ -37,16 +37,18 @@ storiesOf('TasksDashboard/TasksCardsGrid', module)
     );
 
     return (
-      <TasksCardsGrid
-        time={selectTime}
-        query={{
-          state: selectState,
-          result: selectResult,
-          mode: selectMode,
-          time: selectTime,
-        }}
-        data={object('data', MOCKED_DATA)}
-        updateQuery={action('updateQuery')}
-      />
+      <div>
+        <TasksCardsGrid
+          time={selectTime}
+          query={{
+            state: selectState,
+            result: selectResult,
+            mode: selectMode,
+            time: selectTime,
+          }}
+          data={object('data', MOCKED_DATA)}
+          updateQuery={action('updateQuery')}
+        />
+      </div>
     );
   });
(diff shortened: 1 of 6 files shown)


webpack/ForemanTasks/Components/TasksDashboard/TasksDashboard.scss
```diff
@@ -1,6 +1,27 @@
+
```
sharvit: nitpik: can you please remove this empty line and add a new empty line at the end of the file?

sharvit: Does it make sense to move this code into `TasksDashboard/Components/TasksCardsGrid/TasksCardsGrid.scss`?

MariaAga: I'm not sure, as it refers to the `tasks-dashboard-grid` class, which is used only in `TasksDashboard.js`

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/TasksCardsGrid.stories.js
```diff
@@ -37,16 +37,23 @@ storiesOf('TasksDashboard/TasksCardsGrid', module)
...
-          state: selectState,
-          result: selectResult,
-          mode: selectMode,
-          time: selectTime,
-        }}
-        data={object('data', MOCKED_DATA)}
-        updateQuery={action('updateQuery')}
-      />
+      <div>
+        <link
```
sharvit: Do we actually need this `link`?
Can't we add the `patternfly-additions` to https://github.com/theforeman/foreman-tasks/blob/master/webpack/stories/index.scss?

MariaAga: added the font awesome to the file instead, thanks!

adamruzicka:
[test foreman_tasks]

adamruzicka:
[test foreman-tasks]

adamruzicka:
@MariaAga could you please rebase this? Just to make sure the tests are ok

MariaAga:
rebased

adamruzicka:
Thanks @MariaAga and @sharvit !

PR #416, opened 2019-04-26
Fixes #26692 - allow passing partial updates to runner

iNecas:


lib/foreman_tasks_core/otp_manager.rb
```diff
@@ -17,10 +17,12 @@ def passwords
...
       end
 
-      def authenticate(hash)
+      def authenticate(hash, clear: true)
         plain = Base64.decode64(hash)
         username, otp = plain.split(':', 2)
-        drop_otp(username, otp)
+        password_matches = passwords[username] == otp
+        drop_otp(username, otp) if clear
+        password_matches
```
adamruzicka: How about something like
```ruby
password_matches = passwords[username] == otp
passwords.delete(username) if clear && password_matches
password_matches
```
to avoid comparing the otp twice (and an unnecessary function call)

iNecas:
All comments updated: the tests should get green soon hopefully

iNecas:
Test failures unrelated: @adamruzicka this should be good to go now

iNecas:
I've put the `drop_otp` method back, as the rex is using it, and it kind of makes sense

adamruzicka:
[test foreman-tasks]

adamruzicka:
Could you rebase this? The test failures should be fixed by https://github.com/theforeman/foreman-tasks/pull/424

lib/foreman_tasks_core/otp_manager.rb
```diff
@@ -9,18 +9,19 @@ def generate_otp(username)
...
       end
 
-      def authenticate(hash)
+      def authenticate(hash, expected_user: nil, clear: true)
         plain = Base64.decode64(hash)
         username, otp = plain.split(':', 2)
-        drop_otp(username, otp)
+        if expected_user
+          return false unless expected_user == username
+        end
```
adamruzicka: Any reason why not to combine these two conditions into one?

iNecas: Reading some golang code recently :) but I like the two conditions - one checking if we actually care about the equality, and the second for the matching itself. So I prefer to keep it, but I can compress it to one if that's too much for you :)

adamruzicka: No problem with it if was intentional. It just felt like there were originally more things in the outer condition and this is a leftover.

adamruzicka:
Jenkins approves, waiting for other related PRs to be ready

adamruzicka:
Thanks @iNecas !

PR #418, opened 2019-04-30
Fix typo in the gemspec summary

ekohl:


iNecas:
What do you have against `resoruces`?

iNecas:
Thanks @ekohl

ekohl:
> What do you have against `resoruces`?

Nothing against resoruces but I think you can't have enough resources :)

PR #419, opened 2019-05-02
Fixes #26714 - Fix keeping user when running hooks

adamruzicka:


iNecas:
It has been this change https://github.com/theforeman/foreman-tasks/commit/a265d9293cba23f59c442d4256c3836eb58f5360#diff-2351f5677e6b47c13e41d7f98e606471R5 - the thing is the logging hooks are running even before the planning happens, so the `restore_current_taxonomies` ended up cleaning the `User.current` value before it could be stored inside the input.

PR #420, opened 2019-05-06
Refs #26073 - fix params docs

iNecas:
Without this patch, the docs for callback says something like:

```
<p>Must be an array of [[:callback, Hash, nil,
{:param_group=&gt;{:scope=&gt;ForemanTasks::Api::TasksController,
:name=&gt;:callback_target, :options=&gt;{}, :from_concern=&gt;false}},
/.../foreman-tasks/app/controllers/foreman_tasks/api/
tasks_controller.rb:152&gt;],
[:data, Hash, {:desc=&gt;“Data to be sent to the action”},
{:param_group=&gt;nil}, nil], [:callback, Hash, nil,
{:param_group=&gt;{:scope=&gt;ForemanTasks::Api::TasksController,
```

Among others, this actually causes RPM packaging to fail, because
it leaks the buildroot path inside the output.

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -156,13 +156,15 @@ module ForemanTasks
       end
 
       def_param_group :callback do
-        param_group :callback_target
+        param_group :callback_target, TasksController
         param :data, Hash, :desc => N_('Data to be sent to the action')
       end
 
       api :POST, '/tasks/callback', N_('Send data to the task from external executor (such as smart_proxy_dynflow)')
       param_group :callback
-      param :callbacks, Array, :of => param_group(:callback)
+      param :callbacks, Array do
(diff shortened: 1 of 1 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -156,13 +156,15 @@ def index
       end
 
       def_param_group :callback do
-        param_group :callback_target
+        param_group :callback_target, TasksController
```
adamruzicka: Does this really need to be specified? According to the docs the lookup should be done within the current controller

iNecas: Yes it does, when used inside definition of `Array` block: I've found the hard way

adamruzicka:
APJ

adamruzicka:
Thanks @iNecas !

PR #421, opened 2019-05-07
Refs #22602 - explicit instead of implicit casting UUID casting

iNecas:
In production setup, we were getting:

     create cast must be owner of type character varying

The problem is this casting seems to be global, and we might not want
to do that for every case, and we don't have permissions for that.


test/unit/cleaner_test.rb
```diff
@@ -32,6 +32,27 @@ class TasksTest < ActiveSupport::TestCase
                   .find_execution_plans(filters: { 'uuid' => tasks_to_keep.map(&:external_id) }).size.must_equal tasks_to_keep.size
     end
 
+    describe "#orphaned_dynflow_tasks" do
+      # We can't use transactional tests because we're using Sequel for the cleanup query
+      self.use_transactional_tests = false
+      before do
+        skip "Sqlite is running testing Dynlfow DB in memory" if ActiveRecord::Base.connection.adapter_name == 'SQLite'
```
iNecas: We can't test the cleaner in sqite, because ActiveRecord and Sequel are running in separate DBs

iNecas:
JP (well, one is still running, but it already passed the tasks tests)

adamruzicka:
Alright, let's get it it

adamruzicka:
Thanks @iNecas !

PR #422, opened 2019-05-19
Fixes #26827 - Add name to menu divider

tbrisker:
Otherwise, it will be filtered out and the tasks menu items will look
like they belong to the previous submenu.


adamruzicka:
Thanks @tbrisker !

PR #425, opened 2019-06-12
Fixes #26349 - Create tasks details page in react

MariaAga:
The reload Button now only re-fetches the task data instead the whole html
The time calculation has slightly changed and now follows  http://cldr.unicode.org/ format

wip for:
- missing tests
- ruby 'leftovers'


webpack/ForemanTasks/Components/TaskDetails/Components/Errors.js
```diff
@@ -0,0 +1,62 @@
...
+          </span>
+          <span className="param-name">{__('Input')}:</span>
+          <span className="param-value">
+            <pre>{step.input}</pre>
+          </span>
+          <span className="param-name">{__('Output')}:</span>
+          <span className="param-value">
+            <pre>{step.output}</pre>
+          </span>
+          {step.error ? (
```
adamruzicka: Wouldn't `{step.error && (<React.Fragment>...</React.Fragment>)}` do the same thing without the dangling empty string?

webpack/ForemanTasks/Components/TaskDetails/Components/Raw.js
```diff
@@ -0,0 +1,95 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+import { durationInWords } from './TaskHelper';
+
+class Raw extends Component {
+  add = (a, b) => {
```
adamruzicka: But why?

webpack/ForemanTasks/Components/TaskDetails/Components/Task.js
```diff
@@ -0,0 +1,324 @@
...
+      default:
+        icon = 'pficon-help';
+        break;
+    }
+    return `task-status ${icon}`;
+  };
+
+  progressBarType = () => {
+    const { result } = this.props;
+    if (result === 'error') return 'danger';
```
adamruzicka: A `switch` maybe?

webpack/ForemanTasks/Components/TaskDetails/Components/TaskHelper.js
```diff
@@ -0,0 +1,54 @@
+import { translate as __ } from 'foremanReact/common/I18n';
+import React from 'react';
+import { FormattedRelative } from 'react-intl';
+
+const timeSince = date => {
```
adamruzicka: Isn't there a library which does this?

MariaAga: I'm not sure its a good idea to import a whole library for something that can be solved in 20 lines

webpack/ForemanTasks/Components/TaskDetails/Components/Task.js
MariaAga: Yes, thanks!

webpack/ForemanTasks/Components/TaskDetails/Components/Raw.js
MariaAga: Forgot to remove :)
Removed now

webpack/ForemanTasks/Components/TaskDetails/Components/Errors.js
MariaAga: Youre right, thanks!

adamruzicka: There are still some places where this could be applied

MariaAga: I changed them as well, thanks

MariaAga:
Fixed the rubocop issues

webpack/ForemanTasks/Components/TaskDetails/Components/Raw.js
```diff
@@ -0,0 +1,73 @@
...
+  </div>
+);
+
+Raw.propTypes = {
+  id: PropTypes.string,
+  label: PropTypes.string,
+  startedAt: PropTypes.string,
+  endedAt: PropTypes.string,
+  input: PropTypes.oneOfType([PropTypes.array, PropTypes.shape({})]),
+  output: PropTypes.oneOfType([PropTypes.string, PropTypes.shape({})]),
```
adamruzicka: @iNecas is there a possibility for `input` and `output` of a task to be either an `array` or a `string`? I though it always is a hash/object

iNecas: AFAIK it should be always hash/object, as you say

MariaAga: The output is my mistake, but I have some tasks that have:
 "input": [] or       
"input": [{"_aj_hash_with_indifferent_access": true}]

adamruzicka: Interesting, looks like ActiveJob handles things a bit differently, nevermind then.

MariaAga:
@adamruzicka I couldn't reproduce the date issues, can you attach the API data from the task:
/foreman_tasks/api/tasks/TASK_ID/details
?

adamruzicka:
Here you are https://gist.github.com/adamruzicka/c11e6f9e3d3deec6bfb2d970c944b566

MariaAga:
@adamruzicka I loaded all of your data and still didn't encounter this issue, can you maybe re-check?

adamruzicka:
It seems it works in chrome (left, `google-chrome-stable-75.0.3770.90-1`), but doesn't work in firefox (right, `firefox-67.0-4`)
image

MariaAga:
Apparently Firefox can't format dd-mm-yyyy date like chrome, and that's the format that Ruby sends. Fixed this and the styling issues, thanks for the help @adamruzicka !
I still didn't receive the console error (not in Firefox or in Chrome)

adamruzicka:
When the window is narrow, every other column is right aligned.

When the window is wide, the content doesn't cover the entire available space. I'm not saying that's bad, just that it is different from the previous behavior.

https://imgur.com/a/LWQbHRy

MariaAga:
Thanks for bringing this to my attention! It looks wrong to me so I changed it and I think it looks better now

MariaAga:
@adamruzicka added that the component gets an initial_state from ruby to check if its stopped, thanks!

adamruzicka:
Needs a rebase

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -99,6 +99,32 @@ def running_steps
...
 
+    def input_output_failed_steps
+      @failed_steps = failed_steps
+      @failed_steps.map do |f|
+        {
+          error: ({ exception_class: f.error.exception_class, message: f.error.message, backtrace: f.error.backtrace } if f.error),
+        action_class: f.action_class.name,
+        state: f.state,
+        input: f.action(execution_plan).input.pretty_inspect,
+        output: f.action(execution_plan).output.pretty_inspect
```
adamruzicka: Align these, please

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -99,6 +99,32 @@ def running_steps
       execution_plan.try(:steps_in_state, :running, :suspended) || []
     end
 
+    def input_output_failed_steps
+      @failed_steps = failed_steps
+      @failed_steps.map do |f|
```
adamruzicka: Is an instance variable really needed for this?

app/views/foreman_tasks/api/tasks/details.json.rabl
```diff
@@ -0,0 +1,19 @@
+object @task if @task
+
+extends 'foreman_tasks/api/tasks/show'
+
+attributes :start_at
+node(:action) { @task.action }
```
adamruzicka: This is already provided by the show template

app/views/foreman_tasks/api/tasks/details.json.rabl
```diff
@@ -0,0 +1,19 @@
...
+node(:help) { troubleshooting_info_text }
+node(:has_sub_tasks) { @task.sub_tasks.any? }
+node(:allowDangerousActions) { Setting['dynflow_allow_dangerous_actions'] }
+node(:locks) do
+  @task.locks.map do |lock|
+    { name: lock.name, exclusive: lock.exclusive, resource_type: lock.resource_type, resource_id: lock.resource_id }
+  end
+end
+node(:username_path) { username_link_task(@task.owner, @task.username) }
+node(:parent_task_id) { @task.parent_task_id }
```
adamruzicka: Moving `:parent_task_id` to the list of attributes on line 5 would be even better

webpack/ForemanTasks/Components/TaskDetails/Components/TaskInfo.js
```diff
@@ -0,0 +1,234 @@
...
+      progress,
+      username,
+      usernamePath,
+    } = this.props;
+
+    const details = [
+      [
+        {
+          title: 'Name',
+          value: <EllipsisWithTooltip>${action}</EllipsisWithTooltip>,
```
adamruzicka: This is not string interpolation, the dollar sign appears in the final text.

MariaAga: Fixed Thanks!

app/views/foreman_tasks/api/tasks/details.json.rabl
MariaAga: moved, thanks

app/models/foreman_tasks/task/dynflow_task.rb
MariaAga: No, removed

app/models/foreman_tasks/task/dynflow_task.rb
MariaAga: Aligned, thanks

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -27,6 +27,10 @@ def summary
       param :id, :identifier, desc: 'UUID of the task'
       def show; end
 
+      api :GET, '/tasks/:id/details', 'Show task extended details'
+      param :id, :identifier, desc: 'UUID of the task'
+      def details; end
```
adamruzicka: `details` also needs to be mapped to `view` permission in `action_permission` somewhere down in this file.

webpack/ForemanTasks/Components/TaskDetails/Components/TaskHelper.js
sharvit: > I'm not sure its a good idea to import a whole library for something that can be solved in 20 lines

I'm not sure I agree in this case:

1. This file contains 60 lines of code
2. If someone else is responsible for this code (and they do a good job) it means less work for us in term of future maintenance.

I would say if there is a library which is popular, lean and maintains properly I would choose a library.
If not, I would vote for having this code in core so we can reuse it in core and in other plugins.

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsReducer.js
```diff
@@ -0,0 +1,44 @@
...
+        ...payload,
+      });
+    case FOREMAN_TASK_DETAILS_STOP_POLLING:
+      if (payload.timeoutId) {
+        clearTimeout(payload.timeoutId);
+      }
+      return state.merge({ taskReload: false, timeoutId: null });
+    case FOREMAN_TASK_DETAILS_START_POLLING:
+      if (!payload.timeoutId) {
+        payload.timeoutId = setInterval(
```
sharvit: IMO `clearTimeout` and `setInterval` should live in the action.

webpack/ForemanTasks/Components/TaskDetails/Components/TaskHelper.js
MariaAga: Uses 'humanize-duration' library now

MariaAga:
@waldenraines- I fixed this issue, thanks!

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -99,6 +99,30 @@ def running_steps
...
+    end
+
+    def input_output_running_steps
+      running_steps.map do |f|
+        {
+          action_class: f.action_class.name,
+          state: f.state,
+          input: f.action(execution_plan).input.pretty_inspect,
+          output: f.action(execution_plan).output.pretty_inspect,
+          cancellable: cancellable_action?(f.action(execution_plan))
```
adamruzicka: `f.action(execution_plan)` does a query in the db, could you rewrite this in a way that the query is done only once per step?

app/views/foreman_tasks/tasks/show.html.erb
```diff
@@ -1,5 +1,10 @@
 <% stylesheet 'foreman_tasks/tasks' %>
-
+<% content_for(:javascripts) do %>
+  <%= webpacked_plugins_js_for :'foreman-tasks' %>
+<% end %>
+<% content_for(:stylesheets) do %>
+    <%= webpacked_plugins_css_for :'foreman-tasks' %>
```
adamruzicka: too much spacing

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -99,6 +99,30 @@ def running_steps
...
 
+    def input_output_failed_steps
+      failed_steps.map do |f|
+        {
+          error: ({ exception_class: f.error.exception_class, message: f.error.message, backtrace: f.error.backtrace } if f.error),
+          action_class: f.action_class.name,
+          state: f.state,
+          input: f.action(execution_plan).input.pretty_inspect,
+          output: f.action(execution_plan).output.pretty_inspect
+        }.compact
```
adamruzicka: Is the compact needed?

app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -99,6 +99,30 @@ def running_steps
...
     end
 
+    def input_output_failed_steps
+      failed_steps.map do |f|
+        {
+          error: ({ exception_class: f.error.exception_class, message: f.error.message, backtrace: f.error.backtrace } if f.error),
+          action_class: f.action_class.name,
+          state: f.state,
+          input: f.action(execution_plan).input.pretty_inspect,
+          output: f.action(execution_plan).output.pretty_inspect
```
adamruzicka: f.action(execution_plan) does a query in the db, could you rewrite this in a way that the query is done only once per step?

app/models/foreman_tasks/task/dynflow_task.rb
MariaAga: no, thanks

app/views/foreman_tasks/tasks/show.html.erb
MariaAga: removed, thanks

app/models/foreman_tasks/task/dynflow_task.rb
MariaAga: added a variable for it

waldenraines:
[test foreman-tasks]

MariaAga:
@adamruzicka fixed rubocop issue (unrelated to this PR), added start_at to the  deteils .rabl file, started_at comes from show.rabl.
The progress % comes from the back end in this the old and new implementation, and I get different % for different running tasks.
Thanks for the review

adamruzicka:
I meant `start_before`, not `started_at`.

Ok, it is possible I didn't look at the progress at exactly same time.

MariaAga:
@adamruzicka start_before was also missing but unrelated to the delayed calc.

webpack/ForemanTasks/Components/TaskDetails/Components/Task.js
```diff
@@ -0,0 +1,199 @@
...
+                hidden={!allowDangerousActions}
+                className={`reload-button ${taskReload ? 'spin' : ''}`}
+                bsSize="small"
+                onClick={this.taskProgressToggle}
+              >
+                {__(`${taskReload ? 'Stop' : 'Start'}  auto-reloading`)}
+              </Button>
+              <Button
+                bsSize="small"
+                href={`/foreman_tasks/dynflow/${externalId}`}
```
adamruzicka: For me this points to `/foreman_tasks/dynflow/`, even if I patch the details template to return the external id.

app/views/foreman_tasks/api/tasks/details.json.rabl
```diff
@@ -0,0 +1,18 @@
+object @task if @task
+
+extends 'foreman_tasks/api/tasks/show'
+
+attributes :parent_task_id, :start_at, :start_before
```
adamruzicka: This needs to contain `:external_id`.

jeremylenz:
> Apparently Firefox can't format dd-mm-yyyy date like chrome, and that's the format that Ruby sends. Fixed this and the styling issues, thanks for the help @adamruzicka !
> I still didn't receive the console error (not in Firefox or in Chrome)

Hi @MariaAga 

I'm on Chrome (75.0 on Fedora 30) and I'm also getting the "Invalid date" issue.

I think the problem is in `webpack/ForemanTasks/Components/TaskDetails/Components/TaskHelper.js`

In the `formatDate` function when you do

```javascript
return new Date(date.replace(/-/g, '/'));
```

It's replacing the timezone offset with a `/`, which makes it an invalid date.  Like this:

date screenshot

app/views/foreman_tasks/api/tasks/details.json.rabl
MariaAga: Fixed, thanks!

MariaAga:
@jeremylenz thanks for noticing! I changed the replace function to only replace twice now (and not all the occurrences)

MariaAga:
@adamruzicka fixed the odd issue. The problem was that the interval for re fetching the data was set twice but stopped once

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsActions.js
```diff
@@ -0,0 +1,109 @@
...
+const getTasksDetails = async (
+  id,
+  dispatch,
+  timeoutId,
+  refetchTaskDetailsAction
+) => {
+  try {
+    const { data } = await API.get(`/foreman_tasks/api/tasks/${id}/details`);
+    dispatch(requestSuccess(data));
+    if (data.state !== 'stopped') {
```
adamruzicka: Currently clicking the "start auto reloading" when the task is stopped does only one fetch and never really enables the polling. If the user can't really enable the polling for a stopped task, shouldn't the button be disabled?

MariaAga: This was done based on your previous comment:
> When you're looking at a stopped task and turn on auto reload, it keeps refreshing. Before this PR it would refresh once, figure out the task is already stopped and won't change in the future so it stopped reloading.

Did you mean something else?

adamruzicka: Sorry about that, I should have been more clear about it, the old behavior was a bit different.

The old way:
1) User clicks the button
2) Polling is enabled, button text changes
3) A single fetch is done
4) Frontend code determines the task cannot change in the future
5) Polling gets disabled, button text changes back

The new way is
1) User clicks the button
2) The frontend code determines the task cannot change in the future
3) A single fetch is done

MariaAga: Thanks for clearing it up! fixed, after the request is successful if the state is still stopped, then polling is stopped. Is this the right behavior?

adamruzicka: Almost, currently:
1) User clicks the button
2) Polling is enabled, button text changes, `FOREMAN_TASK_DETAILS_START_POLLING` is dispatched
3) A single fetch is done
4) Frontend code determines the task cannot change in the future
5) Polling gets disabled, text button changes back

However if I click the button again, I'd expect it to start over. Instead, `FOREMAN_TASK_DETAILS_STOP_POLLING` is dispatched and nothing happens.

MariaAga: Should work as expected now, thanks!

MariaAga:
failures on travis are unrelated to this PR, and fixed rubocops issues

adamruzicka:
@MariaAga thank you and sorry for the delays.

@sharvit, @jeremylenz thank you for testing this

PR #430, opened 2019-07-04
Fixes #27234 - fix minor issues with DOM tree

mirzal:
* Use `<div>` instead of `<p>`, so `<pre>` can be placed properly
* Terminate `&nbsp` (semicolon was missing)


adamruzicka:
Thank you @mirzal ! And congratulations to first PR to foreman-tasks :)

PR #435, opened 2019-07-12
Fixes #27292 - workaround for c3 destroy issue

iNecas:
There is a known issue in react-c3js that can lead to the charts to
disappear due to some race condition during component unmounting/destroy
phase https://github.com/bcbcarl/react-c3js/issues/22, with this error procuder
in the console log:

```
c3.js:9223 Uncaught TypeError: Cannot read property 'data_types' of null
    at ChartInternal.c3_chart_internal_fn.isPieType (c3.js:9223)
    at ChartInternal.c3_chart_internal_fn.isArcType (c3.js:9234)
    at ChartInternal.c3_chart_internal_fn.getArc (c3.js:4473)
    at c3.js:4816
    at SVGPathElement.<anonymous> (d3.js:8775)
    at Object.tick [as c] (d3.js:8956)
    at d3_timer_mark (d3.js:2166)
    at d3_timer_step (d3.js:2147)
```

Delaying the actual destroy a bit seems to help with workarounding the
issue.

### DIFF
--- a/webpack/ForemanTasks/Components/Chart/Chart.js
+++ b/webpack/ForemanTasks/Components/Chart/Chart.js
@@ -42,6 +42,8 @@ class C3Chart extends React.Component {
       // phase while unmounting/destroying - destroying right away leads
       // to issue described in https://github.com/bcbcarl/react-c3js/issues/22.
       // Delaying the destroy a bit seems to resolve the issue.
+      // The chart API methods are already bind explicitly, therefore we don't need
+      // any special handling when passing the function.
       setTimeout(this.chart.destroy, 1000);
       this.chart = null;
     } catch (err) {
--- a/webpack/ForemanTasks/Components/Chart/Chart.js
+++ b/webpack/ForemanTasks/Components/Chart/Chart.js
@@ -38,7 +38,12 @@ class C3Chart extends React.Component {
 
   destroyChart() {
     try {
-      this.chart = this.chart.destroy();
+      // A workaround for a case, where the chart might be still in transition
+      // phase while unmounting/destroying - destroying right away leads
+      // to issue described in https://github.com/bcbcarl/react-c3js/issues/22.
+      // Delaying the destroy a bit seems to resolve the issue.
+      setTimeout(this.chart.destroy, 1000);
+      this.chart = null;
     } catch (err) {
       throw new Error('Internal C3 error', err);
     }


iNecas:
@sharvit it took me a while to find what was wrong. I can't think if any suitable workaround as this one, but if you had some thoughts, please feel free to experiment with this.

adamruzicka:
Fixes the issue for me

webpack/ForemanTasks/Components/Chart/Chart.js
```diff
@@ -38,7 +38,12 @@ class C3Chart extends React.Component {
 
   destroyChart() {
     try {
-      this.chart = this.chart.destroy();
+      // A workaround for a case, where the chart might be still in transition
+      // phase while unmounting/destroying - destroying right away leads
+      // to issue described in https://github.com/bcbcarl/react-c3js/issues/22.
+      // Delaying the destroy a bit seems to resolve the issue.
+      setTimeout(this.chart.destroy, 1000);
+      this.chart = null;
```
sharvit: When you pass functions this way in javascript, you loose the `this` reference. So if the `destroy` function would use the `this` keywork it will receive something different than the chart.

Therefore I want to suggest using:
```js
setTimeout(() => {
  this.chart.destroy();
  this.chart = null;
}, 1000);
```

iNecas: Notice that we're actually not adding the `this.chart = null` within timeout, which is on purpose (so that we make sure this component is not touching it after that point). I don't think we care about `this` binding within the timeout

sharvit: I don't think `this.chart = null` will actually mean something because the component itself is already destroyed.

The `this` binding work because of the implementation of `chart.destroy` don't use `this`. Once they will this code will break.

iNecas: I see. Actually, the `destroy` function is using `this` already and it works thanks to the fact that `c3` is binding the API functions explicitly https://github.com/c3js/c3/blob/062fc2a93915b94b36daa1e168334109f5edcf20/src/chart.js#L17-L24

On the other hand, I don't know how the component behaves after it's unmounted (and would be referenced later in after timeout). Therefore, I would probably just keep this is it is.

iNecas: I've added an explicit comment though the mention that it's ok to do it this way in this case

iNecas:
Test failures unrelated

adamruzicka:
Thanks @iNecas , @sharvit !

PR #439, opened 2019-07-22
Fixes #27178 - Fix csv export for sub tasks 

adamruzicka:


app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -12,26 +12,15 @@ def show
...
 
     def summary
       render json: Task::Summarizer.new(resource_base, params[:recent_timeframe].to_i).summary
     end
 
     def sub_tasks
-      task   = resource_base.find(params[:id])
-      @tasks = filter(task.sub_tasks)
-      render :index
+      process_index resource_base.find(params[:id]).sub_tasks
```
ares: don't we lose taxonomies filtering, pagination (may be wanted) and searching on sub_tasks with this?

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -103,6 +92,19 @@ def resource_class
 
     private
 
+    def process_index(scope)
```
ares: slightly better name would be `process_list` or `response_with_tasks`, since sub_tasks is not index

app/controllers/foreman_tasks/tasks_controller.rb
adamruzicka: No, the filtering is done inside ~~`process_index`~~ `respond_with_tasks` and seems to work fine.

app/controllers/foreman_tasks/tasks_controller.rb
adamruzicka: Good point, fixed.

ares:
oh, this repo does not prevent merging multiple comments and have that a default policy, sorry for merging them separately

PR #440, opened 2019-07-23
Fixes #27386 - migrate to @theforeman/vendor pkg

sharvit:
Migrate to use the new [@theforeman/vendor](https://github.com/theforeman/foreman-js/tree/master/packages/vendor) package from `npm`.

1. Don't forget to `rm -rf node_modules package-lock.json` and `npm i` before you test it.
2. The storybook is currently broken in master so don't expect it to work in this PR.

### What to test?

1. `npm run lint` should work properly with no errors.
2. `npm test` should work properly with no errors.
3. running foreman dev-server should work
4. We shouldn't notice any UI changes. especially in the scss file that changed.

Thanks

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,11 +13,11 @@
     "transform-class-properties",
     "transform-object-rest-spread",
     "transform-object-assign",
-    "lodash",
     "syntax-dynamic-import"
   ],
   "env": {
     "test": {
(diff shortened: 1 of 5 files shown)


adamruzicka:
Thanks @sharvit !

PR #442, opened 2019-07-29
Fixes #26348 - Create tasks table in React

MariaAga:
Rewrite the task table in react
~blocked by https://github.com/theforeman/foreman/pull/6879~


webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+  getTableItems,
+  error,
+  status,
+  results,
+  loading,
+  query,
+  itemCount,
+  pagination,
+}) => {
+  useEffect(() => {
```
jeremylenz: Yay React hooks! :1st_place_medal:

adamruzicka:
Needs a rebase

MariaAga:
@adamruzicka This is ready to be reviewed, test failures are because of different testing available options in node 8 and node 10

MariaAga:
@sharvit - can you review this please?

waldenraines:
When I attempt to load the tasks dashboard page I see the tasks table loading indefinitely even though I get the `TASKS_SUCCESS` action.

MariaAga:
@waldenraines  works for me, did you use foreman tasks with the new foreman PR? https://github.com/theforeman/foreman/pull/6879

MariaAga:
Foreman function was changed so the action type names were changed in foreman and not in this PR, now fixed @sharvit

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
```diff
@@ -46,50 +46,23 @@ const queryFromUriQuery = uriQuery => {
...
-    },
-    error({ statusText }) {
-      $('#tasks-table').html(statusText);
-    },
-    complete(result) {
-      tfm.tools.hideSpinner();
-    },
-  });
+  if (query.mode === 'last') {
+    query.mode = 'recent';
```
sharvit: Changing the object the function received is a bad practice and can lead to issues if the consumer who passes the object to the function would want to continue and use the object. I don't see an issue in this specific case but consider changing to:

```js
const { state, result, mode, time } = query;
const uriQuery = {
...,
time_mode: mode === 'last' ? 'recent' : '<the-other-value>',
...
};
```

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+    getTableItems();
+  }, [query.state, query.result, query.mode, query.time]);
+  const onPageChange = paginationArgs => {
```
sharvit:
```suggestion
  const onPageChange = ({ page, perPage }) => {
```

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+      <MessageBox
+        key="tasks-table-error"
+        icontype="error-circle-o"
+        msg={__(`Could not receive data: ${error && error.message}`)}
+      />
+    );
+  }
+
+  if (loading) {
+    return <Spinner size="lg" loading />;
```
sharvit: Does showing a spinner gives a good user experience here?
I am wondering if showing blank would give better experience.
TBH spinners are nice but when the data loads fast it gives an annoying flickering effect.

Another option to use something like delayed-spinner that only shows a spinner when it takes more than 300ms to load the data.

Another option which is my favorite is using `react-loading-skeleton`, we added it to the vendor 0.2.0 so it's available already.
See: https://github.com/dvtng/react-loading-skeleton

If you decide to go with the `skelaton`, I would be OK merging it with the spinner and add this a `TableSkelaton` to patternfly-react and then use it when it's available.

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+      />
+    </React.Fragment>
+  );
+};
+
+TasksTable.propTypes = {
+  results: PropTypes.array.isRequired,
+  getTableItems: PropTypes.func.isRequired,
+  status: PropTypes.oneOf(Object.keys(STATUS)),
+  query: PropTypes.shape(),
```
sharvit: Need to add the shape's attributes

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+  error: PropTypes.object,
```
sharvit: Can be `PropTypes.instanceOf(Error)`

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+  loading: PropTypes.bool,
+  itemCount: PropTypes.number.isRequired,
+  pagination: PropTypes.shape({
+    page: PropTypes.number,
+    perPage: PropTypes.number,
+  }).isRequired,
```
sharvit: Why `pagination` is required? Shouldn't we use default values for consistency?

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,50 @@
...
+    case TASKS_REQUEST:
+      return state.set('loading', true);
+    case TASKS_FAILURE:
+      return state.set('loading', false);
+    case TASKS_SUCCESS:
+      return Immutable.merge(state, {
+        loading: false,
+        itemCount: payload.subtotal,
+        pagination: {
+          page: parseInt(payload.page, 10),
```
sharvit: Any reason not to use `Number(page)`?

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,50 @@
...
+          perPage: parseInt(payload.per_page, 10),
+        },
+      });
+    case FOREMAN_TASKS_DASHBOARD_UPDATE_QUERY:
+      if (payload.mode === 'last') {
```
sharvit: It doesn't feel right here, can't we store the data as is? or pass a payload with the right format?

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+    return <span>{__('No Tasks')}</span>;
+  }
+
+  const uriQuery = getURIQuery(window.location.href);
+  return (
+    <React.Fragment>
+      <Table
+        key="tasks-table"
+        columns={createTasksTableSchema(
+          getTableItems,
```
sharvit: Instead of passing the `getTableItems` action, what about passing an action that will just update the redux state with a new `query`. Then, your `useEffect` will call the `getTableItems`?

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,98 @@
...
+}) => {
+  useEffect(() => {
+    getTableItems();
+  }, [query.state, query.result, query.mode, query.time]);
+  const onPageChange = paginationArgs => {
+    updateURlQuery({
+      page: paginationArgs.page,
+      per_page: paginationArgs.perPage,
+    });
+    getTableItems();
```
sharvit: Instead of calling the `getTableItems` action, wouldn't it make more sense to just update the state with the new `page` and `perPage` so then your `useEffect` will do it's job?

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -0,0 +1,12 @@
+import URI from 'urijs';
+
+export const updateURlQuery = query => {
+  const uri = new URI(window.location.href);
+  uri.setSearch(query);
+  window.history.pushState({ path: uri.toString() }, '', uri.toString());
```
sharvit: Shouldn't the `history` object suppose to get imported from `foremanReact`?

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
```diff
@@ -0,0 +1,60 @@
...
+import { translate as __ } from 'foremanReact/common/I18n';
+import { updateURlQuery } from './TasksTableHelpers';
+
+const headFormat = [headerFormatterWithProps];
+const cellFormat = [cellFormatter];
+
+const sortControllerFactory = (apiCall, sortBy, sortOrder) => ({
+  apply: (by, order) => {
+    updateURlQuery({ sort_by: by, sort_order: order });
+    apiCall();
```
sharvit: IMO it should be called `action` because it represents the action you want to run.

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
```diff
@@ -0,0 +1,60 @@
...
+} from 'foremanReact/components/common/table';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { updateURlQuery } from './TasksTableHelpers';
+
+const headFormat = [headerFormatterWithProps];
+const cellFormat = [cellFormatter];
+
+const sortControllerFactory = (apiCall, sortBy, sortOrder) => ({
+  apply: (by, order) => {
+    updateURlQuery({ sort_by: by, sort_order: order });
```
sharvit: This code appearing again and again in many places, how can we reuse it?
What about moving it to inside the action?

webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
```diff
@@ -0,0 +1,44 @@
...
+export const selectLoading = state =>
+  selectTasksTableQuery(state).loading || false;
+
+export const selectPagitation = state =>
+  selectTasksTableQuery(state).pagination || {};
+
+export const selectItemCount = state =>
+  selectTasksTableQuery(state).itemCount || 0;
+
+const formatDate = date =>
```
sharvit: I don't think it belongs here. In the component level, you can import a date component from `foremanReact` who does the same.

webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
```diff
@@ -0,0 +1,44 @@
...
+    : __('N/A');
+
+const formatUsername = username => (username ? username.name : '');
+
+export const selectResults = state => {
+  const { results } = selectTasksTableContent(state);
+  if (!results) return [];
+  return results.map(result => ({
+    ...result,
+    username: formatUsername(result.username),
```
sharvit: Because you don't reuse `formatUsername` IMO it is redundant and you can just do:
```suggestion
    username: result.username ? result.username.name : '',
```

webpack/ForemanTasks/Components/TasksTable/index.js
```diff
@@ -0,0 +1,30 @@
...
+  selectResults,
+  selectTasksTable,
+  selectQuery,
+  selectPagitation,
+  selectItemCount,
+  selectLoading,
+} from './TasksTableSelectors';
+
+const mapStateToProps = state => ({
+  ...selectTasksTable(state),
```
sharvit: IMO it is risky to use `...` from a selector, consider moving the fields here and create a selector for each.

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
MariaAga: This logic is removed in the [next PR](https://github.com/theforeman/foreman-tasks/pull/445/commits/b7b9e2465e3f9d630fdd9094ea2e96cab56dfcc6) and instead the spinner is shown in the buttons row, and the table stays with the old data until its loaded, and when there is no data the table isn't shown, can you take a look?

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
MariaAga: which code? I didn't find anything reappearing

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: It doesn't change the browsers URL(only changes the path in the the `ReactApp`)

MariaAga: The history in foremanReact only changed the path in the main Router component and not in the browser it self. In the next next PR the history object is from the router

MariaAga:
@waldenraines Is the issue is not enough space between the end of the table and the end of the page?
@sharvit Thanks for the review, solved most of the issues.
Please note that right now the Export Button does not update its url after the table changes

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
MariaAga: Why? getTableItems updates the state with the new query and results, and useeffect will have to be changed to listen on these changes

jeremylenz:
@MariaAga I'm not seeing any issues with the export button.  Whenever I do a new search and click Search, the URL of the export button changes to match the current search query.  (If I type a new search query but don't click Search, the URL doesn't change, but I would expect that.)  If you're still having issues, can you give any more specific info about the expected vs. actual behavior?

waldenraines:
> @waldenraines Is the issue is not enough space between the end of the table and the end of the page?

No, the issue is the odd spacing of the number per page selector:

This page:

gscreenshot_2019-08-26-165017


Other pages:

gscreenshot_2019-08-30-132301

waldenraines:
> No, the issue is the odd spacing of the number per page selector:

@rohoover, which of these seems correct to you?

MariaAga:
@waldenraines I see the per page in the following pages: hosts, domains,organizations, users
like this:
image

MariaAga:
@jeremylenz Thanks for checking, the issue is when you filter via the dashboard.
the search causes the page to reload while the dashboard doesn't

MariaAga:
@adamruzicka Yes, this is the foreman standard date component

jeremylenz:
@MariaAga Thanks for clarifying!  I was able to see the issue now regarding the CSV Export button.

1. It looks like that button is rendered with ERB in `(foreman_tasks repo) app/views/foreman_tasks/tasks/index.html.erb`:

```ruby
<% title_actions csv_link, help_button %>
```
2. which calls Foreman's `title_actions` method:
```ruby
  def title_actions(*elements)
    content_for(:title_actions) { elements.join(" ").html_safe }
  end
```

3. which uses the Rails [`content_for` helper](https://apidock.com/rails/ActionView/Helpers/CaptureHelper/content_for),
4. and calls Foreman's `csv_link` method (`(foreman repo) app/helpers/application_helper.rb`):

```ruby
def csv_link(permitted: [])
    link_to(_('Export'), current_url_params(:permitted => permitted).merge(:format => :csv),
      {:title => _('Export to CSV'), :class => 'btn btn-default', 'data-no-turbolink' => true})
end
```

5. which actually generates the HTML for the `<a>` tag.

Whew.  Sometimes Rails magic is also Rails opacity.

Anyway..

If you don't want to go through the effort of rendering that Export button in React, maybe it would work to add an `id` to the button:
`:class => 'btn btn-default', :id => 'export-csv-button'`

Since the `csv_link` method is located in foreman and not foreman_tasks, it's probably better to create a specific method for foreman_tasks and and change your `app/views/foreman_tasks/tasks/index.html.erb` to use your new method instead of `csv_link`.

Once the id is on the button, you should be able to find it using old-fashioned `document.querySelector` or (uggh) jQuery, and update the `href` with the new URL query.  I think maybe a good place to do it would be your `useEffect` in `TasksTable`, since that component knows when the query changes.

MariaAga:
@jeremylenz Thanks for the great explanation about the solution to this issue!
this issue is not the next PR (https://github.com/theforeman/foreman-tasks/pull/445 - switches the export from ruby to React) so I'm leaving this PR for review purposes only :)

Rohoover:
@waldenraines 

image

That is the correct one.

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
sharvit: Im not sure, leave it

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,47 @@
...
+  switch (type) {
+    case TASKS_REQUEST:
+      return state.set('loading', true);
+    case TASKS_FAILURE:
+      return state.set('loading', false);
+    case TASKS_SUCCESS:
+      return Immutable.merge(state, {
+        loading: false,
+        itemCount: payload.subtotal,
+        pagination: {
```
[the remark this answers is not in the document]
MariaAga: Will add to foreman

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,47 @@
...
+          page: Number(payload.page),
+          perPage: Number(payload.per_page),
+        },
+      });
+    case FOREMAN_TASKS_DASHBOARD_UPDATE_QUERY:
+      return state.set('query', payload);
```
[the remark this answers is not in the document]
MariaAga: You're right, changed

webpack/ForemanTasks/Components/TasksTable/__tests__/TasksTableReducer.test.js
```diff
@@ -0,0 +1,35 @@
+import { testReducerSnapshotWithFixtures } from 'react-redux-test-utils';
+import {
+  TASKS_SUCCESS,
+  TASKS_REQUEST,
+  TASKS_FAILURE,
+} from '../TasksTableConstants';
+import reducer from '../TasksTableReducer';
+
+const fixtures = {
+  'should return the initial state': {},
```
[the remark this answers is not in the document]
MariaAga: removed it from this const from the reduced now

webpack/ForemanTasks/Components/TasksTable/index.js
```diff
@@ -0,0 +1,31 @@
...
+  selectTasksTable,
+  selectQuery,
+  selectPagitation,
+  selectItemCount,
+  selectLoading,
+} from './TasksTableSelectors';
+
+const mapStateToProps = state => ({
+  status: selectTasksTable(state).status,
+  error: selectTasksTable(state).error,
```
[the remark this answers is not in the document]
MariaAga: Added

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,47 @@
...
+  TASKS_REQUEST,
+  TASKS_SUCCESS,
+  TASKS_FAILURE,
+} from './TasksTableConstants';
+import { FOREMAN_TASKS_DASHBOARD_UPDATE_QUERY } from '../TasksDashboard/TasksDashboardConstants';
+
+const initialState = Immutable({
+  query: null,
+  loading: true,
+  itemCount: 0,
```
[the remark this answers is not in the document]
MariaAga: Changed to STATUS.PENDING

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
MariaAga: Added setQuery action

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
MariaAga: Added setPagination action

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
```diff
@@ -46,50 +46,20 @@ const queryFromUriQuery = uriQuery => {
...
-    },
-    error({ statusText }) {
-      $('#tasks-table').html(statusText);
-    },
-    complete(result) {
-      tfm.tools.hideSpinner();
-    },
-  });
+  const uriQuery = {
+    state: query.state,
```
laviro: Can you destruct the query object in the function signature?
```
export const resolveQuery = ({ state, result, mode, time }) => {
```

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
```diff
@@ -1,11 +1,11 @@
-import URI from 'urijs';
 import { getURIQuery } from 'foremanReact/common/helpers';
 
 import {
   TASKS_DASHBOARD_AVAILABLE_TIMES,
   TASKS_DASHBOARD_QUERY_KEYS_TEXT,
   TASKS_DASHBOARD_QUERY_VALUES_TEXT,
 } from './TasksDashboardConstants';
+import { updateURlQuery } from '../TasksTable/TasksTableHelpers';
```
laviro: Can you use the foreman helper for URI query?

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,116 @@
...
+    pagination.perPage,
+    sort.by,
+    sort.order,
+  ]);
+  const onPageChange = ({ page, perPage }) => {
+    updateURlQuery({
+      page,
+      per_page: perPage,
+    });
+    setPagination({ page, perPage });
```
laviro: should we use `async & await` here?

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  if (status === STATUS.ERROR) {
+    return (
+      <MessageBox
+        key="tasks-table-error"
+        icontype="error-circle-o"
+        msg={__(`Could not receive data: ${error && error.message}`)}
```
laviro: I didn't see this props are actually used in the MessageBox component,
which currently looks like: `export const MessageBox = () => <div className="message-box-root" />;`

the check for `error` and `error.message` could be moved to the upper `if` statement.

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
laviro: Thinking UX, I would suggest using an image of blurred table instead of a spinner
so the table won't jump and switch between the spinner to table.
I would also add debounce so the transitions wouldn't be too fast,
there are times when the table appears and then the spinner shows up again, but this can be in the next PR.

webpack/ForemanTasks/Components/TasksTable/TasksTable.scss
```diff
@@ -0,0 +1,6 @@
+.tasks-pagination {
+  margin-top: -6px;
+}
```
laviro: can you add an empty line here?
`[stylelint] Expected empty line before rule (rule-empty-line-before) [rule-empty-line-before]`

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,53 @@
...
+  query: null,
+  itemCount: 0,
+  pagination: {
+    page: 1,
+    perPage: 20,
+  },
+});
+
+export const TasksTableQueryReducer = (state = initialState, action) => {
+  const { type, payload } = action;
```
laviro: can you destruct `payload` into `page, per_page, by, order`?

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
laviro: oops, I saw now that the `MessageBox` is being mocked :rofl:

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
MariaAga: why?

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
MariaAga: Nope, this is not in foreman

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
MariaAga: I will have to split it to `{subtotal, page, per_page, by, order, perPage } ` and except for `page` none are reused, are you sure it's a good idea? :)

webpack/ForemanTasks/Components/TasksTable/TasksTable.scss
MariaAga: Added thanks

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
MariaAga: destructed thanks

MariaAga:
@laviro Thanks for the review.
The search button will do a full page reload in this PR so there is no need to keep the search query in the redux store for now

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
jeremylenz: @MariaAga why not just `const { page, per_page, by, order } = payload;` ? You don't have to extract every value if you don't need them..

MariaAga: I meant that each one of the values `{subtotal, page, per_page, by, order, perPage } ` is used once in the reducer (in different cases), so I'm not sure that destructing payload will be more readable. Also for each case only 1 or 2 values are used so I think its less readable to destruct all the 6 values

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
laviro: What I meant is that in Foreman we have the [urlHelpers](https://github.com/theforeman/foreman/blob/9394ecb277798c488312f46866ab5d2a319506c6/webpack/assets/javascripts/react_app/common/urlHelpers.js) file,
and since this code is being repeated all over Foreman & plugins, you can use the helpers or contribute your own because this is a very trivial usage.

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
laviro: I thought you were updating the url and only after that the table will get updated with pagination,
same as with the query, but now I see that it is only the action that updates it in this case, thanks

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
laviro: This is only cosmetic, I prefer not calling the `payload` object each time,
it is not a blocker..

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -0,0 +1,116 @@
...
+    query.state,
+    query.result,
+    query.mode,
+    query.time,
+    pagination.page,
+    pagination.perPage,
+    sort.by,
+    sort.order,
+  ]);
+  const onPageChange = ({ page, perPage }) => {
```
laviro: Can you move it to the actions file?

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,53 @@
...
+  switch (type) {
+    case TASKS_SUCCESS:
+      return Immutable.merge(state, {
+        itemCount: payload.subtotal,
+        pagination: {
+          page: Number(payload.page),
+          perPage: Number(payload.per_page),
```
laviro: Can you refactor it into a number in the action that fires this instead?

MariaAga: Nope, this is how we get it from the api and the action is in foreman (not foreman tasks)

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
MariaAga: I will contribute it to foreman and replace it in a separate PR

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
MariaAga: I tried it but it causes errors as payload can be undefined and you can't put `const {...} = ...` inside switch case

MariaAga: Done, thanks!

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
MariaAga: Moved, thanks

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
laviro: Debugging it both in Foreman and here, I see it is returning a number.

laviro: but when clicking on pagination I see it changes into a string.. weird
anyway it is non blocking for this PR.

webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
```diff
@@ -0,0 +1,31 @@
...
+    getURIQuery(window.location.href),
+    getApiPathname()
+  );
+
+export const setSort = sorting => ({
+  type: SET_SORT_TASKS_TABLE,
+  payload: sorting,
+});
+
+export const onPageChange = ({ page, perPage }) => {
```
sharvit: IMO it should be called `changeTablePage`

webpack/ForemanTasks/Components/TasksTable/TasksTableConstants.js
```diff
@@ -0,0 +1,8 @@
+import { getControllerSearchProps } from 'foremanReact/constants';
+
+export const TASKS_CONTROLLER = 'tasks';
+export const TASKS_SUCCESS = 'TASKS_SUCCESS';
```
sharvit: The convention is to have long const names, e.g. `TASKS_TABLE_CONTROLLER`, `TASKS_TABLE_GET_ITEMS_SUCCESS`.

sharvit: `TASKS_TABLE_GET_ITEMS_SUCCESS` can be generated using:
https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/table/actionsHelpers/actionTypeCreator.js#L1

webpack/ForemanTasks/Components/TasksTable/TasksTableConstants.js
```diff
@@ -0,0 +1,8 @@
...
+export const SET_SORT_TASKS_TABLE = 'SET_SORT_TASKS_TABLE';
+export const SET_PAGINATION_TASKS_TABLE = 'SET_PAGINATION_TASKS_TABLE';
```
sharvit: The convention is to first have the base name, e.g. `TASKS_TABLE_SET_SORT`, `TASKS_TABLE_SET_PAGINATION`

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
```diff
@@ -0,0 +1,63 @@
...
+import { createTableReducer } from 'foremanReact/components/common/table';
+import {
+  TASKS_CONTROLLER,
+  TASKS_SUCCESS,
+  SET_SORT_TASKS_TABLE,
+  SET_PAGINATION_TASKS_TABLE,
+} from './TasksTableConstants';
+
+const initialState = Immutable({
+  query: null,
```
sharvit: Shouldn't it suppose to be removed once it using the base reducer?

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
```diff
@@ -0,0 +1,68 @@
...
+import { translate as __ } from 'foremanReact/common/I18n';
+import LongDateTime from 'foremanReact/components/common/dates/LongDateTime';
+import { updateURlQuery } from './TasksTableHelpers';
+
+const headFormat = [headerFormatterWithProps];
+const cellFormat = [cellFormatter];
+
+const sortControllerFactory = (setSort, sortBy, sortOrder) => ({
+  apply: (by, order) => {
+    updateURlQuery({ sort_by: by, sort_order: order });
```
sharvit: Can it move to the action?

webpack/ForemanTasks/Components/TasksTable/TasksTableConstants.js
MariaAga: Changed to TASKS_TABLE_ID = TASKS_TABLE for clarity

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
MariaAga: Moved, thanks

webpack/ForemanTasks/Components/TasksTable/TasksTableReducer.js
MariaAga: Not sure what you mean, but I see now that ` query: null,` has no meaning

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
```diff
@@ -0,0 +1,66 @@
...
+ * Generate a table schema to the Hardware Tasks page.
+ * @param  {Function} setSort a Redux async action that sets new sort values
+ * @param  {String}   by      by which column the table is sorted.
+ *                            If none then set it to undefined/null.
+ * @param  {String}   order   in what order to sort a column. If none then set it to undefined/null.
+ *                            Otherwise, 'ASC' for ascending and 'DESC' for descending
+ * @return {Array}
+ */
+const createTasksTableSchema = (setSort, by, order) => {
+  const sortController = sortControllerFactory(setSort, by, order);
```
sharvit: Does the `sortControllerFactory` actually need now? e.g.

```js
const sortController = {
  property: sortBy,
  order: sortOrder,
  apply: (by, order) => setSort({ by, order }),
};
```

MariaAga: Nope, removed, thanks

sharvit:
@adamruzicka can it be merged?

adamruzicka:
https://imgur.com/a/EtS6qjX

I noticed several things. The dates now have a popup which shows the date as shown on the page and after a while second popup appears, showing the date in relative words.

Next one is that the tiny cross to clear out the search bar overlaps the search button.

Also if not probably really caused by this PR but worth noting. If I click around the index for a bit (change per-page, change page, change filters), the backend freezes after several requests. I've seen this happen before, but haven't found the cause. With this applied I managed to run into much more often.

MariaAga:
@adamruzicka about the search X, this happens because of the marked for translation.
This also happens to me in master as the search bar didn't change in this PR
screenshot-192 168 122 137-2019 09 26-13-42-18

adamruzicka:
Fair enough, but I'd still like to see the popups get sorted out somehow.

MariaAga:
For now we wont use the foreman component until the '5 days ago' tooltip will be removed from there (by me). 
@adamruzicka

adamruzicka:
Needs a rebase

MariaAga:
@sharvit / @laviro  I had to add `getURIPagination` function to this PR because there were double API calls without it as the first api call set the pagination and triggered the second call.
Rebased

webpack/ForemanTasks/Components/TasksTable/TaskTableFormmatter.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
sharvit: nitpick: the file name should be `TaskTableFormatters.js`

webpack/ForemanTasks/Components/TasksTable/TaskTableFormmatter.js
```diff
@@ -0,0 +1,28 @@
...
+import { FormattedDate } from 'react-intl';
+import { isoCompatibleDate } from 'foremanReact/common/helpers';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { cellFormatter } from 'foremanReact/components/common/table';
+
+export const dateCellFormmatter = value => {
```
sharvit: Would it make sense donating to core?

MariaAga: already in my tasks list :)

webpack/ForemanTasks/Components/TasksTable/TaskTableFormmatter.js
MariaAga: changed, thanks

sharvit:
Notice the linting error

MariaAga:
@adamruzicka Fixed the columns to work as expected, thanks!

adamruzicka:
[test foreman-tasks]

adamruzicka:
Thank you @MariaAga and thanks to @jeremylenz, @waldenraines and @sharvit for reviews!

PR #444, opened 2019-07-29
 Refs #26346 - Adds react export button to tasks table

MariaAga:
Based on https://github.com/theforeman/foreman-tasks/pull/442


MariaAga:
Moved to core

PR #445, opened 2019-07-30
Refs #26346 - Move tasks table to react router

MariaAga:
Blocked by:
https://github.com/theforeman/foreman-tasks/pull/442
~~https://github.com/theforeman/foreman/pull/6997~~
~~https://github.com/theforeman/foreman/pull/6944~~


app/views/foreman_tasks/layouts/react.html.erb
```diff
@@ -1,6 +1,9 @@
 <% content_for(:javascripts) do %>
   <%= webpacked_plugins_js_for :'foreman-tasks' %>
 <% end %>
+<% content_for(:stylesheets) do %>
+    <%= webpacked_plugins_css_for :'foreman-tasks' %>
```
adamruzicka: Indented too much

webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const selectPagitation = state =>
+  selectTasksTableQuery(state).pagination || {};
+
+export const selectItemCount = state =>
+  selectTasksTableQuery(state).itemCount || 0;
+
+const formatDate = date =>
+  date
+    ? new Date(date).toLocaleString('en-GB', {
```
adamruzicka: Hardcoded locale? Shouldn't it be retrieved from the environment or something?

MariaAga: I didnt want there to be a confusion with the search text as the date for 'en-US' would be MM/DD/YYYY in the table, but it can't be searched in this format

adamruzicka:
Needs a rebase

waldenraines:
@adamruzicka I think this is ready for a review now.

app/views/foreman_tasks/layouts/react.html.erb
```diff
@@ -9,4 +12,4 @@
   <div id="foremanTasksReactRoot"></div>
 <% end %>
 <%= render file: "layouts/base" %>
-<%= mount_react_component('ForemanTasks', '#foremanTasksReactRoot',{ :flatten_data => true }) %>
+<%= mount_react_component('ForemanTasks', '#foremanTasksReactRoot') %>
```
laviro: why is it still needed after adding react-router?

MariaAga: To load the router

laviro: Ok, pretty soon after https://github.com/theforeman/foreman/pull/6936 will get merge this won't be needed anymore.

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
```diff
@@ -36,24 +36,24 @@ const createTasksTableSchema = (setSort, by, order) => {
   return [
     column(
       'action',
-      __('Action'),
+      'Action',
```
sharvit: What is the reason behind removing the `__`?

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -10,3 +10,21 @@ export const getApiPathname = () => {
...
+  const uriQuery = {
+    search,
+    page: 1,
+    per_page: 20,
+  };
+  updateURlQuery(uriQuery);
+};
+
+export const getSubTaskID = () =>
+  window.location.pathname.split('/').slice(-2)[0];
```
sharvit: I think you should be able to receive this value from the react-router `location` prop.
See: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md

webpack/ForemanTasks/Routes/ForemanTasksRoutes.js
```diff
@@ -1,12 +1,17 @@
...
-    path: '/foreman_tasks/ex_tasks',
+    path: '/foreman_tasks/tasks',
     exact: true,
-    render: props => <IndexTasks {...props} />,
+    render: props => <TasksTableWrapper {...props} />,
+  },
+  subTasks: {
+    path: '/foreman_tasks/tasks/:id/sub_tasks',
+    exact: true,
+    render: props => <TasksTableWrapper isSubTask {...props} />,
```
sharvit: IMO `indexTasks` and `subTasks` should not use the `TasksTableWrapper` directly, instead the can have their own components. e.g. `<TasksIndexPage />` and `<SubTaskPage />`

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
MariaAga: The `column` translates it and I noticed it here :)
https://github.com/theforeman/foreman/blob/7f113cb4b781283ea7b50ee78d980414dd3bc741/webpack/assets/javascripts/react_app/components/common/table/schemaHelpers/column.js#L28

sharvit: Actually it's bad, the `column` should not do that because we have a script that searches for the `__` in order to extract all the strings needs to be translated and upload them to the translation service.

This way, when you have a variable instead of the actual string, the extraction script won't be able the extract the strings.

IMO, bring back the translations and send a PR to foreman that removes the translation from the column.

webpack/ForemanTasks/Routes/ForemanTasksRoutes.js
MariaAga: Why? They are the same and the only difference is the breadcrumbs

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: Changed, thanks

webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
MariaAga: Added back, will open a PR in foreman

webpack/ForemanTasks/Routes/ForemanTasksRoutes.js
sharvit: At the future we might want to make 2 types of changes:
1. Changes that affect both of the pages. Now can be achieved easily
2. Changes that affect only one of the pages. Will be tricky to do ATM.

Another reason is to keep a unified convention for routes. ATM each route have its own component.

MariaAga: If the subtasks component changes enough to be a different component then it should be a different component. But right now it has the same design and features as the index plus breadcrumbs. 
I don't think it would make sense to copy the whole TasksTableWrapper into 2 components which will have 2 of the same index files.
And if we just reuse TasksTableWrapper then changes that affect only one of the pages will be tricky to do anyway.

sharvit: I didn't mean removing the `TasksTableWrapper`, just creating 2 different components that uses the `TasksTableWrapper`. They should be very small and should just render the `TasksTableWrapper` with the right props.

I would consider renaming `TasksTableWrapper` to `TasksTablePage`.

MariaAga: I agree on the name.
If both the components use `TasksTableWrapper` then changing one of them will still involve changing `TasksTableWrapper` for both

webpack/ForemanTasks/Components/TasksTable/SubTasksPage.js
```diff
@@ -0,0 +1,18 @@
...
+import TasksTablePage from './';
+import { getSubTaskID } from './TasksTableHelpers';
+
+export const SubTasksPage = props => {
+  const getBreadcrumbs = actionName => ({
+    breadcrumbItems: [
+      { caption: __('Tasks'), url: `/foreman_tasks/tasks` },
+      {
+        caption: actionName,
+        url: `/foreman_tasks/tasks/${getSubTaskID(window.location)}`,
```
sharvit: Instead `window.location`, can't we use the one provided by `react-router`?

MariaAga: Chagned, thanks

webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
```diff
@@ -0,0 +1,63 @@
...
+import { getURIQuery } from 'foremanReact/common/helpers';
+import ExportButton from 'foremanReact/pages/common/PageLayout/components/ExportButton/ExportButton';
+import { STATUS } from 'foremanReact/constants';
+import TasksDashboard from '../TasksDashboard';
+import TasksTable from './TasksTable';
+import { resolveSearchQuery, addSearchToURL } from './TasksTableHelpers';
+import { TASKS_SEARCH_PROPS } from './TasksTableConstants';
+import './TasksTablePage.scss';
+
+const TasksTablePage = ({ ...props }) => {
```
sharvit: Does it make sense to do destruct the `getTableItems` from the `props` so the `TasksTable` won't receive it?
```js
const TasksTablePage = ({ getBreadcrumbs, ...props }) => { ... };
```

webpack/ForemanTasks/Routes/ForemanTasksRoutes.js
```diff
@@ -1,12 +1,19 @@
 import React from 'react';
-import IndexTasks from './IndexTasks';
+// import TasksTablePage from '../Components/TasksTable';
```
sharvit: Leftover?

webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
MariaAga: Yup, changed!

webpack/ForemanTasks/Routes/ForemanTasksRoutes.js
MariaAga: Removed, thanks for noticing

sharvit:
Needs a rebase

sharvit:
Please notice the testing error

MariaAga:
@adamruzicka  looks like this PR(https://github.com/theforeman/foreman/pull/6936) will be in soon and will change the React Router behavior in Foreman plugins, so I think we should wait for it

adamruzicka:
Agreed, I'll keep an eye out for it

PR #446, opened 2019-07-31
 Refs #26346 - tasks table uses router history

MariaAga:
Instead of using `window.history`


adamruzicka:
Needs a rebase

sharvit:
Looking at the commits, looks like need a rebase?

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -50,11 +40,31 @@ const TasksTable = ({
...
+      },
+      history
+    );
+  };
+
+  const setSortHistory = (by, order) => {
+    updateURlQuery({ sort_by: by, sort_order: order }, history);
+  };
+
+  const uriQuery = getURIQuery(window.location.href);
```
sharvit: Can't it use the `history` object as well?

MariaAga: You're right! changed here and in other places that I missed as well

adamruzicka:
Thank you @MariaAga for the patch, thank you @sharvit and @waldenraines for reviews.

PR #448, opened 2019-08-11
Refs #27081 - Gray-out unfocused cards

MariaAga:
inactive cards will be grayed-out as a whole and not only part of the card is grayed out
image


adamruzicka:
Isn't it too much? I can hardly see there are some actually clickable elements in the greyed-out cards. 

CC @rohoover

MariaAga:
Another suggestion: (opacity here is 0.6 instead of 0.3)
image

adamruzicka:
Thank you @MariaAga

PR #450, opened 2019-08-13
Fixes #27611 - ignore permission scope for bulk actions

jlsherrill:


jlsherrill:
my thinking is that whatever is being passed in should have done a permissions check (or decided to bypass it), so foreman-tasks shouldn't enforce that at run time

adamruzicka:
Thank you @jlsherrill

PR #453, opened 2019-08-22
Fixes #27683 - tasks details should use the core date components

MariaAga:
https://github.com/theforeman/foreman/tree/develop/webpack/assets/javascripts/react_app/components/common/dates

instead of rewriting the components, foreman tasks now reuses foreman components


MariaAga:
@adamruzicka Can this be merged?

adamruzicka:
Thanks @MariaAga for the fix and @jeremylenz and @waldenraines for the reviews

PR #457, opened 2019-09-05
 Fixes #27792 - Make task dashboard honor set taxonomies 

adamruzicka:


xprazak2:
The taxonomy search gets applied to scope and fixes the issue for running tasks, but the task count for stopped tasks still does not add up: 33 + 8 + 197 != 233 of total items in the table

Before:
paused-before
stopped-before

After:

paused-after
stopped-after

ares:
Test failures related. Do I get it correctly, no further changes on frontend needed?

adamruzicka:
No further changes in the frontend are needed. Test failures indeed look related, looking into it

ares:
tests are still red, not sure if it's caused by this PR though

adamruzicka:
It shouldn't be related, also I'm failing to reproduce locally.

adamruzicka:
Task failures were indeed unrelated.

PR #458, opened 2019-09-09
Fixes #27803 - Add missing c3 dependency

adamruzicka:


ares:
Thanks @adamruzicka, merging!

PR #461, opened 2019-09-15
Fixes #27801 - Add cancel button to the tasks table

MariaAga:
Adds a button that will cancel a task in the table (if the task can be canceled) and display a toast notification based on the answer
Based on: https://github.com/theforeman/foreman-tasks/pull/452

PR #462, opened 2019-09-15
Fixes #27802 - Add resume button to the tasks table 

MariaAga:
Changes the cancel button to resume if the task can be resumed and display a toast notification based on the answer
Based on: https://github.com/theforeman/foreman-tasks/pull/461


adamruzicka:
Needs rebase

webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
```diff
@@ -40,3 +40,28 @@ export const cancelTask = (id, name) => async dispatch => {
     );
   }
 };
+
+export const resumeTaskAction = (id, name, url) => async dispatch => {
```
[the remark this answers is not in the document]
sharvit: ?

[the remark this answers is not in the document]
MariaAga: just got to it this morning :)
Changed resumeTask, cancelTask

webpack/ForemanTasks/Components/common/ActionButton/ActionButton.js
```diff
@@ -0,0 +1,64 @@
...
+      cancel={taskActions.cancel}
+    />
+  );
+};
+
+ResumeButton.propTypes = {
+  id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  resume: PropTypes.func.isRequired,
+  resumable: PropTypes.bool.isRequired,
```
sharvit: Shouldn't we use default values for the buttons?
```js
ResumeButton.defaultProps = {
  resume: noop,
  resumable: false,
};
```

sharvit: In the context of `ResumeButton` and `CancelButton`, does it make more sense to call it `onClick` and `disabled`?

webpack/ForemanTasks/Components/common/ActionButton/ActionButton.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { Button } from 'patternfly-react';
+import PropTypes from 'prop-types';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+export const CancelButton = ({ id, name, cancellable, cancel }) => (
```
sharvit: IMO and based on the foreman convention `CancelButton` and `ResumeButton` should live in their own files.

webpack/ForemanTasks/Components/common/ActionButton/ActionButton.js
MariaAga: Why?

sharvit: So it becomes more generic

MariaAga: Why would anyone need a button that just says `resume` with no functionality?

sharvit: I don't believe someone will use it without functionality but it will reduce the gap needed to render the component and experience with it.

We do have a general rule that says if I can provide a default value without breaking the component, it is better to do so.

MariaAga: why a "button without a functionality" is not a broken component?

sharvit: Because I can still render it and experience with it

MariaAga:
added the title "Action" to the table

adamruzicka:
Thank you @MariaAga and @sharvit !

PR #463, opened 2019-09-15
Fixes #27855 - Add bulk cancel ui to tasks

MariaAga:
Select the tasks you want to cancel by using select boxes in the table and click 'cancel selected'.
If one of the selected tasks is not cancelable that task will not be attempted to be cancel.
Select all button will select only the tasks in the page
blocked by:https://github.com/theforeman/foreman-tasks/pull/461


adamruzicka:
@laviro but cancel describes precisely what it does, it does not delete the tasks. It just tells a running task to stop whatever it is doing and move to the stopped state. The task will still be there, just in a different state

laviro:
@adamruzicka Thanks for clarifying this,
so maybe "stop & resume" instead of "cancel & resume",
cc a native speaker @jeremylenz ^^

webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
```diff
@@ -0,0 +1,111 @@
...
+      type: 'info',
+      message: __('Trying to cancel the task'),
+    })
+  );
+  try {
+    await API.post(`/foreman_tasks/tasks/${id}/cancel`);
+    dispatch(
+      addToast({
+        type: 'success',
+        message: __('Task cancelled'),
```
jeremylenz: It would be really helpful here to know which task was successfully canceled (especially since this is for bulk and not just a single task.)  I know the API only provides `{statusText: "OK"}` but I think this would be worth making the change, either by using our available data on the frontend or making the API give us the task name in the response.

jeremylenz:
@laviro @adamruzicka I think "cancel selected" makes sense for the button verbiage. 
Maybe we can clarify in the confirmation dialog that we're adding.  Something like:
```
This will stop 3 tasks, putting them in the "canceled" state.  Are you sure?

Cancel 3 tasks   |   Never mind
```

jeremylenz:
^^ cc @Rohoover
cancel_tasks

MariaAga:
Adding the confirm dialog, and task ID to the toast.
`Cancel Selected` button can be `Cancel Selected Tasks` if its clearer  
image
image

webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
MariaAga: Added id to the toasts, thanks!

jeremylenz:
@MariaAga  "Cancel Selected Tasks" reads better to me.  If it makes the button too big, maybe just change the modal title instead?

For the toast, I was thinking the task name would be much better.  People don't want to read the ID.  So, something like 

`Task "Add subscription" successfully cancelled.`

MariaAga:
@jeremylenz Changed the title, and added a name to the toast. will add the same changes to the rest of the PRs about the cancel\resume

Rohoover:
Usually we don't indicate that the button is for the selected items, we just say "Cancel". Much like how we use "Delete". 

Is the button disabled until something is selected?

jeremylenz:
@MariaAga Per Roxanne's comment above, we should change the "Cancel Selected" button in the table back to just "Cancel."  I'm fine with this since it's clear from context what will happen (if you have multiple selected, it makes sense that that Cancel button will apply to all of them.)

MariaAga:
@Rohoover thanks for clearing this up.

> Is the button disabled until something is selected?

yes

webpack/ForemanTasks/Components/TasksTable/Components/ActionConfirmation.js
```diff
@@ -0,0 +1,49 @@
...
+}) => (
+  <Modal show={showModal} onHide={closeModal}>
+    <Modal.Header>
+      <Button
+        className="close"
+        onClick={closeModal}
+        aria-hidden="true"
+        aria-label="Close"
+      >
+        &times;
```
amirfefer: translation needed?

webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
```diff
@@ -65,3 +72,48 @@ export const resumeTask = (id, name) => async dispatch => {
     );
   }
 };
+
+export const selectAllRows = results => ({
```
amirfefer: IMHO these actions should be part of the general Table actions in foreman core, so we can reuse bulk actions in other tables as well. can you add it to our technical debt please?

webpack/ForemanTasks/Components/TasksTable/Components/ActionConfirmation.js
MariaAga: for the × symbol?

adamruzicka:
Needs rebase

MariaAga:
@Rohoover  I just noticed that there are 2 buttons next to each other-
Export - that will export all tasks
Cancel - that will cancel only the selected tasks
Are you sure the button shouldn't be "Cancel Selected? 
image

adamruzicka:
Needs a rebase

MariaAga:
rebased

adamruzicka:
[test foreman-tasks]

adamruzicka:
Thank you @MariaAga ! Also thanks to the reviewers

PR #464, opened 2019-09-18
Fixes #27880 - change page removes active filters

MariaAga:
currently on the tasks index, if you select for example stopped from the dashboard and then change a page (or anything pagination related), then the active filter (result = stopped) gets unselected.
This was caused by the JQuery that was added instead of the full page reload(which changed the URL)

### DIFF
--- a/webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
+++ b/webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
@@ -66,14 +66,29 @@ export const getQueryFromUrl = () => {
   return queryFromUriQuery(uriQuery);
 };
 
+export const updateURLQuery = ({ state, result, mode, time, uri }) => {
+  const query = {
+    state,
+    result,
+    time_mode: mode === 'last' ? 'recent' : mode,
+    time_horizon: time,
+    page: 1,
+  };
+
+  uri.setSearch(query);
+  window.history.pushState({ path: uri.toString() }, '', uri.toString());
+};
+
 export const resolveQuery = query => {
   const uriQuery = uriQueryFromQuery(query);
 
   const uri = new URI(window.location.href);
-  const { search } = uri.query(true);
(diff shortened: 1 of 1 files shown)


MariaAga:
@adamruzicka Fixed this issue as well, the JQuery didn't get the info about the per page from anywhere and now it does

jeremylenz:
Yes, confirming the per-page is working now.

jeremylenz:
@MariaAga I noticed that when you have enough tasks the tasks table goes to the very bottom of the screen and makes it hard to use the per-page controls.  Can you add some margin at the bottom of the table?
tasks

MariaAga:
@jeremylenz Its added in the react table PR and I don't want to add this to this PR as it's not related to its issue

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardHelper.js
```diff
@@ -66,15 +66,31 @@ export const getQueryFromUrl = () => {
...
 
   const uri = new URI(window.location.href);
-  const { search } = uri.query(true);
+  const { search, per_page: perPage } = uri.query(true);
 
-  const data = { search, ...uriQuery, page: 1 };
+  const data = { search, ...uriQuery, page: 1, per_page: perPage };
   const { $, tfm } = window;
 
+  updateURLQuery(query);
```
laviro: Why creating a separate function: `updateURLQuery` ?
instead of creating a new instance of `URI`, you can use the `uri` which was created here.

MariaAga: wanted to have 2 small functions instead of 1 big function, reused the uri, thanks!

MariaAga:
@adamruzicka  Can this be merged?

adamruzicka:
Thank you @MariaAga for the fix and @laviro, @sharvit and @jeremylenz for reviews!

PR #465, opened 2019-09-19
Fixes #27858 - Add bulk resume ui to tasks

MariaAga:
Select the tasks you want to resume/cancel by using select boxes in the table and choose the action with the 'select action' button drop down
If one of the selected tasks is not cancelable/resumable that task will not be attempted to be canceled/resumed.
Select all button will select only the tasks in the page

### DIFF
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
@@ -8,8 +8,10 @@ import {
   SELECT_ROWS,
   UNSELECT_ALL_ROWS,
   UNSELECT_ROWS,
-  TASKS_TABLE_SHOW_CANCEL_ALL_MODAL,
-  TASKS_TABLE_HIDE_CANCEL_ALL_MODAL,
+  TASKS_TABLE_SELECTED_MODAL,
+  CLOSED,
+  RESUME,
+  CANCEL,
 } from './TasksTableConstants';
 import { getApiPathname } from './TasksTableHelpers';
 import { fetchTasksSummary } from '../TasksDashboard/TasksDashboardActions';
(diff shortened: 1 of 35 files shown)


webpack/ForemanTasks/Components/TasksTable/TasksTableActions.js
```diff
@@ -93,18 +95,25 @@ export const unselectRow = id => ({
...
-    notAllCancelleble = true;
+    if (actionType === RESUME && task.isResumeble)
+      return dispatch(resumeTask(task.id));
+    else if (actionType === CANCEL && task.isCancelleble)
+      return dispatch(cancelTask(task.id));
+    notAllActionable = true;
     return null;
   });
-  if (notAllCancelleble)
+  if (notAllActionable)
```
amirfefer: tasks table reloads even though none of the tasks are resumable/cancelable and no data has been changed

adamruzicka:
Needs rebase

MariaAga:
@amirfefer Added the modal for resume action as well,
@adamruzicka The confirm text for resume says "This will resume 3 tasks, putting them in the running state.  Are you sure?" is this correct?

adamruzicka:
Yes, that sounds right

MariaAga:
@amirfefer I dont think it makes sense to make the button a drop down because a task can't be cancelable and resumable at the same time. (@adamruzicka  am correct?)

MariaAga:
Cancel confirmation will be added it a separate PR.
Title actions will be added in https://github.com/theforeman/foreman-tasks/pull/462

adamruzicka:
Task can be either resumable or cancelable, but not both at the same time. Needs a rebase

MariaAga:
rebased

MariaAga:
@adamruzicka 
1. Would that be from the server side?
2. What if you select 20 tasks but only 10 are canceled?

adamruzicka:
1) The server would receive a search query and remove everything that matches.
2) I'm afraid I don't have an answer to that yet, maybe we could bring the UXD team to this

adamruzicka:
Needs a rebase now that #463 was merged

adamruzicka:
@MariaAga could you please fix travis?

PR #466, opened 2019-09-23
Add a duration column to the tasks table

MariaAga:
The duration column replaces ended at column.
Duration text is in the same format as in task details.
For running tasks the duration is presented with a + (2 minutes +)
For Canceled tasks duration is presented as N\A with a tool tip that says that the task is canceled
For tasks without a start and end time, duration is presented as N\A with a tool tip about it
For sorting purposes N\A is the shortest time.

### DIFF
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
@@ -1,4 +1,7 @@
 import URI from 'urijs';
+import { translate as __, documentLocale } from 'foremanReact/common/I18n';
+import humanizeDuration from 'humanize-duration';
+import { isoCompatibleDate } from 'foremanReact/common/helpers';
 
 export const updateURlQuery = (query, history) => {
   const uri = new URI(history.location.pathname + history.location.search);
@@ -23,4 +26,39 @@ export const addSearchToURL = (path, query) => {
   const url = new URI(path);
   url.addSearch({ ...query, include_permissions: true });
   return url.toString();
+};
+
+export const getDuration = (start, finish) => {
+  if (!start && !finish)
+    return { text: __('N/A'), tooltip: __('No start or end dates') };
+
+  if (!start && finish) {
+    return { text: __('N/A'), tooltip: __('Task was canceled') };
+  }
+
+  const dateOptions = {
+    largest: 1,
+    language: documentLocale(),
+    fallbacks: ['en'],
+    round: true,
+  };
+
+  const startDate = new Date(isoCompatibleDate(start));
+
+  if (!finish) {
+    const finishDate = new Date();
+    const duration = finishDate - startDate;
+    return {
+      text: `${__('More than')} ${humanizeDuration(duration, dateOptions)}`,
+    };
+  }
+
+  const finishDate = new Date(isoCompatibleDate(finish));
+  const duration = finishDate - startDate;
+  return {
+    text:
+      duration > 0 && duration < 1000
+        ? __('Less than a second')
+        : humanizeDuration(duration, dateOptions),
+  };
 };
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -237,6 +237,7 @@ module ForemanTasks
       def ordering_scope(scope, ordering_params)
         sort_by = ordering_params[:sort_by] || 'started_at'
         sort_order = ordering_params[:sort_order] || 'DESC'
+        scope = scope.select("*, coalesce(ended_at, current_timestamp) - coalesce(coalesce(started_at, ended_at), current_timestamp) as duration")
         scope.order("#{sort_by} #{sort_order}")
       end
 
--- a/webpack/ForemanTasks/Components/TasksTable/TaskTableFormmatters.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TaskTableFormmatters.js
@@ -45,3 +45,9 @@ export const actionCellFormatter = taskActions => (
       name={action}
     />
   );
+
+export const durationCellFormmatter = value => (
+  <span className="param-value" title={value.tooltip}>
+    {value.text}
+  </span>
+);
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTableSchema.js
@@ -9,6 +9,7 @@ import {
   dateCellFormmatter,
   actionCellFormatter,
   actionNameCellFormatter,
+  durationCellFormmatter,
 } from './TaskTableFormmatters';
 
 const headFormat = [headerFormatterWithProps];
@@ -48,8 +49,8 @@ const createTasksTableSchema = (setSort, by, order, taskActions) => {
     sortableColumn('started_at', __('Started at'), 3, sortController, [
       dateCellFormmatter,
     ]),
-    sortableColumn('ended_at', __('Ended at'), 3, sortController, [
-      dateCellFormmatter,
+    sortableColumn('duration', __('Duration'), 3, sortController, [
+      durationCellFormmatter,
     ]),
     column(
       'available_actions',
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
@@ -1,5 +1,6 @@
 import { translate as __ } from 'foremanReact/common/I18n';
 import { selectForemanTasks } from '../../ForemanTasksSelectors';
+import { getDuration } from './TasksTableHelpers';
 
 export const selectTasksTable = state =>
   selectForemanTasks(state).tasksTable || {};
@@ -26,6 +27,7 @@ export const selectResults = state => {
     ...result,
     username: result.username || '',
     state: result.state + (result.frozen ? ` ${__('Disabled')}` : ''),
+    duration: getDuration(result.started_at, result.ended_at),
   }));
 };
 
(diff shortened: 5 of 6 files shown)


adamruzicka:
Needs a rebase

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -26,3 +30,40 @@ export const addSearchToURL = (path, query) => {
...
+
+  const dateOptions = {
+    largest: 1,
+    language: selectedLocale,
+    fallbacks: ['en'],
+  };
+
+  start = new Date(isoCompatibleDate(start)).getTime();
+  if (start && !finish) {
+    finish = new Date().getTime();
```
sharvit: Looks like `getTime` is redundant. I assume this will bring the same results:
```js
start = new Date(isoCompatibleDate(start));
finish = new Date();
duration = finish - start; // should contain the number of the difference
```

sharvit:
Thanks @MariaAga it looks good, waiting for the other PRs to get merged

laviro:
@MariaAga could you rebase ? thanks

MariaAga:
rebased

MariaAga:
The duration column value is now rounded

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -1,4 +1,8 @@
 import URI from 'urijs';
+import { translate as __, documentLocale } from 'foremanReact/common/I18n';
+import humanizeDuration from 'humanize-duration';
```
laviro: Foreman core uses React-intl and while they doesn't support duration API yet,
there is a lib called [react-intl-formatted-duration](https://github.com/en-japan-air/react-intl-formatted-duration) which support the React-intl provider so you don't need to configure all the localizations by yourself.

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,39 @@ export const addSearchToURL = (path, query) => {
...
+  }
+
+  const dateOptions = {
+    largest: 1,
+    language: documentLocale(),
+    fallbacks: ['en'],
+    round: true,
+  };
+
+  start = new Date(isoCompatibleDate(start));
```
laviro: Not sure if it is a good idea to amend args,
please declare a new const for that.

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,39 @@ export const addSearchToURL = (path, query) => {
...
+  if (start && !finish) {
+    finish = new Date();
+    const duration = finish - start;
+    return {
+      text: `${humanizeDuration(duration, dateOptions)} +`,
+      tooltip: `${__('More than')} ${numberWithDelimiter(
+        (finish - start) / 1000
```
laviro: you can use the duration const which is declared above, instead of re-calculating `finish - start`

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,39 @@ export const addSearchToURL = (path, query) => {
...
+      )} ${__('seconds')}`,
+    };
+  }
+
+  finish = new Date(isoCompatibleDate(finish));
```
laviro: ditto

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,39 @@ export const addSearchToURL = (path, query) => {
...
+    fallbacks: ['en'],
+    round: true,
+  };
+
+  start = new Date(isoCompatibleDate(start));
+  if (start && !finish) {
+    finish = new Date();
+    const duration = finish - start;
+    return {
+      text: `${humanizeDuration(duration, dateOptions)} +`,
```
laviro: I think using `"more than"` as you used in the tooltip would make more sense than the plus sign.

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,39 @@ export const addSearchToURL = (path, query) => {
...
+      tooltip: `${__('More than')} ${numberWithDelimiter(
+        (finish - start) / 1000
+      )} ${__('seconds')}`,
```
laviro: Screenshot from 2019-11-04 15-21-43

I think that when it comes to minutes, the user doesn't need to know how many seconds are in those minutes.

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,39 @@ export const addSearchToURL = (path, query) => {
...
+    };
+  }
+
+  finish = new Date(isoCompatibleDate(finish));
+
+  const duration = finish - start;
+  return {
+    text: humanizeDuration(duration, dateOptions),
```
laviro: Selection_233

if the duration was 0 seconds, I prefer to say "Less than a second"

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: it looks like this library is less used, and updated less often, I'm not sure we should use it

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: Why should it be different then what is in the `Task Raw` page? 
[previous ruby code](https://github.com/MariaAga/foreman-tasks/blob/90a49ddfaae7fb27749939d49371b42b6a2a43d2/app/helpers/foreman_tasks/foreman_tasks_helper.rb#L46)

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: Good catch, thanks!

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: Changed, thanks

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: Added "Less than a second" calc

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
laviro: Does it supports `i18n` like `react-intl`?
I didn't test the `react-intl-formatted-duration` `i18n` but I think it worth to investigate it.
@MariaAga could you add it into our backlog? thanks

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
laviro: @MariaAga I think that when moving a page to React we should always think on how to improve it, or if it deserves a redesign.

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
MariaAga: Added to the backlog, thanks

adamruzicka:
Needs rebase

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,43 @@ export const addSearchToURL = (path, query) => {
...
+
+  const startDate = new Date(isoCompatibleDate(start));
+
+  if (!finish) {
+    finish = new Date();
+    const finishDate = new Date();
+    const duration = finishDate - startDate;
+    return {
+      text: `${__('More than')} ${humanizeDuration(duration, dateOptions)}`,
+      tooltip: `${__('More than')} ${numberWithDelimiter(duration / 1000)} ${__(
```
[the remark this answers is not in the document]
sharvit: Make sense to me, can we see it as a bug fix and do it with a follow-up PR?

webpack/ForemanTasks/Components/TasksTable/TasksTableHelpers.js
```diff
@@ -24,3 +28,43 @@ export const addSearchToURL = (path, query) => {
...
+    largest: 1,
+    language: documentLocale(),
+    fallbacks: ['en'],
+    round: true,
+  };
+
+  const startDate = new Date(isoCompatibleDate(start));
+
+  if (!finish) {
+    finish = new Date();
```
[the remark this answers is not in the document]
MariaAga: nope

adamruzicka:
I'm getting double popups on hover (one appears a bit later after the other).

image

adamruzicka:
Also needs a rebase

MariaAga:
> I'm getting double popups on hover (one appears a bit later after the other).
> 
@adamruzicka 
One of them is for the exact number of seconds, and the other is the cells full name (when the cell is too small to be shown completely.
So for 2 hours one tooltip will be in seconds and the other in hours).
Should I remove the seconds tooltip completely, or maybe not show it if the value is already in seconds?

adamruzicka:
I'd say having two different popups with different styling appearing at different times would be weird, even though they would have different contents.

@laviro opinions?

laviro:
I agree @adamruzicka, it is confusing.
how much work would it be @MariaAga ?
also this PR needs another rebase.

MariaAga:
@adamruzicka  I think to solve this I will remove the exact seconds tooltip and we will discuss later if we need it?

adamruzicka:
Agreed

adamruzicka:
Thank you @MariaAga ! Thank you @sharvit and @laviro for reviews

PR #468, opened 2019-10-22
Fixes #27413 - mount sidekiq console

ezr-ondrej:


lib/foreman_tasks/dynflow/console_authorizer.rb
```diff
@@ -1,7 +1,18 @@
...
+      Setting[:dynflow_enable_console] &&
+        (!Setting[:dynflow_console_require_auth] || Dynflow::ConsoleAuthorizer.new(request).allow?)
+    end
+  end
+
   class Dynflow::ConsoleAuthorizer
-    def initialize(env)
-      @rack_request = Rack::Request.new(env)
+    def self.from_env(env)
+      self.new( Rack::Request.new(env) )
```
adamruzicka: Odd spacing

config/routes.rb
```diff
@@ -50,6 +50,13 @@
     if ForemanTasks.dynflow.required?
       require 'dynflow/web'
       mount ForemanTasks.dynflow.web_console => '/dynflow'
+      if defined? ::Sidekiq
+        require 'sidekiq/web'
+        redis_url = SETTINGS.dig(:dynflow, :redis_url)
+        Sidekiq.redis = { url: redis_url }
+        Sidekiq::Web.set :sessions, false
+        mount Sidekiq::Web => '/sidekiq', :constraints => ForemanTasks::Dynflow::SidekiqConsoleConstraint.new
```
adamruzicka: If I try to access `/foreman_tasks/sidekiq` as a non-admin user, I get routing error. `403` would be nicer, but in general it prevents unauthorized users to go there. I guess it is ok for now

ezr-ondrej: yeah, true, that's what the constraint does.. I will try to improve it - but rather in some followup.

adamruzicka:
Thanks @ezr-ondrej !

ehelms:
Should the 'Dynflow console' button on an individual task now take me to the sidekiq console if I am running the new services?

adamruzicka:
No, dynflow console should take you to dynflow console, sidekiq doesn't know anything about dynflow. If you want to get there you have to manually go to `/foreman_tasks/sidekiq`

ehelms:
Gotcha, thanks. So even with Sidekiq, dynflow console is still the main next level debug location after the tasks page?

adamruzicka:
Exactly, sidekiq console exponses details which could be considered internal

PR #470, opened 2019-11-06
Fixes #28152 - Link tasks with user only if the user exists

adamruzicka:
Commit 2f37f56957260facecdb278fffaa1b2bd7f94eb0 changed the way how we link
tasks to users. Before we used locks, but we didn't have any constraints in
place. This could lead to a situation, where a task has a lock, which links the
task to a user, which doesn't exist anymore. If we tried to perform the
migration introduced in that commit, it would fail with foreign key violation
errors.


Alucard616:
I tested this PR with my foreman instance, which triggered this problem and workd perfectly fine.
No Errors and I was able to upgrade to foreman 1.23.0 + Katello 3.13

Alucard616:
> Lgtm, we should probably cherry pick this to stable branches and release a new version for 1.23 and 1.24

That would be nice, so the update is availabe for me too and i don't have to edit this file manually :)

adamruzicka:
Don't worry, I haven't forgotten about this :) It is coming later today

PR #477, opened 2019-11-15
fixes #28276 - Improve cleanup task description

domitea:


lib/foreman_tasks/tasks/cleanup.rake
```diff
@@ -4,14 +4,14 @@ namespace :foreman_tasks do
...
-        * AFTER       : delete tasks created after *AFTER* period. Expected format is a number followed by the time unit (s,h,m,y), such as '10d' for 10 days
+        * AFTER       : delete tasks started after *AFTER* period. Expected format is a number followed by the time unit (s = seconds, m = minutes, h = hours, d = days, y = years), such as '10d' for 10 days
         * STATES      : comma separated list of task states to touch with the cleanup, by default only stopped tasks are covered, special value all can be used to clean the tasks, disregarding their states
         * NOOP        : set to "true" if the task should not actually perform the deletion
         * VERBOSE     : set to "true" for more verbose output
         * BATCH_SIZE  : the size of batches the tasks get processed in (1000 by default)
         * TASK_BACKUP : set to "true" or "false" to enable/disable task backup
 
-      If none of TASK_SEARCH, BEFORE, STATES is specified, the tasks will be cleaned based
+      If none of TASK_SEARCH, STATES is specified, the tasks will be cleaned based
```
ares: should this mentione AFTER instead of BEFORE? (cc @adamruzicka if that's actually truth)

adamruzicka: It should be completely reworder to describe things better. If `TASK_SEARCH` is set, then it is used and the configuration is ignored. Also `AFTER` and `STATES` are valid only in this scenario. If search is unset, then the cleanup is performed based on the configuration (and setting `AFTER` or `STATES` will raise an exception)

ares:
Thanks, @domitea could you please incorporate the suggestion from above?

domitea:
@ares it is better now and ready for squash?

ares:
On small.PRs like this, you can squash right-away. I think the text is OK, just one one. Instead of "then configuration is used" I'd say "then the cleanup respects the configuration file". Leaving final review and ack to @adamruzicka

adamruzicka:
What @ares said, will merge afterwards

adamruzicka:
Thank you @domitea !

PR #478, opened 2019-11-19
Fixes #28302 - cleanup assets

ezr-ondrej:
Cleanup assets to work with sprockets 4.0 - this will work with sprockets < 4.0 as well, so can be merged safely before upgrade.


adamruzicka:
Thank you @ezr-ondrej !

PR #479, opened 2019-11-19
Fixes #28306 - use label as fallback for tasks action name

MariaAga:
uses label as fallback for tasks action name
and replaces '::' in the label so if the name is too long the name will be in a few lines instead of being on top of the next cell


sharvit:
We are using `reselect` (https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/package.json#L78) to memorize selectors if they based on calculations.
You can read more about memorized selectors here: https://github.com/reduxjs/reselect

adamruzicka:
Created https://projects.theforeman.org/issues/28319 for using `reselect`

adamruzicka:
Thank you @MariaAga !

PR #481, opened 2019-12-04
Fixes #28415 - Fix PG::AmbiguousColumn error on tasks search

adamruzicka:
The original error was
```
'Action failed' error (ActiveRecord::StatementInvalid): PG::AmbiguousColumn:
ERROR: ORDER BY "started_at" is ambiguous | LINE 1: ...,4))
```

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -240,7 +240,7 @@ module ForemanTasks
         sort_by = ordering_params[:sort_by] || 'started_at'
         sort_order = ordering_params[:sort_order] || 'DESC'
         scope = scope.select("*, coalesce(ended_at, current_timestamp) - coalesce(coalesce(started_at, ended_at), current_timestamp) as duration")
-        scope.order("#{sort_by} #{sort_order}")
+        scope.order(sort_by => sort_order)
       end
 
       def task_hash(task)
--- a/test/controllers/api/tasks_controller_test.rb
+++ b/test/controllers/api/tasks_controller_test.rb
@@ -11,6 +11,24 @@ module ForemanTasks
         @request.env['CONTENT_TYPE'] = 'application/json'
       end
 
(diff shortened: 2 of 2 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -286,7 +286,7 @@ def tasks_list(filtered_scope)
         filtered_scope = search_scope.select('DISTINCT foreman_tasks_tasks.*')
 
         ordering_params = {
-          sort_by: params[:sort_by] || 'started_at',
+          sort_by: 'foreman_tasks_tasks.' + (params[:sort_by] || 'started_at'),
```
ezr-ondrej: Couldn't we instead up our game, and use `scope.order(sort_by => sort_order)` in the `ordering_scope`?
I believe than rails will do the table prefixing by itself.

ezr-ondrej: Maybe it needs to be `(sort_by.to_sym => sort_order.downcase.to_sym)`

adamruzicka: Good point

test/controllers/api/tasks_controller_test.rb
```diff
@@ -11,6 +11,24 @@ class Api::TasksControllerTest < ActionController::TestCase
         @request.env['CONTENT_TYPE'] = 'application/json'
       end
 
+      describe 'GET /api/tasks' do
+        it 'lists all tasks with default sorting' do
+          (1..5).map { FactoryBot.create(:dynflow_task, :user_create_task) }
```
ezr-ondrej: This can be done like:
```suggestion
          FactoryBot.create_list(:dynflow_task, 5, :user_create_task) }
```
but it's not a blocker, just a nitpick

ares:
Tests failed on integration tests in core, I don't think this should case regressions. Merging, thanks @adamruzicka and @ezr-ondrej for review!

PR #485, opened 2019-12-11
Fixes #28485 - Fix PG::AmbiguousColumn error on tasks search

adamruzicka:
The original error was
PG::UndefinedColumn: ERROR: column foreman_tasks_tasks.duration does not exist |
LINE 1: ...) as duration FROM "foreman_tasks_tasks"

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -238,9 +238,10 @@ module ForemanTasks
 
       def ordering_scope(scope, ordering_params)
         sort_by = ordering_params[:sort_by] || 'started_at'
+        sort_by = 'foreman_tasks_tasks.' + sort_by if sort_by == 'started_at'
         sort_order = ordering_params[:sort_order] || 'DESC'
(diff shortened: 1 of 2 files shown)


adamruzicka:
CC @ezr-ondrej in the end, using a has for ordering broke sorting by duration, which is not a real column

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -238,9 +238,10 @@ def pagination_scope(scope, search_params)
 
       def ordering_scope(scope, ordering_params)
         sort_by = ordering_params[:sort_by] || 'started_at'
+        sort_by = 'foreman_tasks_tasks.' + sort_by if sort_by == 'started_at'
```
ezr-ondrej: this raises a question if other columns aren't going to be ambigous :(

ares: as of today the sort_by is possible only by started_at and duration. We can improve in future by prefixing everything except for duration quite easily if the need be, so I'm fine taking this as is.

ezr-ondrej:
@ares feel free to merge, my concern is not a blocker.

ares:
Thanks, merging!

PR #486, opened 2019-12-11
Fixes #28487 - Allow searching for tasks by duration

adamruzicka:


app/models/foreman_tasks/task/search.rb
```diff
@@ -22,6 +22,15 @@ def search_by_taxonomy(key, operator, value)
         sql = "foreman_tasks_locks_taxonomy#{uniq_suffix}.resource_id #{operator} ? OR foreman_tasks_locks_taxonomy#{uniq_suffix}.resource_id IS NULL"
         { :conditions => sanitize_sql_for_conditions([sql, value]), :joins => joins }
       end
+
+      # Expects the time in the format "\d+ (seconds|minutes|hours|days|months|years)"
+      SUPPORTED_DURATION_FORMAT = /\A\s*(\d+\s+\b(seconds?|minutes?|hours?|days?|months?|years?)\b)\b\s*\z/i
+      def search_by_duration(key, operator, value)
+        raise "Unsupported duration '#{value}' specified for searching" unless value =~ SUPPORTED_DURATION_FORMAT
+        value = value.strip
+        { :conditions => "coalesce(ended_at, current_timestamp) - coalesce(coalesce(started_at, ended_at), current_timestamp) #{operator} ?::interval",
```
adamruzicka: the `::interval` may very well be a postgresism and we may need to find a workaround for sqlite

adamruzicka: tests are failing on sqlite because of this

pondrejk:
Hi @adamruzicka,
trying this out I'm getting a warning for presumably valid search strings:
 
tasks-msg

Happens basically for everything I put in. Also this is probably not the way the warning was intended to be styled.

Also when hitting search and refreshing the page the greater than sign is not sanitized in the url `/tasks?search=duration+>+1&page=1`, probably leading to

tasks-refresh

adamruzicka:
You are supposed to provide units (`1 second`, `5 minutes` and so on). We could drop this requirement. If we do so, how should we interpret just a number?

pondrejk:
Ok, `duration > "2 seconds"` works as expected, `duration > 2 seconds` and `duration > '2 seconds'` yield the same error.

Requiring units is ok, the challenge is to make user aware of what is required. I'd at least extend the warning with examples of intended usage. 

I also don't get the context help for duration as for other keys:
tasks-hint

Would it be possible to make the warning not to overlap with the table?

Also not sure how N/A is treated in the search, I get it in the results for 'duration < "2 seconds"' but not in  'duration > "2 seconds"'.

pondrejk:
The issue on the message we agreed on is here https://projects.theforeman.org/issues/28524

ares:
The test is not stable

```
--- expected
+++ actual
@@ -1 +1 @@
-[#<ForemanTasks::Task::DynflowTask id: "5d1f9930-cda0-498f-995e-8f80db614d10", type: "ForemanTasks::Task::DynflowTask", label: "Support::DummyDynflowAction", started_at: "2014-10-01 11:15:55", ended_at: "2014-10-01 11:15:57", state: "stopped", result: "success", external_id: "0ac75a12-0ce3-4e9e-9565-da41f32a9b76", parent_task_id: nil, start_at: nil, start_before: nil, action: "", user_id: nil, state_updated_at: "2019-12-17 14:14:39">, #<ForemanTasks::Task::DynflowTask id: "561c2479-1f73-482a-af4a-c2b246b0e3b7", type: "ForemanTasks::Task::DynflowTask", label: "Support::DummyDynflowAction", started_at: "2014-10-01 11:15:55", ended_at: "2014-10-01 11:15:57", state: "stopped", result: "success", external_id: "4e789a94-a337-4c6a-a707-14a897b7c0bc", parent_task_id: nil, start_at: nil, start_before: nil, action: "", user_id: nil, state_updated_at: "2019-12-17 14:14:39">]
+#<ActiveRecord::Relation [#<ForemanTasks::Task::DynflowTask id: "561c2479-1f73-482a-af4a-c2b246b0e3b7", type: "ForemanTasks::Task::DynflowTask", label: "Support::DummyDynflowAction", started_at: "2014-10-01 11:15:55", ended_at: "2014-10-01 11:15:57", state: "stopped", result: "success", external_id: "4e789a94-a337-4c6a-a707-14a897b7c0bc", parent_task_id: nil, start_at: nil, start_before: nil, action: "", user_id: nil, state_updated_at: "2019-12-17 14:14:39">, #<ForemanTasks::Task::DynflowTask id: "5d1f9930-cda0-498f-995e-8f80db614d10", type: "ForemanTasks::Task::DynflowTask", label: "Support::DummyDynflowAction", started_at: "2014-10-01 11:15:55", ended_at: "2014-10-01 11:15:57", state: "stopped", result: "success", external_id: "0ac75a12-0ce3-4e9e-9565-da41f32a9b76", parent_task_id: nil, start_at: nil, start_before: nil, action: "", user_id: nil, state_updated_at: "2019-12-17 14:14:39">]>
 (Minitest::Assertion)
```

It seems you need to enforce loading the collection or compare objects by their id.

adamruzicka:
:green_apple:

adamruzicka:
We're :green_apple:

ares:
It works but there are some unexpected things,
* `duration > 1` returns tasks that says Duration 1 second, I guess that may be some rounding issue
* `duration = 0` returns both tasks with 0 and N/A duration
* `duration < 1 month` does not raise any error while it filters based on seconds, however I'm getting different results that for `duration < 1"

I'd be ok with first two, third one is a bit weird. I assume the right way is to wrapt "1 month" (and other words) in quotes and then it seems to works fine

I wonder if this is a generic scoped search limitation or this custom search function oddity

adamruzicka:
What would happen with the third one if you wrote it as `duration < "1 month"`? I'm pretty sure that if you don't quote it, scoped search interprets it as "(duration < 1) OR (has word month in any searchable field)"

adamruzicka:
And if I recall tasks shown with duration `N/A` have in fact duration of 0

ares:
There seems to be difference between N/A and 0. I'll look again when I'm at the computer, but IIRC and based on my comment, using `< "1 month"` worked as expected. Now with the "or" explanation that makes sense. I didn't understand why these two behaved differently "< 1" and "< 1 month".

I'll look once again and will check sql queries. But probably ready to merge then :-)

adamruzicka:
> There seems to be difference between N/A and 0.

Maybe N/A is "precisely zero" and 0 as shown in the ui is "rounded to zero"?

PR #488, opened 2020-01-02
Fixes #28614 - core API refactor breaks tasks pagination

MariaAga:
in foreman tasks table
After this change:
https://github.com/theforeman/foreman/pull/7301
All the pagination info is in the response and not in the payload


adamruzicka:
Thank you @MariaAga & @sharvit

PR #489, opened 2020-01-02
Fixes #28095 - Make concurrency control work with batch triggering

adamruzicka:
Having both concurrency control and batch triggering enabled led to various issues with tasks being triggered at wrong time. This was caused by introduction of Remote Tasks and their triggering using a middleware.

This change makes sure the two mechanisms will not conflict with each other. If use_concurrency_control is passed to the proxy action, the remote tasks are created in 'external' state, meaning the batch triggering mechanism should ignore them. This allows the concurrency control to work properly.

### DIFF
--- a/app/lib/actions/proxy_action.rb
+++ b/app/lib/actions/proxy_action.rb
@@ -41,16 +41,7 @@ module Actions
       with_connection_error_handling(event) do |event|
         case event
         when nil
-          if remote_task
-            if remote_task.state == 'external'
-              trigger_remote_task
-            else
-              on_resume
-            end
-          else
-            trigger_proxy_task
-          end
-          suspend
+          start_or_resume
         when ::Dynflow::Action::Skip
           # do nothing
         when ::Dynflow::Action::Cancellable::Cancel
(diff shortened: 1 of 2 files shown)


ares:
code LGTM, one question - does that mean that the concurrency value is completely ignored? also it seems rubocop is unhappy about cyclomatic complexity of something in core, perhaps we need to sync our cop definitions?

adamruzicka:
> does that mean that the concurrency value is completely ignored?

No, concurrency value is still honored, but batch triggering is ignored. It still uses the new batch triggering API, but uses it for each host separately.

ares:
ah ok, we'll need to explain that in REX setting and concurrency limit inline help, but that belongs to https://github.com/theforeman/foreman_remote_execution/pull/458

adamruzicka:
We could in theory come up with a way that batch triggering would be used for the first `[batch_size, concurrency_level].min` tasks. After this initial batch, the other tasks would be triggered one by one. It would be a bigger change and I don't think it is worth it

adamruzicka:
[test foreman-tasks]

adamruzicka:
Naming is hard :/

adamruzicka:
Thank you @ezr-ondrej for review!

PR #490, opened 2020-01-02
Fixes #28776 - Adding Jail for task model

domitea:
Piece of code that is necessary for correct working of https://github.com/theforeman/foreman_remote_execution/pull/453

### DIFF
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -70,6 +70,10 @@ module ForemanTasks
            end)
     scope :for_action_types, (->(action_types) { where('foreman_tasks_tasks.label IN (?)', Array(action_types)) })
 
+    class Jail < Safemode::Jail
+      allow :started_at, :ended_at, :result, :state, :label, :main_action
+    end
+
     def input
       {}
     end


app/models/foreman_tasks/task.rb
```diff
@@ -223,6 +227,13 @@ def to_label
       parts.join(' ').strip
     end
 
+    def output_by_type
+      return {} unless main_action && main_action.live_output.any?
+      main_action.live_output.each_with_object({}) do |output, o_hash|
```
ares: would it be more readable if we hardcode 'stderr' and 'stdout' here? I don't think we'll see any more output tymes in foreseeable future :-)

also I think it should always return hash with stdout and stderr keys, they should just have empty value

something like
```
return { :stderr => nil, :stdout => nil} unless main_action && main_action.live_output.any?
{ :stderr => ..., :stdout => ... }
```

domitea: I like to avoid hardcoding because it tends to create piece of code that is forgotten when the new "out" (ie. `stdwarn`) stream is created. I know it's a extreme example but I don't know how frequently can happen. What do you think  @ares

ares: There are more places where we hardcode this, e.g. https://github.com/theforeman/foreman_remote_execution/blob/master/app/views/template_invocations/show.html.erb#L17-L22

if user defines STDWARN, our tool would somehow have to detecte it, but the way we run SSH / Ansible does not allow it today. Given the chance here is close to zero, I'd be fine with hardcoding.

But if you feel strongly it's better to be smart here and iterate over all available statuses, I don't insist :-) I'd find the return value of this method more readable, perhaps adding a comment describing the result would do.

adamruzicka:
What's the status here?

domitea:
@ares Attributes added, issue created, commits squashed.

app/models/foreman_tasks/task.rb
```diff
@@ -223,6 +227,15 @@ def to_label
       parts.join(' ').strip
     end
 
+    # Method that produces hash where key is output type (ie. stdout, stderr...) and value is output for given type
+    # So the returned hash look like {'stdout': 'normal output', 'stderr': 'error'}
+    def output_by_type
+      return {} unless main_action && main_action.live_output.any?
```
adamruzicka: There's only one action class which has `#live_output` defined. If I'm reading this right, this would raise an `NoMethodError` if called on anything else than `Actions::RemoteExecution::RunHostJob`

app/models/foreman_tasks/task.rb
```diff
@@ -223,6 +227,15 @@ def to_label
       parts.join(' ').strip
     end
 
+    # Method that produces hash where key is output type (ie. stdout, stderr...) and value is output for given type
+    # So the returned hash look like {'stdout': 'normal output', 'stderr': 'error'}
+    def output_by_type
```
adamruzicka: Given there is only one action for which this will actually return something meaningful and that one action comes from a different plugin, shouldn't it live there?

domitea: This method was originally in macro version in different plugin, but with the discussion with @ares I moved it to here due to better calling in templates and possibility of future usage in another places. Do you suggest to move to another place?

ares: well, if we want to keep it here, it needs to check whether `live_output` is defined on `main_action` right? perhaps I confused something, however I find task output in safe mode useful. However this macro is for job template output (stdout and stderr), which is only relevant for rex tasks, so it would probably make sense to live there. Not every task has stdout/stderr

adamruzicka: > if we want to keep it here, it needs to check whether live_output is defined on main_action right?

It needs to check whether `live_output` is defined on `main_action`, no matter where it lives, just to stay on the safe side of things.

> however I find task output in safe mode useful.

By task output you mean the thing you get by calling `Task#output` or something like REX does?

tbrisker:
the commit message should probably be "fixes.."

tbrisker:
ping @domitea

ares:
There's also new request for "failed tasks report" at https://bugzilla.redhat.com/show_bug.cgi?id=1836396, it may be good to incorporate requested attributes in this PR.

PR #491, opened 2020-01-02
Fixes #28319 - selectResults selector should be memorized

MariaAga:
Using  https://github.com/reduxjs/reselect so the selector/formatter for the results is not recomputed unless the result list changes.

### DIFF
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
@@ -1,3 +1,4 @@
+import { createSelector } from 'reselect';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { selectForemanTasks } from '../../ForemanTasksSelectors';
 import { getDuration } from './TasksTableHelpers';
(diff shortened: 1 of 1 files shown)


webpack/ForemanTasks/Components/TasksTable/TasksTableSelectors.js
```diff
@@ -27,18 +28,20 @@ export const selectSelectedRows = state =>
...
-    state: result.state + (result.frozen ? ` ${__('Disabled')}` : ''),
-    duration: getDuration(result.started_at, result.ended_at),
-    availableActions: result.available_actions,
-  }));
-};
+export const selectResults = createSelector(
+  selectTasksTableContent,
+  ({ results }) =>
+    results.length
+      ? results.map(result => ({
```
adamruzicka: Is this necessary? `[].map(resutl => $something)` is still `[]`. Or am I missing something and `results` can be something that cannot be `map`ped over?

MariaAga:
@jeremylenz I agree we shouldn't use reselect across the entire code base.
The use of it here is because of the many calculations that the selector does.
Thanks for checking the performance!

adamruzicka:
[test foreman-tasks]

adamruzicka:
@jeremylenz does your last comment mean this PR has your blessing?

jeremylenz:
@adamruzicka @MariaAga yes, ACK

adamruzicka:
Thank you @MariaAga and @jeremylenz for review!

PR #492, opened 2020-01-03
Fixes #28623 - fix table reducer action structure

laviro:
After Foreman updated the API middleware structure,
we forgot to update the foreman-tasks table that uses it.


adamruzicka:
I'm going to close this in favor of https://github.com/theforeman/foreman-tasks/pull/488 which was opened first. Thank you for opening this PR, love how we got two exactly same patches for this :)

laviro:
> love how we got two exactly same patches for this :)

Great minds think alike... or thats the only fix :rofl:

PR #495, opened 2020-01-07
Fixes #28670 - missing identity-obj-proxy package for tests

MariaAga:
Every JS test that has scss fails because of missing identity-obj-proxy package

```
    Configuration error:
    Could not locate module ./TaskDetails.scss mapped as:
    identity-obj-proxy.
    Please check your configuration for these entries:
    {
      "moduleNameMapper": {
        "/^.+\.(png|gif|css|scss)$/": "identity-obj-proxy" 
      },
      "resolver": null
    }
```


adamruzicka:
Thank you both!

PR #496, opened 2020-01-12
Fixes #28729 - Add ForemanContext for metadata

amirfefer:
Using https://github.com/theforeman/foreman/pull/7139


amirfefer:
`ForemanContext` will be on every component by default, no need to add it manually on each plugin. closing...

PR #499, opened 2020-01-21
Fixes #28822 - task Search requires clicking twice on Search

MariaAga:
The url got updated and then the get tasks api was called twice - once with the right url and once with the old url.
So on the second click the old url was already the new url so when it was called twice it was called with the right info


adamruzicka:
Thank you @MariaAga !

PR #500, opened 2020-02-02
Fixes #20042 - add a clear button to recurring logics

MariaAga:
In the index of recurring logics there should be a button to clear cancelled recurring logics
image

### DIFF
--- a/app/controllers/foreman_tasks/recurring_logics_controller.rb
+++ b/app/controllers/foreman_tasks/recurring_logics_controller.rb
@@ -22,6 +22,12 @@ module ForemanTasks
       redirect_to :action => :index
     end
 
+    def clear_cancelled
+      scope = resource_base.search_for('state=cancelled')
+      scope.destroy_all
+      redirect_to :action => :index
+    end
+
     def controller_name
       'foreman_tasks_recurring_logics'
     end
@@ -48,6 +54,15 @@ module ForemanTasks
     def filter(scope)
       scope.search_for(params[:search])
            .paginate(:page => params[:page], :per_page => params[:per_page])
+    end
+
+    def action_permission
+      case params[:action]
+      when 'clear_cancelled'
+        'edit'
+      else
+        super
+      end
     end
   end
 end
--- a/app/views/foreman_tasks/recurring_logics/index.html.erb
+++ b/app/views/foreman_tasks/recurring_logics/index.html.erb
@@ -5,6 +5,36 @@
 <%= alert(:class => 'alert-info', :id => 'multiple-alert', :close => false, :header => '', :text => @errors) %>
 <% end %>
 
+<% if authorized_for(:permission => :edit_recurring_logics, :auth_object => @recurring_logics) %>
+  <% title_actions link_to(_('Clear Cancelled'),
+              clear_cancelled_foreman_tasks_recurring_logics_path,
+              class:  ['btn', 'btn-sm', 'btn-danger'],                  
+              :'data-toggle' => "modal",
+              :'data-target' => "#clear_modal") 
+              %>
+
+  <div class="modal fade" id="clear_modal" tabindex="-1" role="dialog" aria-labelledby="Deploy" aria-hidden="true">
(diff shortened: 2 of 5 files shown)


app/controllers/foreman_tasks/recurring_logics_controller.rb
```diff
@@ -22,6 +22,16 @@ def cancel
       redirect_to :action => :index
     end
 
+    def clear_cancelled
+      scope = resource_base.search_for('state=cancelled') || []
+      scope.each do |recurring_logic|
+        begin
+          recurring_logic.destroy
+        end
```
adamruzicka: You can remove the `begin` and `end`, they don't really do anything here

app/controllers/foreman_tasks/recurring_logics_controller.rb
```diff
@@ -22,6 +22,16 @@ def cancel
       redirect_to :action => :index
     end
 
+    def clear_cancelled
+      scope = resource_base.search_for('state=cancelled') || []
+      scope.each do |recurring_logic|
```
adamruzicka: I think you should be able to call `#destroy_all` on the collection instead of having to iterate through the items and destroying them one by one

app/controllers/foreman_tasks/recurring_logics_controller.rb
```diff
@@ -22,6 +22,16 @@ def cancel
       redirect_to :action => :index
     end
 
+    def clear_cancelled
+      scope = resource_base.search_for('state=cancelled') || []
```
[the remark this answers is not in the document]
MariaAga: thanks! I'm still learning :)

MariaAga:
@pondrejk Thanks for the review, I moved the button to the right of the table so it will be "the foerman way".
But I think disabling the button when there are no cancelled logics is too complicated and only has a cosmetic purpose

adamruzicka:
Ideally there should be some kind of "empty page", but that is out of scope of this.

app/views/foreman_tasks/recurring_logics/index.html.erb
```diff
@@ -5,6 +5,11 @@
 <%= alert(:class => 'alert-info', :id => 'multiple-alert', :close => false, :header => '', :text => @errors) %>
 <% end %>
 
+<% title_actions link_to(_('Clear Cancelled'),
+            clear_cancelled_foreman_tasks_recurring_logics_path,
+            class:  ['btn', 'btn-sm', 'btn-primary'].compact,
+            method: :post ) %>
```
adamruzicka: The button should not be shown if the user doesn't have the `edit_recurring_logics` permission assigned. For reference take a look at https://github.com/theforeman/foreman_remote_execution/blob/master/app/helpers/remote_execution_helper.rb#L69

adamruzicka: Also that `.compact` doesn't really do anything there so you can drop it

MariaAga: Thanks for the explanation, fixed

app/views/foreman_tasks/recurring_logics/index.html.erb
```diff
@@ -5,6 +5,13 @@
 <%= alert(:class => 'alert-info', :id => 'multiple-alert', :close => false, :header => '', :text => @errors) %>
 <% end %>
 
+<% if authorized_for(:permission => :edit_recurring_logics, :auth_object => @recurring_logics) %>
+  <% title_actions link_to(_('Clear Cancelled'),
+              clear_cancelled_foreman_tasks_recurring_logics_path,
+              class:  ['btn', 'btn-sm', 'btn-primary'],
```
adamruzicka: Shouldn't the button be red to denote it is a possibly destructive action?

MariaAga: Yeah you're right. I also added a confirm dialog

adamruzicka:
Thank you @MariaAga & @pondrejk !

PR #501, opened 2020-02-05
Fixes #28933 - add api to delete recurring logics

MariaAga:
recurring logics by search query, to prevent accidental removal of all the recurring logics the search is required.

### DIFF
--- a/app/controllers/foreman_tasks/api/recurring_logics_controller.rb
+++ b/app/controllers/foreman_tasks/api/recurring_logics_controller.rb
@@ -10,6 +10,12 @@ module ForemanTasks
         api_base_url '/foreman_tasks/api'
       end
 
+      # Foreman right now doesn't have mechanism to
+      # cause general BadRequest handling, resuing the Apipie::ParamError
+      # for now http://projects.theforeman.org/issues/3957
+      class BadRequest < Apipie::ParamError
+      end
+
       before_action :find_resource, :only => %w[show cancel update]
 
       api :GET, '/recurring_logics', N_('List recurring logics')
@@ -39,9 +45,22 @@ module ForemanTasks
         ForemanTasks::RecurringLogic
       end
 
+      api :POST, '/recurring_logics/bulk_destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        if params[:search].blank?
+          raise BadRequest, _('Please provide a search parameter in the request')
+        end
+        scope = resource_scope.search_for(params[:search])
+        scope.each(&:destroy!)
+        render json: { destroyed: scope }
+      rescue ActiveRecord::RecordNotDestroyed => error
+        render json: { error: error, scope: scope }, status: :bad_request
+      end
+
       def action_permission
         case params[:action]
-        when 'cancel'
+        when 'cancel', 'bulk_destroy'
           'edit'
         else
           super
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -42,6 +42,9 @@ Foreman::Application.routes.draw do
         member do
           post :cancel
         end
+        collection do
+          post :bulk_destroy
+        end
       end
 
       resources :tasks, :only => [:show, :index] do
--- a/lib/foreman_tasks/engine.rb
+++ b/lib/foreman_tasks/engine.rb
@@ -61,7 +61,7 @@ module ForemanTasks
                                                :'foreman_tasks/api/recurring_logics' => [:index, :show] }, :resource_type => ForemanTasks::RecurringLogic.name
 
           permission :edit_recurring_logics, { :'foreman_tasks/recurring_logics' => [:cancel, :enable, :disable, :clear_cancelled],
-                                               :'foreman_tasks/api/recurring_logics' => [:cancel, :update] }, :resource_type => ForemanTasks::RecurringLogic.name
+                                               :'foreman_tasks/api/recurring_logics' => [:cancel, :update, :bulk_destroy] }, :resource_type => ForemanTasks::RecurringLogic.name
         end
 
         add_all_permissions_to_default_roles


app/controllers/foreman_tasks/api/recurring_logics_controller.rb
```diff
@@ -38,10 +38,22 @@ def cancel
...
+                      
+      api :POST, '/recurring_logics/bulk/destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        scope = resource_scope.search_for(params[:search])
+        destroyed = scope.destroy_all
+        if destroyed
+          render json: { destroyed: destroyed }
+        else
+          render json: {}, status: :bad_request
```
xprazak2: I haven't tested the errors yet, but would we want to return some sort of error message?

MariaAga: What kind of error message? `recurring logics could not be destroyed`?

xprazak2: The message that would tells us what went wrong and why this is a bad request. Returning error status with empty response does not provide any hints.

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
```diff
@@ -38,10 +38,22 @@ def cancel
       def resource_class
         ForemanTasks::RecurringLogic
       end
+                      
+      api :POST, '/recurring_logics/bulk/destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        scope = resource_scope.search_for(params[:search])
+        destroyed = scope.destroy_all
+        if destroyed
```
xprazak2: `destroy_all` returns a collection of objects that were destroyed and calling it on an empty scope returns empty array, so the condition will always be true

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
MariaAga: added the error if destroy fails

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
MariaAga: changed to `destroy!` to get the error, and `destroyed` is now just the scope

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
```diff
@@ -39,9 +39,19 @@ def resource_class
...
       end
 
+      api :POST, '/recurring_logics/bulk/destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        scope = resource_scope.search_for(params[:search])
+        scope.each(&:destroy!)
+        render json: { destroyed: scope }
+        rescue ActiveRecord::RecordNotDestroyed => error
+          render json: { error: error, scope: scope }, status: :bad_request
```
adamruzicka: This should be indented one level to the left

config/routes.rb
```diff
@@ -39,6 +39,9 @@
         member do
           post :cancel
         end
+        collection do
+          match '/bulk/destroy' => :bulk_destroy, :via => [:post]
```
adamruzicka: How about we make the path match what we do for [other bulk actions](https://github.com/theforeman/foreman-tasks/blob/master/config/routes.rb#L51) (`bulk_destroy` vs `bulk/destroy`). Then you should be able to do something like
```ruby
collection do
  post :bulk_destroy
end
```

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
MariaAga: why? this line should happen only for the rescue

adamruzicka: If you do not use explicit `begin` and `end`, then `rescue` should be aligned to the same level as `def` and `end`

All these three are functionally equivalent, however the first one is preferred.

```ruby
def foo
  raise "Stop"
rescue => e
  puts e
end
```

```ruby
def foo
  raise "Stop"
  rescue => e
    puts e
end
```

```ruby
def foo
  begin
    raise "Stop"
  rescue => e
    puts e
  end
end
```

MariaAga: Thanks for the explanation! fixed

config/routes.rb
MariaAga: I was taking inspiration from hosts but this makes more sense, thanks!

adamruzicka:
[test foreman_tasks]

adamruzicka:
[test foreman-tasks]

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
```diff
@@ -39,9 +39,19 @@ def resource_class
         ForemanTasks::RecurringLogic
       end
 
+      api :POST, '/recurring_logics/bulk_destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        scope = resource_scope.search_for(params[:search])
```
adamruzicka: We need to check if `params[:search]` is actually provided. Otherwise it matches *all* recurring logics.

MariaAga: `:required => true` doesn't this block it?

adamruzicka: It just shows up in the api documentation as required, but you have to enforce it yourself. I just accidentally wiped all my recurring logics :)

MariaAga: Oh no! Added an `ArgumentError`

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
```diff
@@ -39,9 +39,22 @@ def resource_class
         ForemanTasks::RecurringLogic
       end
 
+      api :POST, '/recurring_logics/bulk_destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        if params[:search].nil?
```
adamruzicka: How about using `blank?` instead of `nil?` so it also catches empty strings?

MariaAga: I thought we wanted to keep the option to destroy all by sending an empty string?

adamruzicka: People will still be able to destroy all by explicitly matching on all with something like `state ~ *`

MariaAga: Changed to blank and changed the argument error to be `BadRequest`

app/controllers/foreman_tasks/api/recurring_logics_controller.rb
```diff
@@ -39,9 +45,22 @@ def resource_class
         ForemanTasks::RecurringLogic
       end
 
+      api :POST, '/recurring_logics/bulk_destroy', N_('Delete recurring logics by search query')
+      param :search, String, :desc => N_('Search query'), :required => true
+      def bulk_destroy
+        if params[:search].blank?
+          raise BadRequest, _('Please provide a search parameter in the request')
```
adamruzicka: How about using something like this instead of creating a custom exception class?
```ruby
payload = { :error => { :message => _("Required parameter 'search' missing in the request") } }
render :json => payload, :status => :bad_request
return
```

MariaAga: Why not use `BadRequest`? it's used in tasks

adamruzicka: Ah, you're right, completely forgot about it. Nevermind then

adamruzicka:
Thank you @MariaAga !

PR #505, opened 2020-02-17
Fix search for multiple resource_ids

m-bucher:
@adamruzicka something like that?

see also https://github.com/Katello/katello/pull/8562#discussion_r379355124

### DIFF
--- a/app/models/foreman_tasks/task/search.rb
+++ b/app/models/foreman_tasks/task/search.rb
@@ -4,7 +4,8 @@ module ForemanTasks
       def search_by_generic_resource(key, operator, value)
         key = 'resource_type' if key.blank?
         key_name = connection.quote_column_name(key.sub(/^.*\./, ''))
-        condition = sanitize_sql_for_conditions(["foreman_tasks_locks.#{key_name} #{operator} ?", value])
+        value = value.split(',') if operator.index(/IN/i)
+        condition = sanitize_sql_for_conditions(["foreman_tasks_locks.#{key_name} #{operator} (?)", value])
 
         { :conditions => condition, :joins => :locks }
       end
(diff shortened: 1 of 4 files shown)


app/models/foreman_tasks/task/search.rb
```diff
@@ -4,7 +4,8 @@ module Search
       def search_by_generic_resource(key, operator, value)
         key = 'resource_type' if key.blank?
         key_name = connection.quote_column_name(key.sub(/^.*\./, ''))
-        condition = sanitize_sql_for_conditions(["foreman_tasks_locks.#{key_name} #{operator} ?", value])
+        value = value.split(',') unless operator.index(/IN/i).nil?
```
adamruzicka: I'd just switch from `unless it does not match` to `if it matches`

m-bucher:
While working on a unit-test for this, I saw that this will return duplicates of the task for every Lock that is found.
So I think we have to add a `distinct` to the search.
However, I do not know how to do this for scoped_search.
@adamruzicka any idea?

adamruzicka:
This cannot be done in scoped search directly. I'd probably add something like

```ruby
scope :for_distinct_search,
      (->(search) { search_for(search).select("DISTINCTI foreman_tasks_tasks.*") })
```

Into `app/models/foreman_tasks/task.rb` and then use `ForemanTasks::Task.for_distinct_search` instead of `ForemanTasks::Task.search_for` when only dinstinct record are needed

m-bucher:
I see, should I add something like that to this PR or should we leave it to those using this search?
I will add a `.distinct` to my test for now.

adamruzicka:
Just add a comment above the search definition that it may return duplicates

adamruzicka:
[test foreman-tasks]

adamruzicka:
Thank you @m-bucher and congratulations to first merged PR into foreman-tasks!

jturel:
very nice!

PR #506, opened 2020-02-18
Fixes #29023 - Add title to tasks page

xprazak2:


adamruzicka:
While you're at it, could you also fix it for task details?

Also when you look at a task's subtasks, the title still says just "Tasks", maybe something like "Subtasks of #{task.label}" would be nicer, but I won't block this PR on this.

xprazak2:
Updated.

adamruzicka:
Thank you @xprazak2 and @MariaAga !

PR #507, opened 2020-02-23
Fixes #29136 - unable to cancel-resume task from table

MariaAga:
When trying to click on the Cancel (or Resume) cell in the tasks table the task isn't cancelled and a toast error shows:  ' "undefined" Task cannot be cancelled at the moment.'


adamruzicka:
Thank you @MariaAga !

PR #508, opened 2020-02-25
Fixes #29045 - HTML tags visible on paused task page

MariaAga:
adds the `react-html-parser` library that "Converts HTML strings directly into React components avoiding the need to use dangerouslySetInnerHTML"
and uses it for the troubleshooting


MariaAga:
Fixed :)

adamruzicka:
Thank you @MariaAga !

PR #510, opened 2020-02-26
Fixes #29180 - Remove unused tar verbose flag in tasks export

netzvieh:
Removes the unused verbose flag from the tar command to compress
exported tasks along with the now unnecessary output redirection to
/dev/null.

### DIFF
--- a/lib/foreman_tasks/tasks/export_tasks.rake
+++ b/lib/foreman_tasks/tasks/export_tasks.rake
@@ -259,7 +259,7 @@ namespace :foreman_tasks do
 
         File.open(File.join(tmp_dir, 'index.html'), 'w') { |file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks))) }
 
-        sh("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
+        system("tar", "czf", export_filename, tmp_dir)
       end
     elsif format == 'csv'
       CSV.open(export_filename, 'wb') do |csv|


netzvieh:
So do I need to open a redmine issue for this first?

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -259,7 +259,7 @@ namespace :foreman_tasks do
 
         File.open(File.join(tmp_dir, 'index.html'), 'w') { |file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks))) }
 
-        sh("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
+        sh("tar czf #{export_filename} #{tmp_dir}")
```
adamruzicka: Since we don't need to spawn a shell anymore, could you change this to use [Kernel#system](https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-system) using the `cmdname, arg1, ...` form?

netzvieh: Sure! Done.

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -259,7 +259,7 @@ namespace :foreman_tasks do
 
         File.open(File.join(tmp_dir, 'index.html'), 'w') { |file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks))) }
 
-        sh("tar cvzf #{export_filename} #{tmp_dir} > /dev/null")
+        system("tar", "czf", "#{export_filename}", "#{tmp_dir}")
```
adamruzicka: Now you can drop the string interpolation as it serves no purpose and do `system("tar", "czf", export_filename, tmp_dir)`

netzvieh: Thanks for the suggestion and sorry for the delay. Changed as suggested.

adamruzicka:
Thank you @netzvieh !

PR #512, opened 2020-02-27
Fixes #29203 - Fix an edge case with excessive progress precision

adamruzicka:
Before this patch we took a float in the range from 0 to 1, turned it into a
string with desired precision and then multiplied it with 100 to get the
percentage. However, this didn't really work for all inputs, because of the well
known issues when multiplying floats.

```javascript
> x = 0.004545454545454546
0.004545454545454546

// Old way
> x.toFixed(4) * 100
0.44999999999999996

// New way
> parseFloat((x * 100).toFixed(2))
0.45
```


adamruzicka:
Thank you for review

PR #513, opened 2020-02-27
Fixes #29030 - dynflow console button is no longer disabled

MariaAga:
Now rails sends the `dynflow_enable_console` info to react.
fixes the snapshots for the new prop and also for the header PR

### DIFF
--- a/app/views/foreman_tasks/api/tasks/details.json.rabl
+++ b/app/views/foreman_tasks/api/tasks/details.json.rabl
@@ -16,3 +16,4 @@ node(:locks) do
   end
 end
 node(:username_path) { username_link_task(@task.owner, @task.username) }
+node(:dynflow_enable_console ) { Setting['dynflow_enable_console'] }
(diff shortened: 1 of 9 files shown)


adamruzicka:
When I hover over other disabled buttons (such as resume), the mouse cursor turns into a stop sign, however when hovering over disabled dynflow console, it doesn't change. Is this something easily fixable?

MariaAga:
It didn't have one because the button had `href` instead of `onClick`. Added the right cursor in the css

adamruzicka:
> Fixed! It didn't have one because the button had `href` instead of `onClick`. Added the right cursor in the css.

I'm still seeing the same behavior on both firefox and chrome.

MariaAga:
weird, it worked once and not again.
Looks like its a css prop from bootstrap that I can't really override.

MariaAga:
Yes that is because in Patternfly a Button component with **clickOn** is a `<button>` but with an **href** its an `<a>`.

PR #514, opened 2020-02-28
Fixes #29223 - add latest_tasks by resource_ids

m-bucher:
see also https://github.com/Katello/katello/pull/8562#discussion_r379009700

### DIFF
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -255,6 +255,32 @@ module ForemanTasks
       main_action.continuous_output.raw_outputs
     end
 
+    def self.latest_tasks_by_resource_ids(label, resource_type, resource_ids)
+      tasks = arel_table
+      links = ForemanTasks::Link.arel_table
+      started_at = tasks[:started_at]
+      resource_id = links[:resource_id]
+
+      base_combined_table = tasks
+                            .join(links).on(tasks[:id].eq(links[:task_id]))
+                            .where(tasks[:label].eq(label)
+                                                .and(links[:resource_type].eq(resource_type))
+                                                .and(links[:resource_id].in(resource_ids)))
+
+      grouped = base_combined_table.project(
+        started_at.maximum.as('started_at_max'),
+        resource_id
+      ).group(resource_id).order(resource_id).as('grouped')
+
+      max_per_resource_id = tasks
+                            .join(links).on(tasks[:id].eq(links[:task_id]))
+                            .join(grouped).on(grouped[:started_at_max].eq(started_at).and(grouped[:resource_id].eq(resource_id)))
+                            .distinct
+                            .project(tasks[Arel.star], grouped[:resource_id])
+
+      find_by_sql(max_per_resource_id.to_sql).index_by(&:resource_id)
+    end
+
     protected
 
     def generate_id
--- a/test/unit/task_test.rb
+++ b/test/unit/task_test.rb
@@ -296,12 +296,12 @@ class TasksTest < ActiveSupport::TestCase
   end
 
   describe 'search for resource_ids' do
-    it 'finds tasks' do
-      label = 'label1'
-      resource_ids = [1, 2]
-      resource_type = 'restype1'
+    label = 'label1'
+    resource_ids = [1, 2]
+    resource_type = 'restype1'
 
-      task1_old = FactoryBot.create(
+    let(:task1_old) do
+      FactoryBot.create(
         :task_with_links,
         started_at: '2019-10-01 11:15:55',
         ended_at: '2019-10-01 11:15:57',
@@ -309,7 +309,9 @@ class TasksTest < ActiveSupport::TestCase
         label: label,
         resource_type: resource_type
       )
-      task1_new = FactoryBot.create(
+    end
+    let(:task1_new) do
+      FactoryBot.create(
         :task_with_links,
         started_at: '2019-10-02 11:15:55',
         ended_at: '2019-10-02 11:15:57',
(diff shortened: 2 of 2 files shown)


app/models/foreman_tasks/task.rb
```diff
@@ -225,6 +225,28 @@ def to_label
...
+
+      max_per_resource_id = self::DynflowTask.joins(
+        :locks
+      ).select(
+        "#{self::DynflowTask.table_name}.*, inner_select.resource_id"
+      ).joins(
+        "INNER JOIN (#{grouped.to_sql}) inner_select
+        ON inner_select.started_at = #{self::DynflowTask.table_name}.started_at
+        AND inner_select.resource_id = locks_foreman_tasks_tasks.resource_id"
+      ).distinct
```
adamruzicka: I'm not really happy about this, it's a lot of sql strings :/

app/models/foreman_tasks/task.rb
```diff
@@ -225,6 +225,28 @@ def to_label
...
+        "MAX(#{self::DynflowTask.table_name}.started_at) AS started_at",
+        "#{ForemanTasks::Lock.table_name}.resource_id AS resource_id"
+      ).group("#{ForemanTasks::Lock.table_name}.resource_id")
+
+      max_per_resource_id = self::DynflowTask.joins(
+        :locks
+      ).select(
+        "#{self::DynflowTask.table_name}.*, inner_select.resource_id"
+      ).joins(
+        "INNER JOIN (#{grouped.to_sql}) inner_select
```
adamruzicka: Shouldn't it be `INNER JOIN ($subquery) AS inner_select`? Or is the `AS` actually optional?

m-bucher: As far as I know with tables, which is how the subquery is used here, the `AS` is not necessary.
See also https://www.postgresql.org/docs/10/sql-select.html:
```
where from_item can be one of:

    [ ONLY ] table_name [ * ] [ [ AS ] alias [ ( column_alias [, ...] ) ] ]
...
    from_item [ NATURAL ] join_type from_item [ ON join_condition | USING ( join_column [, ...] ) ]

File: app/models/foreman_tasks/task.rb (hunk 1 above)
Comment (m-bucher) [reply app/models/foreman_tasks/task.rb]: Typical collision of ruby, ActiveRecord and SQL :wink: 
You can do this with nice ruby-code, but it will be terribly slow.

I agree it is not really self-explaining, but has the best performance.

File: app/models/foreman_tasks/task.rb (hunk 2 above)
Comment (adamruzicka) [reply app/models/foreman_tasks/task.rb]: Good to know, thanks

File: app/models/foreman_tasks/task.rb (hunk 1 above)
Comment (adamruzicka) [reply app/models/foreman_tasks/task.rb]: How about using Arel directly? It wouldn't be nice ruby, but at least it would be ruby.

Comment (m-bucher) [reply app/models/foreman_tasks/task.rb]: I will look into that, thanks @adamruzicka

File: app/models/foreman_tasks/task.rb [hunk 3]
```diff
@@ -225,6 +225,31 @@ def to_label
...
+      ).group(resource_id)
+
+      max_per_resource_id = self::DynflowTask.joins(
+        :locks
+      ).select(
+        "#{self::DynflowTask.table_name}.*, inner_select.resource_id"
+      ).joins(
+        "INNER JOIN (#{grouped.to_sql}) inner_select
+        ON inner_select.started_at = #{started_at}
+        AND inner_select.resource_id = locks_foreman_tasks_tasks.resource_id"
```
m-bucher: I reckon this might be tricky, because of the very customized looking alias `locks_foreman_tasks_tasks`.
Is there a way to use arel for that?

m-bucher:
@adamruzicka sorry it took me a while to get back at this.
I finally managed to do a complete arel-approach.
Unfortunately, two of the tests did not terminate. Does this have to do with my changes?

app/models/foreman_tasks/task.rb
```diff
@@ -255,6 +255,33 @@ def action_continuous_output
       main_action.continuous_output.raw_outputs
     end
 
+    def self.get_latest_tasks_by_resource_ids(label, resource_ids, resource_type)
+      tasks = arel_table
+      dynflow_tasks = self::DynflowTask.arel_table
```
adamruzicka, line 260: is the `self::` necessary?

adamruzicka (commented):
Code wise looks good and most likely doesn't cause the tests to timeout. I'll need to do a bit of testing with it

app/models/foreman_tasks/task.rb
m-bucher: Probably not, but rubocop did not complain.
I am not fully sure, whether I need to differentiate between `dynflow_tasks` and `tasks` because they resolve the same SQL-table :thinking:

adamruzicka: Probably not

adamruzicka: Let's get rid of it, it seems to work just fine without it

app/models/foreman_tasks/task.rb
```diff
@@ -255,6 +255,33 @@ def action_continuous_output
       main_action.continuous_output.raw_outputs
     end
 
+    def self.get_latest_tasks_by_resource_ids(label, resource_ids, resource_type)
```
adamruzicka, line 258: Could we drop the `get_` prefix?

Could we swap the `resource_type` and `resource_ids` args so it reads more like "give me all tasks with $label which are related to $resource_type with ids $resource_ids"?

adamruzicka (commented):
A couple of nitpicks, but it works nicely :+1:

adamruzicka (approved):

adamruzicka:
Thank you @m-bucher !

PR #515, opened 2020-03-02
Fixes #27951 - add dynflow requirement

mmoll:


adamruzicka:
Thank you @mmoll !

mmoll:
@adamruzicka Thanks, it would be nice to get a new foreman-tasks_core (0.3.4?) release with that.

adamruzicka:
I'm aware of that, it is coming later today

PR #516, opened 2020-03-04
Fixes #28246 - Create bulk cancel api

MariaAga:
Adds a way to bulk cancel tasks using a query or an array of ids.
Return information about canceled tasks and failed to cancel tasks

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -89,19 +89,13 @@ module ForemanTasks
       param :search, String, :desc => N_('Resume tasks matching search string')
       param :task_ids, Array, :desc => N_('Resume specific tasks by ID')
       def bulk_resume
-        scope = resource_scope
-        scope = scope.search_for(params[:search]) if params[:search]
-        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
         if params[:search].nil? && params[:task_ids].nil?
-          scope = scope.where(:state => :paused)
-          scope = scope.where(:result => :error)
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
         end
-        scope = scope.where(:id => params[:task_ids]) if params[:task_ids]
-
         resumed = []
         failed = []
         skipped = []
-        scope.each do |task|
+        bulk_scope.each do |task|
           if task.resumable?
             begin
               ForemanTasks.dynflow.world.execute(task.execution_plan.id)
@@ -118,6 +112,24 @@ module ForemanTasks
           total: resumed.length + failed.length + skipped.length,
           resumed: resumed,
           failed: failed,
+          skipped: skipped
+        }
+      end
+
+      api :POST, '/tasks/bulk_cancel', N_('Cancel all cancellable tasks')
+      param :search, String, :desc => N_('Cancel tasks matching search string')
+      param :task_ids, Array, :desc => N_('Cancel specific tasks by ID')
+      def bulk_cancel
+        if params[:search].nil? && params[:task_ids].nil?
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+        end
+
+        cancelled, skipped = bulk_scope.partition(&:cancellable?)
+
+        cancelled.each(&:cancel)
+        render :json => {
+          total: cancelled.length + skipped.length,
+          cancelled: cancelled,
           skipped: skipped
         }
       end
@@ -272,7 +284,7 @@ module ForemanTasks
         case params[:action]
         when 'bulk_search', 'summary', 'details', 'sub_tasks'
           :view
-        when 'bulk_resume'
+        when 'bulk_resume', 'bulk_cancel'
           :edit
         else
           super
@@ -310,6 +322,14 @@ module ForemanTasks
           },
           results: results
         }
+      end
+
+      def bulk_scope
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        scope.where(:id => params[:task_ids]) if params[:task_ids]
+        scope
       end
     end
   end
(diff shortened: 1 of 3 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -122,6 +122,42 @@ def bulk_resume
...
 
+      api :POST, '/tasks/bulk_cancel', N_('Cancel all cancellable tasks')
+      param :search, String, :desc => N_('Cancel tasks matching search string')
+      param :task_ids, Array, :desc => N_('Cancel specific tasks by ID')
+      def bulk_cancel
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        if params[:search].nil? && params[:task_ids].nil?
+          scope = scope.where(:state => :scheduled)
```
MariaAga: @adamruzicka Is this correct? I'm not sure if there are other kind of tasks that can be canceled

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -122,6 +121,42 @@ def bulk_resume
...
+      api :POST, '/tasks/bulk_cancel', N_('Cancel all cancellable tasks')
+      param :search, String, :desc => N_('Cancel tasks matching search string')
+      param :task_ids, Array, :desc => N_('Cancel specific tasks by ID')
+      def bulk_cancel
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        if params[:search].nil? && params[:task_ids].nil?
+          raise ArgumentError, _('Please provide at least one of search or task_ids parameters in the request')
+        end
```
adamruzicka: I would move this to the top of the method

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -122,6 +121,42 @@ def bulk_resume
...
+            begin
+              task.cancel
+              cancelled << task_hash(task)
+            rescue RuntimeError
+              failed << task_hash(task)
+            end
+          else
+            skipped << task_hash(task)
+          end
+        end
```
adamruzicka: How about using `Enumerable#partition` to first split the scope into tasks which can be cancelled and which cannot and then operating only on those which can be cancelled?

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -122,6 +121,42 @@ def bulk_resume
...
+
+        cancelled = []
+        failed = []
+        skipped = []
+        scope.each do |task|
+          if task.cancellable?
+            begin
+              task.cancel
+              cancelled << task_hash(task)
+            rescue RuntimeError
```
adamruzicka: An exception should be raised only if the task's execution plan cannot be loaded. However all tasks which are `cancellable?` should have an execution plan. That would mean we can drop the `begin-rescue-end` block

app/models/foreman_tasks/task.rb
```diff
@@ -123,6 +123,10 @@ def pending?
     def resumable?
       false
     end
+    
+    def cancellable?
+      execution_plan&.cancellable?
+    end
```
adamruzicka: This is already defined on `ForemanTasks::Task::DynflowTask` and should not be needed.

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: Does this mean task canceling can't fail?

app/models/foreman_tasks/task.rb
MariaAga: Removed, thanks!

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: Looks much better that way, thanks!

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: Moved this and the second one up

app/controllers/foreman_tasks/api/tasks_controller.rb
adamruzicka: If a task is `cancellable?` then the cancelling should not raise, unless something happened between calls to `cancellable?` and `cancel`.

Under normal conditions, if a task is `cancellable?` then the worst thing that can happen is that the cancel won't actually do anything.

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -311,6 +323,13 @@ def tasks_list(filtered_scope)
           results: results
         }
       end
+
+      def bulk_scope
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        scope.where(:id => params[:task_ids]) if params[:task_ids]
```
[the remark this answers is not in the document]
MariaAga: added, thanks!

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -89,19 +89,13 @@ def bulk_search
       param :search, String, :desc => N_('Resume tasks matching search string')
       param :task_ids, Array, :desc => N_('Resume specific tasks by ID')
       def bulk_resume
-        scope = resource_scope
-        scope = scope.search_for(params[:search]) if params[:search]
-        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
         if params[:search].nil? && params[:task_ids].nil?
-          scope = scope.where(:state => :paused)
-          scope = scope.where(:result => :error)
+          raise ArgumentError, _('Please provide at least one of search or task_ids parameters in the request')
```
adamruzicka: See https://github.com/theforeman/foreman-tasks/pull/517/files#r394391514

MariaAga: Changed to `BadRequest`, Thanks!

adamruzicka:
Thank you @MariaAga !

PR #519, opened 2020-03-10
Fixes #29088 - add select all to tasks table

MariaAga:
The way it works:
click on the select box in the table header -> if there are more tasks then in the page a message comes out that asks if you want to select more than the page.
select all selects all the tasks that fit the search (including the dashboard filter)
If you un-select one item -> only the page items without the un-selected item are selected.
If you change the search (or the dashboard filter) it un-selects all tasks.
If you do an action and it doesn't work the selected tasks keep being selected
you can undo the all tasks selection in 2 ways:
1. unselect the header checkbox
2. click the undo button in the message


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -311,6 +323,13 @@ def tasks_list(filtered_scope)
           results: results
         }
       end
+
+      def bulk_scope
+        scope = resource_scope
+        scope = scope.search_for(params[:search]) if params[:search]
+        scope = scope.select('DISTINCT foreman_tasks_tasks.*')
+        scope.where(:id => params[:task_ids]) if params[:task_ids]
```
adamruzicka: If `params[:task_ids]` is `nil`, this method would return `nil`. Just add `scope` as the last line

adamruzicka:
I went to the UI, checked a task, clicked select all, clicked bulk resume and the request which hit the server looked like this `{"search"=>"(state_updated_at>2020-03-15T08:51:06.421Z )     ", "task"=>{}}`. Why would it match only tasks which were updated recently?

adamruzicka:
Could you please rebase so rubocop gets fixed?

MariaAga:
Will rebase after fix finish for long requests

MariaAga:
@pondrejk About the next page: in hosts it has the same behavior (And gmail :) ), so I think we should keep it. About the per page: in hosts in only selects the first page (lets say 20 hosts), and on change per page (10 per page) only the 2 first pages are selected which I think is confusing.

adamruzicka:
Needs a rebase

MariaAga:
Rebased and changed the toast notification at the end of the process to a regular notification

adamruzicka:
Thank you @MariaAga !

PR #523, opened 2020-03-26
Fixes #29415 - Ansible job failed when disabling batch triggering

hao-yu:
Use the new API to trigger remote task when batch triggering is off.

### DIFF
--- a/app/lib/actions/proxy_action.rb
+++ b/app/lib/actions/proxy_action.rb
@@ -67,7 +67,7 @@ module Actions
     def trigger_proxy_task
       suspend do |_suspended_action|
         remote_task = prepare_remote_task
-        remote_task.trigger(proxy_action_name, proxy_input)
+        ForemanTasks::RemoteTask.batch_trigger(remote_task.operation, [remote_task])
         output[:proxy_task_id] = remote_task.remote_task_id
       end
     end
--- a/test/controllers/api/tasks_controller_test.rb
+++ b/test/controllers/api/tasks_controller_test.rb
@@ -135,17 +135,20 @@ module ForemanTasks
           wait_for { ForemanTasks::Task.find_by(external_id: triggered.id).state == 'running' }
 
           task = ForemanTasks::Task.where(:external_id => triggered.id).first
+          action = task.execution_plan.actions.detect { |a| a.class == Support::DummyProxyAction }
+          remote_task = action.remote_task
           task.state.must_equal 'running'
           task.result.must_equal 'pending'
 
-          callback = Support::DummyProxyAction.proxy.log[:trigger_task].first[1][:callback]
+          input_hash = Support::DummyProxyAction.proxy.log[:trigger_task].first[1]
+          callback = input_hash[task.external_id][:action_input]['callback']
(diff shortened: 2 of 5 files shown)


hao-yu:
will try to fix the test tomorrow

adamruzicka:
Have you seen this one https://github.com/theforeman/foreman_ansible/pull/323 ?

hao-yu:
oh! i didn't see it. Then this is probably a duplicate then. Or maybe still worth to move it to the new API?

adamruzicka:
The only drawback with this is that if we used the new api this way, it would incur quite a bit of overhead on the smart proxy side. When you use batch triggering then for a batch of N hosts it triggers N+2 tasks on the smart proxy side, which is acceptable if N is around 100, but not as much if N is 1.

There is a way how we could use the new API to trigger only a single task on the smart proxy side, but it would require more changes (we would need to get a different operation name for this case).

So how about we go with this for now and open an issue to make the bigger changes later?

hao-yu:
@adamruzicka Thank you so much for the clarification! I am ok with your suggestion. I think in the real world, people probably won't set the N to a value that is unreasonably low like N = 1 unless for a reason. Just wondering is the old API going to be deprecated in the near coming future?

adamruzicka:
When we get this in, the old api shouldn't be used anywhere anymore. I'd be in favor of deprecating it and getting rid of it eventually.

hao-yu:
@adamruzicka If I understand correctly, this new batch triggering API can only bring benefit to batch_triggering enabled + no concurrency level set? Setting concurrency level or disabling batch trigger will create 3(1 + 2) tasks per call?

adamruzicka:
> this new batch triggering API can only bring benefit to batch_triggering enabled + no concurrency level set?

Yes, the point of the batch triggering API is to reduce the number of requests you have to do to trigger a certain amount of tasks. Once you bring concurrency control into the mix, you lose the benefits of the new api.

> Setting concurrency level or disabling batch trigger will create 3(1 + 2) tasks per call?

It will when you're using ansible-runner, unless we make additional changes. For SSH (or when using ansible-playbook) it would create 1+1 tasks

hao-yu:
@adamruzicka I think probably not worth to merge this patch since it is making the option to disable the batch triggering meaningless.

Just wondering what is the reason to call the batch trigger API here instead of calling the single trigger API?
https://github.com/theforeman/foreman-tasks/pull/489/files#diff-03dde90ec7e9caa322015dee15f4e956R77

hao-yu:
I guess it is because to workaround the incorrect "proxy_action_name" bug.

hao-yu:
I believe "https://github.com/theforeman/foreman_ansible/pull/323" is a better fix so I will close this request. 

@adamruzicka Any idea why https://github.com/theforeman/foreman_ansible/pull/323 hasn't merge yet?

adamruzicka:
> Just wondering what is the reason to call the batch trigger API here instead of calling the single trigger API?

If I recall we can't use the old API at that point, because if we get there, we've already discarded the information we could use to trigger it using the old api.

> @adamruzicka Any idea why theforeman/foreman_ansible#323 hasn't merge yet?

No idea, I'll see if we can get it moving

PR #524, opened 2020-03-27
Fixes #29423 - Add setting for increasing polling intervals

adamruzicka:


adamruzicka:
[test foreman-tasks]

mccun934:
I've tested this patch for a week in our internal lab with scale tests and it works perfectly. :+1:  from me

ezr-ondrej:
Code wise :+1: but I don't know if I like to have global setting, that override all the `Action` defaults. The actions could be short, they decided to have long intervals... why can't we make the intervals longer for the actions (tasks) that take long time?

adamruzicka:
We considered doing this with per-action settings, but we decided to have only one to not add another dozen or so settings.

I originally did this so the setting would be just appended at the end of the default array, but this gives more power to the user.

ezr-ondrej:
I'd actually like to add it at the end, as this gives user only option leave it up to the action, or override globaly. What does he do if he has problem with one action slow polling? he fine tune it to fit that action, but now all the others are slown down, because their defaults are overriden, right? :thinking:

ezr-ondrej:
BTW @adamruzicka what about constat, that would be used to multiply the action defaults? That seems to me like an elegant solution and it's even more understandable to users. They don't need to figure out what's gradual increas of polling interval to understand the value. They just say poll it 2x long interval or 0.5 shorter :thinking: WDYT?

adamruzicka:
The only downside with the multiplier is that if we allowed numbers lesser than 1, it would make the already too small 0.5 second interval at the beginning even smaller.

@mccun934 any opinions?

ezr-ondrej:
We could have an lower limit `[delay*const, 0.5].min` ?

mccun934:
+1 to a minimum delay setting, that should work well.

app/models/foreman_tasks/concerns/polling_action_extensions.rb
```diff
@@ -0,0 +1,12 @@
+module ForemanTasks
+  module Concerns
+    module PollingActionExtensions
+      def poll_intervals
+        multiplier = Setting[:foreman_tasks_polling_multiplier]
```
ezr-ondrej, line 5: wouln't this be good just to be on the safe side as we don't have type check for the setting if I read it correctly?
```suggestion
        multiplier = Setting[:foreman_tasks_polling_multiplier].to_f
        multiplier = 1.0 unless multiplier.positive?
```

app/models/setting/foreman_tasks.rb
```diff
@@ -13,7 +13,12 @@ def self.default_settings
           N_('Url pointing to the task troubleshooting documentation. '\
              'It should contain %{label} placeholder, that will be replaced with normalized task label '\
              '(restricted to only alphanumeric characters)). %{version} placeholder is also available.'),
-          nil)
+          nil),
+      set('foreman_tasks_polling_multiplier',
+          N_('Polling multiplier which is used to multiply the default polling intervals. '\
+             'This can be used to prevent polling too frequently for long running tasks.'),
+          1,
```
ezr-ondrej, line 20: maybe 1.0 to note user can use Float?

adamruzicka: I tried that, it gets truncated to an integer. Eseentialy it allows the user to input only non-negative integers.

app/models/foreman_tasks/concerns/polling_action_extensions.rb
adamruzicka: It automatically determines that the setting is an integer and typechecks it as that. We'll always get a non-negative integer here.

ezr-ondrej:
needs this: #532 for tests to pass.

adamruzicka:
Rebased on top of b1f4392

app/models/setting/foreman_tasks.rb
ezr-ondrej: well, I'm ok with that. shortening intervals was not subject of this PR :)

app/models/foreman_tasks/concerns/polling_action_extensions.rb
ezr-ondrej: in that case, ignore me :+1:

ezr-ondrej:
[test foreman-tasks] if it pulls newest core `develop`, could pass now

adamruzicka:
We're :green_apple:

mccun934:
Can we get this merged? we need this downstream

ezr-ondrej:
@adamruzicka can you move the file to `lib/concerns`? Or if you decide to keep it, it's fine by me as well :)

ezr-ondrej:
Test failure unrelated :+1:

adamruzicka:
Great, thank you for reviewing this

mccun934:
:clap:

PR #525, opened 2020-04-02
Fixes #29457 - hide task actions for user with no permission

MariaAga:
In tasks details and table user without can_edit permission still sees the cancel/resume buttons. if he clicks them there is a back end error as he has no permissions

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -315,6 +315,10 @@ module ForemanTasks
         @resource_scope ||= ForemanTasks::Task.authorized("#{action_permission}_foreman_tasks")
       end
 
+      def controller_permission
+        'foreman_tasks'
+      end
+
       def action_permission
         case params[:action]
         when 'bulk_search', 'summary', 'details', 'sub_tasks'
--- a/app/views/foreman_tasks/api/tasks/show.json.rabl
+++ b/app/views/foreman_tasks/api/tasks/show.json.rabl
@@ -1,5 +1,7 @@
 object @task if @task
 
+extends 'api/v2/layouts/permissions'
+
 attributes :id, :label, :pending, :action
 attributes :username, :started_at, :ended_at, :state, :result, :progress
 attributes :input, :output, :humanized, :cli_example, :start_at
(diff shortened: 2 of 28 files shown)


MariaAga:
After talking with the UE team, it's better to disable the button with explanation tooltips

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -101,13 +101,14 @@ const TasksTable = ({
...
-          uriQuery.sort_order,
+        columns={createTasksTableSchema({
+          setSort: setSortHistory,
+          by: uriQuery.sort_by,
+          order: uriQuery.sort_order,
           taskActions,
-          getSelectionController()
-        )}
+          selectionController: getSelectionController(),
+          canEdit: results[0].canEdit,
```
adamruzicka, line 110: Does this mean that if the user cannot edit first task in the response, then they won't be able to edit any other task on the same page?

MariaAga: This is used for the bulk select all.
Isn't the `edit_foreman_tasks` permission is user based and not task based?

MariaAga:
@pondrejk nice catch! fixed it (added `controller_permission` to the controller)

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
adamruzicka: Foreman's permission system allows to "scope" the permissions. That means that even though a user has `edit_foreman_tasks`, they may not be able to edit *any* task.

MariaAga: I think we should still leave this and notify that some of the selected tasks can't be canceled because the user doesn't have permissions

MariaAga: Added Refs #29457 - add canEdit check to bulk actions

adamruzicka:
Needs a rebase

MariaAga:
Rebased with all the changes made (removed setting, added force cancel)

adamruzicka:
Needs a rebase

MariaAga:
rebased

adamruzicka:
One more rebase please

MariaAga:
Rebased, had to use `data-original-title` and `title` for buttons as it was giving the wrong tooltip without it

adamruzicka:
Thank you @MariaAga !

PR #526, opened 2020-04-02
Fixes #29462 - bulk tasks api doesnt work with tasks ids

MariaAga:
it doesn't filter out from the tasks list only the tasks with the relevant ids from params


adamruzicka:
Thank you @MariaAga !

PR #530, opened 2020-04-26
Fixes #29623 - export tasks doesn't filter subtasks

MariaAga:
In tasks table when in subtasks mode the export button will export all tasks instead of only the subtasks


adamruzicka:
Thank you @MariaAga !

PR #531, opened 2020-04-26
Fixes #29597 - refactor task actions

MariaAga:
Tasks actions parts - task actions, buttons, modals.
https://github.com/theforeman/foreman-tasks/pull/537 - All parts that can be used in details and table should be in one place.
https://github.com/theforeman/foreman-tasks/pull/536/ - Cancel and Resume buttons are replaced with the use of Foremans ActionButtons that gets one or many buttons and converts it to a single button or a drop down.
https://github.com/theforeman/foreman-tasks/pull/538 The Confirm modal can be one dynamic modal instead of 4 modals.


MariaAga:
Adding tests!

adamruzicka:
Needs a rebase

MariaAga:
Split this PR into 3 smaller ones

PR #534, opened 2020-05-04
Add sqlite as runtime dependency for tests

adamruzicka:


foreman-tasks-core.gemspec
```diff
@@ -19,4 +19,5 @@ Common code used both at Forman and Foreman proxy regarding tasks
             ['lib/foreman_tasks_core.rb', 'LICENSE']
 
   s.add_dependency "dynflow", '>= 1.2.0'
+  s.add_dependency "sqlite3"
```
tbrisker, line 22: Shouldn't this also be dev dependency? Since otherwise tasks will pull it in as it depends on core

adamruzicka: This approach is bad.

We need to pull in sqlite, but only for the tests. Setting it as dev dependency doesn't pull it in. So either we set it as a dependency and then we'll get it always, which is wrong but fixes the tests. Or we set it as a dev dependency and then we won't get it, it won't fix the tests but it would be the correct solution in principle if it worked.

My understanding is in this case, Foreman would have to provide the dev dependency for it to be actually used. Could we maybe bring back sqlite in a test group or something?

An alternative would be to rewrite the tests completely, but that wouldn't be a five minute fix

Is there any other option I'm missing?

tbrisker: Another options is perhaps we could install the sqlite gem manually as part of the test pipelines? 
maybe @ekohl has some other idea?

ekohl: There is `gemfile.d` which is read in tests by Jenkins. That's the equivalent of a `Gemfile`. https://github.com/theforeman/foreman_bootdisk/pull/90/files#diff-58ea1c31745b5ab09c79ba3d6f5529bb uses this too. Would that suffice?

adamruzicka: That might do the trick, I'll check it out

tbrisker: doesn't seem to work, souldn't it be `bundler.d/`?

tbrisker: or rather, the file name seems wrong:
```
+ '[' -e /home/jenkins/workspace/foreman-tasks-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/gemfile.d/foreman-tasks.rb ']'

tbrisker:
Test failures are transient.

PR #536, opened 2020-05-07
Refs #29597 - refactor task actions buttons

MariaAga:
Cancel and Resume buttons are replaced with the use of Foremans ActionButtons that gets one or many buttons and converts it to a single button or a drop down.
Added clicking tests for Task buttons.


webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
```diff
@@ -1,27 +1,38 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { CancelButton } from './CancelButton';
-import { ResumeButton } from './ResumeButton';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { ActionButtons } from 'foremanReact/components/common/ActionButtons/ActionButtons';
 
 export const ActionButton = ({ id, name, availableActions, taskActions }) => {
   const isResume = availableActions.resumable;
+  const isCancel = availableActions.cancellable;
```
laviro, line 8: instead of re-declaring these boolean variables,
you can destruct `availableActions` at line 6:
```js
export const ActionButton = ({ id, name, availableActions: { resumable, cancellable }, taskActions }) => {
```

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
```diff
@@ -1,27 +1,38 @@
...
+    });
   }
   return (
-    <CancelButton
-      id={id}
-      name={name}
-      disabled={!availableActions.cancellable}
-      onClick={taskActions.cancelTask}
-    />
+    <div title={title}>
```
laviro, line 33: should we use `OverlayTrigger` here instead?

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
```diff
@@ -1,27 +1,38 @@
...
+import { translate as __ } from 'foremanReact/common/I18n';
+import { ActionButtons } from 'foremanReact/components/common/ActionButtons/ActionButtons';
 
 export const ActionButton = ({ id, name, availableActions, taskActions }) => {
   const isResume = availableActions.resumable;
+  const isCancel = availableActions.cancellable;
+  const buttons = [];
+  let title;
+  if (!isResume && !isCancel) {
+    title = __('Task cannot be canceled');
```
laviro, line 12: looks like `title` is being set only here,
so I assume we could use a `const` instead

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
MariaAga: I don't think we should as it's here only to provide extra basic information

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
MariaAga: Can't because then title will be considered undefined by eslint (as its only defined inside the if block)

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
MariaAga: Changed, thanks!

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
laviro: ok, should we use a `span` instead?

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
laviro: you can do something like:
```js
const title = !isResume && !isCancel ? __('Task cannot be canceled') : null
```

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
MariaAga: I'm not sure what you mean, instead of what?

laviro: oh, I mean instead of the `div`, since it's an inline feature from what I understand,
if it's being rendered inside a table cell.

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
MariaAga: Changed, thanks

webpack/ForemanTasks/Components/common/ActionButtons/ActionButton.js
MariaAga: Yeah that makes sense, thanks!

adamruzicka:
Test failures are indeed unrelated, merging

adamruzicka:

PR #537, opened 2020-05-07
Refs #29597 - refactor task actions folders

MariaAga:
Task actions that can be used in the table page and details page should be in one place: taskActions.
depends on https://github.com/theforeman/foreman-tasks/pull/536/

PR #538, opened 2020-05-07
Refs #29597 - refactor ConfirmationModal for tasks

MariaAga:
The Confirm modal can be one dynamic modal instead of 4 modals.
Depends on https://github.com/theforeman/foreman-tasks/pull/537 and https://github.com/theforeman/foreman-tasks/pull/536


webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -92,19 +93,19 @@ const TasksTable = ({
     updateURlQuery({ sort_by: by, sort_order: order }, history);
   };
 
-  const taskActions = {
+  const modalActions = {
```
laviro, line 96: why do we need the rename of taskActions all over the PR actually?
I think it make more sense to have taskActions which we insert into a specific modal.
because we can have other modals where this actions doesn't really fit.

MariaAga: I started to get confused with the other var named tasksaction from here:
https://github.com/theforeman/foreman-tasks/blob/3faffd08f7c8f5b36f92f574c6283cb6c12373a0/webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js#L65
As modalActions had actions that open modals and tasks actions had just task actions

laviro: yes, it is quite confusing.. since also `tasksActions` contains modal functions.
I would suggest to pass each individual prop to the `ConfirmationModals`, e.g:
```js
<ConfirmationModals
       cancelTask={cancelTask}
       resumeTask={resumeTask}
       cancelSelectedTasks={cancelSelectedTasks}
       resumeSelectedTasks={resumeSelectedTasks}
/>
```

MariaAga: When the force cancel will be added I think it's might be too much (6 actions total), and the `resumeTask` will be something like `resumeTaskWithID` which looks to me not that organized

laviro: I don't see problem with having many props, I prefer that instead of passing a huge object.

btw, we can always create more generic actions which will handle regular or bulk function, based on the amount of IDs that was provided, right?

MariaAga: I think i generalized them as much as I could, do you have a suggestion to make them more general?

MariaAga: Also the consts in `tasksActions` can be used comfortably when choosing  which function to use for which modal, and moving it to props will cause an (in my opinion) unnecessary switch case block

laviro: We could investigate about it more in another PR though meanwhile I think passing each func as a prop would be great.

MariaAga: Yeah but instead of passing the object there would have to be a convertion to 4 consts to pass as props, and then a switch case to choose which prop to use:
```JS 
TasksTablePage.js
const cancelTask={
...
};
const resumeTask={
...
};
const cancelSelectedTasks={
...
};
const resumeSelectedTasks={
...
};

CofirmationModals.js
const
const getFunc = type =>{
  switch (type) {
    case CANCEL_MODAL:
      cancelTask;
    case RESUME_MODAL:
      resumeTask;
    case CANCEL_SELECTED_MODAL:
      cancelSelectedTasks;
    case RESUME_SELECTED_MODAL:
      resumeSelectedTasks;
  }  
}
```
Why not to just pass an object instead?

laviro: lets continue the discussion to the right line,
as for this specific line in the TaskTable I guess it could be renamed to `modalActions` as you did, because it uses `createTasksTableSchema`

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmationModals.js
```diff
@@ -1,49 +1,44 @@
...
-import { ResumeConfirm } from './ResumeConfirm';
+import { ConfirmModal } from './ConfirmModal';
+
+const modalProps = type => {
+  if ([CANCEL_MODAL, CANCEL_SELECTED_MODAL].includes(type)) {
+    return { actionText: 'cancel', actionState: 'stopped' };
+  } else if ([RESUME_MODAL, RESUME_SELECTED_MODAL].includes(type))
+    return { actionText: 'resume', actionState: 'running' };
+  return { actionText: '', actionState: '' };
+};
```
laviro, line 18: continue the discussion from https://github.com/theforeman/foreman-tasks/pull/538#discussion_r423713827
can we pass the right props from the parent component without adding extra login here?

MariaAga: Added in a new commit `Refs #29597 - move modalProps to tasksTablePage`

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmationModals.js
```diff
@@ -1,55 +1,45 @@
...
-      action={tasksActions.cancelSelectedTasks}
-      selectedRowsLen={selectedRowsLen}
-      id={CANCEL_SELECTED_CONFIRM_MODAL_ID}
-    />
-    <ResumeConfirm
-      closeModal={modalProps.resumeSelectedModal.setModalClosed}
-      action={tasksActions.resumeSelectedTasks}
-      selectedRowsLen={selectedRowsLen}
-      id={RESUME_SELECTED_CONFIRM_MODAL_ID}
+    <ConfirmModal
```
laviro, line 17: since we are just rendering `ConfirmModal` in `ConfirmationModals`,
can we avoid this file and render `ConfirmModal` in the parent component?

MariaAga: I'll think it's better to make a new file to split some of the logics, also this file will have more logic in the next PR so it's going to be needed for that.

laviro: I think the best approach would be to have all of the logic in the parent component,
and than to have very simple component which mostly handle the UI elements rendering,

we also won't need extra wrappers, and components with similar names such as `ConfirmationModals` and `ConfirmModal`, and the props that we pass will be more simple to understand.

MariaAga: Yeah that makes sense, changed

webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
```diff
@@ -71,50 +63,74 @@ const TasksTablePage = ({
...
   return (
     <div className="tasks-table-wrapper">
-      <ConfirmationModals
-        tasksActions={tasksActions}
+      <ConfirmModal
+        closeModal={setModalClosed}
+        {...tasksActions[modalID]}
         selectedRowsLen={
-          props.allRowsSelected ? props.itemCount : props.selectedRows.length
+          [CANCEL_MODAL, RESUME_MODAL].includes(modalID) ? 1 : selectedRowsLen
```
laviro, line 131: Can this logic be moved to line 122 where `const selectedRowsLen` is declared?

webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
```diff
@@ -71,50 +63,74 @@ const TasksTablePage = ({
...
+      action: () => {
+        resumeTask({
+          taskId: clicked.taskId,
+          taskName: clicked.taskName,
+          url,
+          parentTaskID,
+        });
+      },
+      actionText: 'resume',
+      actionState: 'running',
```
laviro, line 115: it feels like something that should be part of redux,
for example when this action is triggered, these values should be updated in the store, and than consumed by the modal.

MariaAga: What do you about making the confirm modal connected to redux? (with those fields and more)

laviro: yeah I think it's a great idea, this way the `TasksTablePage` will become much simpler,
if you could move those action to be consumed by the modal it will also be awesome

MariaAga: I think the actions depend on too much information from the page so I only moved the `UPDATE_MODAL` to the `ConfirmModal` new reducer. during the update the actionText and actionState get updated as well

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,71 @@
...
+  tasksActions,
+  actionType,
+  selectedRowsLen,
+  id,
+}) => {
+  const { setModalClosed } = useForemanModal({
+    id,
+  });
+  if ([CANCEL_MODAL, RESUME_MODAL].includes(actionType)) {
+    selectedRowsLen = 1;
```
laviro, line 22: it's not a good practice to mutate the props,
can this logic be part of [TasksTablePage selectedRowsLen const](https://github.com/theforeman/foreman-tasks/pull/538/files#diff-8de9b4e575996966ab1cae4c0ed76713R107)

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModalReducer.js
```diff
@@ -0,0 +1,36 @@
...
+          return state
+            .set('actionText', 'cancel')
+            .set('actionState', 'stopped')
+            .set('actionType', payload.modalID);
+        case RESUME_SELECTED_MODAL:
+        case RESUME_MODAL:
+          return state
+            .set('actionText', 'resume')
+            .set('actionState', 'running')
+            .set('actionType', payload.modalID);
```
laviro, line 28: can you use `merge` in both cases?

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,71 @@
...
+        { action: actionText, number: selectedRowsLen, state: actionState }
+      )}
+      <ForemanModal.Footer>
+        <Button onClick={setModalClosed}>{__('No')}</Button>
+        <Button
+          bsStyle="primary"
+          onClick={() => {
+            tasksActions[actionType]();
+            setModalClosed();
+          }}
```
laviro, line 42: can you move this logic too into the parent `TasksTablePage`?
this way it will be more generic `onClick={onClick}`

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <ForemanModal
+      title={sprintf(__('%s Selected Tasks'), startCase(actionText))}
+      id={id}
+    >
+      {sprintf(
+        __(
+          `This will %(action)s %(number)s task(s), putting them in the %(state)s state. Are you sure?`
+        ),
+        { action: actionText, number: selectedRowsLen, state: actionState }
```
laviro, line 33: this two `sprintf` can be moved to the parent too,
and consumed here by `title, body` props for example.

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
MariaAga: The issue is that tasksTablePage doesn't have the `actionType`. But I'll change this logic in confirmModal

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
MariaAga: I think it's better to add it here because the action should always close the modal and adding it into the parent will mean adding it 4 times

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
MariaAga: Why? this is not intended to be a generic component, but a component for the tasks table to divide the logic in the tasksTablePage

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModalReducer.js
MariaAga: Yeah! changed

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
laviro: I am talking more about the `tasksActions[actionType]`,
yes, the setModalClosed can stay here.

MariaAga: The parent doesn't have the actionType so I'll leave it here

MariaAga:
@laviro Moved tasks actions in a the new commit, thanks for your help in these reviews!

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,135 @@
...
+      allRowsSelected
+        ? bulkCancelBySearch({ query: uriQuery, parentTaskID })
+        : bulkCancelById({
+            selected: getSelectedTasks(),
+            url,
+            parentTaskID,
+          }),
+    [CANCEL_MODAL]: () =>
+      cancelTask({
+        taskId: clicked.taskId,
```
laviro, line 46: can you move this action into a `ConfirmModalActions` file,
for example it can be implemented like this: https://pastebin.com/3zR6j9qW
and than import it in the `index.js` file

webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
```diff
@@ -50,71 +50,24 @@ const TasksTablePage = ({
     }));
```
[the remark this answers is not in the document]
MariaAga: It's a function thats only called when needed in the modal, instead of always sending the array of selected objects. are you sure this should be moved?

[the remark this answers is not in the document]
MariaAga: I see it can be useful for the confirm actions, nvm

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
MariaAga: Moved, thanks!

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,135 @@
...
+        url,
+        parentTaskID,
+      }),
+  };
+
+  const { setModalClosed } = useForemanModal({
+    id,
+  });
+  if ([CANCEL_MODAL, RESUME_MODAL].includes(actionType)) {
+    selectedRowsLen = 1;
```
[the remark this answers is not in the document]
MariaAga: Moved it to be a part of the selectors

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const ConfirmModal = ({
+  actionText,
+  actionState,
+  actionType,
+  selectedRowsLen,
+  id,
+  parentTaskID,
+  url,
+  ...tasksActions
```
laviro, line 18: can you rename it to `...props`

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,69 @@
...
+        ),
+        { action: actionText, number: selectedRowsLen, state: actionState }
+      )}
+      <ForemanModal.Footer>
+        <Button onClick={setModalClosed}>{__('No')}</Button>
+        <Button
+          className="confirm-button"
+          bsStyle="primary"
+          onClick={() => {
+            tasksActions[actionType]({ url, query, parentTaskID });
```
laviro, line 41: it can be `props[actionType]()` instead

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,69 @@
...
+  actionText,
+  actionState,
+  actionType,
+  selectedRowsLen,
+  id,
+  parentTaskID,
+  url,
+  ...tasksActions
+}) => {
+  const query = getURIQuery(url);
```
laviro, line 20: can you use the `uriQuery` from the parent?

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,69 @@
...
+  const { setModalClosed } = useForemanModal({
+    id,
+  });
```
laviro, line 23: the parent already uses the `useForemanModal` hook for an id which equals also to CONFIRM_MODAL,
can you pass it from there?

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModalActions.js
```diff
@@ -0,0 +1,71 @@
...
+      bulkCancelById({
+        selected: selectSelectedTasks(getState()),
+        url,
+        parentTaskID,
+      })
+    );
+  },
+  [CANCEL_MODAL]: ({ url, parentTaskID }) => (dispatch, getState) => {
+    const { taskId, taskName } = selectClicked(getState());
+    dispatch(
```
laviro, line 35: can you add a `return` here ?

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
MariaAga: renamed

MariaAga:
Thanks for the review @laviro, updated the last commit with the needed changes

MariaAga:
*Added tests for the actions

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -0,0 +1,66 @@
...
+      ),
+      { action: actionText, number: selectedRowsLen, state: actionState }
+    )}
+    <ForemanModal.Footer>
+      <Button onClick={setModalClosed}>{__('No')}</Button>
+      <Button
+        className="confirm-button"
+        bsStyle="primary"
+        onClick={() => {
+          props[actionType]({ url, query: uriQuery, parentTaskID });
```
laviro, line 36: nit: can you rename `uriQuery` to `query` in the props that are passed from the parent?

MariaAga: Sure, renamed

adamruzicka:

PR #540, opened 2020-05-11
Fixes #29495 Filter trailing forward slash from task URL

stejskalleos:


webpack/ForemanTasks/Components/TaskDetails/TasksDetailsHelper.js
```diff
@@ -1 +1 @@
-export const getTaskID = () => window.location.pathname.split('/').slice(-1)[0];
+export const getTaskID = () => window.location.pathname.split('/').filter(i => i).slice(-1)[0];
```
adamruzicka, line 1: Could you add a comment why it is there? Maybe it would be obvious to js folks, but it took me a while to realize why filtering by identity would help

adamruzicka:
Failures in travis are most likely related

adamruzicka:
@MariaAga could you take a look?

adamruzicka:
Thank you @stejskalleos & @MariaAga !

PR #539, opened 2020-05-10
Refs #29597 - task unlock use redux instead of href

MariaAga:
Task unlock and force unlock should use redux with post requests instead of redirecting to the unlock page with href
Blocked by https://github.com/theforeman/foreman-tasks/pull/533

PR #541, opened 2020-05-11
Fixes #29729 - add force unlock to tasks table rows

MariaAga:
Each task - row should have a force cancel button (if the task is not already stopped)
Does not include unlock for multiple tasks
depends on https://github.com/theforeman/foreman-tasks/pull/539

### DIFF
--- a/webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
+++ b/webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { translate as __ } from 'foremanReact/common/I18n';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
 import { ClickConfirmation } from '../common/ClickConfirmation';
 import { UNLOCK_MODAL, FORCE_UNLOCK_MODAL } from './TaskActionsConstants';
 
@@ -22,12 +22,15 @@ export const UnlockModal = ({ onClick, id }) => (
   />
 );
 
-export const ForceUnlockModal = ({ onClick, id }) => (
+export const ForceUnlockModal = ({ onClick, id, selectedRowsLen }) => (
   <ClickConfirmation
     id={id}
     title={__('Force Unlock')}
-    body={__(
-      'Resources will be unlocked and will not prevent other tasks from being run. As the task might be still running, it should be avoided to use this unless you are really sure the task got stuck'
+    body={sprintf(
+      __(
+        `Resources for %s task(s) will be unlocked and will not prevent other tasks from being run. As the task might be still running, it should be avoided to use this unless you are really sure the task got stuck.`
+      ),
+      selectedRowsLen
     )}
     confirmationMessage={confirmationMessage}
     confirmAction={__('Force Unlock')}
@@ -44,6 +47,7 @@ UnlockModal.propTypes = {
 ForceUnlockModal.propTypes = {
   onClick: PropTypes.func.isRequired,
   id: PropTypes.string,
+  selectedRowsLen: PropTypes.number,
 };
 
 UnlockModal.defaultProps = {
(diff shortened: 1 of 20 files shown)


adamruzicka:
Needs a rebase

adamruzicka:
Needs a rebase

webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
```diff
@@ -22,12 +22,12 @@ export const UnlockModal = ({ onClick, id }) => (
...
 );
 
-export const ForceUnlockModal = ({ onClick, id }) => (
+export const ForceUnlockModal = ({ onClick, id, selectedRowsLen }) => (
   <ClickConfirmation
     id={id}
     title={__('Force Unlock')}
     body={__(
-      'Resources will be unlocked and will not prevent other tasks from being run. As the task might be still running, it should be avoided to use this unless you are really sure the task got stuck'
+      `Resources for ${selectedRowsLen} task(s) will be unlocked and will not prevent other tasks from being run. As the task might be still running, it should be avoided to use this unless you are really sure the task got stuck.`
```
sharvit, line 30: I don't think the translation work that way, see: http://theforeman.github.io/foreman/?path=/docs/introduction-internationalization--page

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
```diff
@@ -16,32 +18,46 @@ export const ConfirmModal = ({
...
-      ),
-      { action: actionText, number: selectedRowsLen, state: actionState }
-    )}
-    <ForemanModal.Footer>
-      <Button onClick={setModalClosed}>{__('No')}</Button>
-      <Button
-        className="confirm-button"
-        bsStyle="primary"
+}) => {
+  if (actionType === FORCE_UNLOCK_MODAL) {
```
sharvit, line 22: It doesn't feel right, why not using the `ForceUnlockModal` from the parent instead?

MariaAga: Can you explain what do you mean to use it from the parent?

webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
MariaAga: You're right, thanks!

MariaAga:
Which part is confusing? Maybe I can add/remove something to make it less confusing

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
sharvit: I guess there is a parent component that rendering `<ConfirmModal ... />` to render `ForemanModal` and `<ConfirmModal ... actionType={FORCE_UNLOCK_MODAL} />` to render `ForceUnlockModal`.

Why not just rendering `ForceUnlockModal` instead `<ConfirmModal ... actionType={FORCE_UNLOCK_MODAL} />`?

sharvit:
> @sharvit Which part is confusing? Maybe I can add/remove something to make it less confusing

I see it confusing having switch inside another switch but It's not part of this PR right?

MariaAga:
@sharvit It was made to split the logic so it won't all be in the `TasksTablePage`, it was part of the previous PR which was merged: https://github.com/theforeman/foreman-tasks/pull/538

webpack/ForemanTasks/Components/TasksTable/Components/ConfirmModal/ConfirmModal.js
MariaAga: The parent component doesn't have the info about actionType

sharvit:
> @sharvit It was made to split the logic so it won't all be in the `TasksTablePage`, it was part of the previous PR which was merged: #538

I see, would it make sense just moving the inner switch to a separate function to increase readability?

MariaAga:
@sharvit  I don't think it would make sense as the whole reducer is just those 2 switches with nothing else

adamruzicka:
Thank you @MariaAga & @sharvit !

PR #542, opened 2020-05-11
i18n - pulling from tx

upadhyeammit:


upadhyeammit:
@adamruzicka request to review, thank you !

adamruzicka:
This seems to deal only with translations from ruby files, but not ones from react components. Is that expected?

upadhyeammit:
> This seems to deal only with translations from ruby files, but not ones from react components. Is that expected?

Nope, we need to have translations for both, but I dont know procedure for react components translations, if you aware about that maybe you can push the translations or share the process and I will try that.

adamruzicka:
The procedure is the same, but https://github.com/theforeman/foreman/pull/7649 needs to be merged first

adamruzicka:
The PR in foreman core is now in

adamruzicka:
Thank you @upadhyeammit !

PR #543, opened 2020-05-12
Fixes #29731 - Add bulk force unlock api for tasks

MariaAga:
Can bulk cancel tasks using search or tasks ids using the api `/bulk_stop/`
Response is the stopped tasks and the number of tasks in the request that were already stopped.
User also gets a notification when its done.

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -141,6 +141,34 @@ module ForemanTasks
         }
       end
 
+      api :POST, '/tasks/bulk_stop', N_('Stop all stoppable tasks')
+      param :search, String, :desc => N_('Stop tasks matching search string')
+      param :task_ids, Array, :desc => N_('Stop specific tasks by ID')
+      def bulk_stop
+        if params[:search].nil? && params[:task_ids].nil?
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+        end
+
+        filtered_scope = bulk_scope
+        total_length = filtered_scope.count
+        to_stop = filtered_scope.where.not(state: :stopped)
+        to_stop_length = to_stop.count
+        skipped_length = total_length - to_stop_length
+
+        to_stop.find_each { |task| task.update(state: :stopped) }
+
+        if params[:search]
+          notification = UINotifications::Tasks::TaskBulkStop.new(filtered_scope.first, to_stop_length, skipped_length)
+          notification.deliver!
+        end
+
+        render :json => {
+          total: total_length,
+          stopped_length: to_stop_length,
+          skipped_length: skipped_length
+        }
+      end
+
       api :GET, '/tasks', N_('List tasks')
       param :search, String, :desc => N_('Search string')
       param :page, :number, :desc => N_('Page number, starting at 1')
@@ -291,7 +319,7 @@ module ForemanTasks
         case params[:action]
         when 'bulk_search', 'summary', 'details', 'sub_tasks'
           :view
-        when 'bulk_resume', 'bulk_cancel'
+        when 'bulk_resume', 'bulk_cancel', 'bulk_stop'
           :edit
         else
           super
--- a/app/services/ui_notifications/tasks/task_bulk_stop.rb
+++ b/app/services/ui_notifications/tasks/task_bulk_stop.rb
@@ -1,0 +1,36 @@
+module UINotifications
+  module Tasks
+    class TaskBulkStop < ::UINotifications::Base
+      def initialize(task, stopped_length, skipped_length)
+        @subject = task
+        @stopped_length = stopped_length
+        @skipped_length = skipped_length
+      end
+
+      def create
+        Notification.create!(
+          initiator: initiator,
+          audience: audience,
+          subject: subject,
+          notification_blueprint: blueprint,
+          message: message,
+          notification_recipients: [NotificationRecipient.create(:user => User.current)]
+        )
+      end
+
+      def audience
+        Notification::AUDIENCE_GLOBAL
+      end
+
+      def message
+        ('%{stopped} Tasks were stopped. %{skipped} Tasks were already stopped. ' %
+          { stopped: @stopped_length,
+          skipped: @skipped_length })
+      end
+
+      def blueprint
+        @blueprint ||= NotificationBlueprint.find_by(name: 'tasks_bulk_stop')
+      end
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -56,6 +56,7 @@ Foreman::Application.routes.draw do
           post :bulk_search
           post :bulk_resume
           post :bulk_cancel
+          post :bulk_stop
           get :summary
           get '/summary/:id/sub_tasks/', action: 'summary_sub_tasks'
           post :callback
--- a/db/seeds.d/30-notification_blueprints.rb
+++ b/db/seeds.d/30-notification_blueprints.rb
@@ -41,6 +41,13 @@ blueprints = [
     name: 'tasks_bulk_cancel',
     level: 'info',
     message: "DYNAMIC",
+  },
+
+  {
+    group: N_('Tasks'),
+    name: 'tasks_bulk_stop',
+    level: 'info',
+    message: "DYNAMIC",
   }
 ]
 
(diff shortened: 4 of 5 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,30 @@ def bulk_cancel
...
+      param :search, String, :desc => N_('Stop tasks matching search string')
+      param :task_ids, Array, :desc => N_('Stop specific tasks by ID')
+      def bulk_stop
+        if params[:search].nil? && params[:task_ids].nil?
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+        end
+
+        filtered_scope = bulk_scope
+        to_stop = filtered_scope.where.not(state: :stopped)
+        already_stopped_length = filtered_scope.length - to_stop.length
```
adamruzicka, line 154: Using `length` would load all the records into memory and then count them. On the other hand, using `count` would make the database count the records and return their number, which is more efficient.

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,30 @@ def bulk_cancel
...
+        to_stop.update_all(state: :stopped)
+
+        if params[:search]
+          notification = UINotifications::Tasks::TaskBulkStop.new(filtered_scope.first, to_stop.length, already_stopped_length)
+          notification.deliver!
+        end
+        render :json => {
+          total: filtered_scope.length,
```
adamruzicka, line 162: It would be nice if you could get the total count earlier in the code and then just reuse that, instead of counting the tasks in several places

MariaAga: Changed, thanks

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: using now length for `to_stop` as its loaded anyway and count for `filtered_scope`

MariaAga:
@adamruzicka Can you take another look?

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,31 @@ def bulk_cancel
         }
       end
 
+      api :POST, '/tasks/bulk_stop', N_('Stop all stoppable tasks')
+      param :search, String, :desc => N_('Stop tasks matching search string')
+      param :task_ids, Array, :desc => N_('Stop specific tasks by ID')
```
adamruzicka, line 146: Probably out of scope of this PR, but how about we drop the `task_ids` param everywhere. If someone wants to target tasks by id, they would use search with something like `id ^ (1, 2, 3)`?

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,31 @@ def bulk_cancel
...
+      def bulk_stop
+        if params[:search].nil? && params[:task_ids].nil?
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+        end
+
+        filtered_scope = bulk_scope
+        to_stop = filtered_scope.where.not(state: :stopped)
+        total_length = filtered_scope.count
+        to_stop_length = to_stop.length
+        to_stop.update_all(state: :stopped)
```
adamruzicka, line 156: `update_all` does not instantiate the objects and does not trigger callbacks. Because of this, the tasks wouldn't have their `state_updated_at` updated properly.

adamruzicka, line 156: `update_all` returns the number of records it updated, so you can merge those two lines into one

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: Yeah that makes sense, added: https://projects.theforeman.org/issues/29913

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: Should I add ` state_updated_at: DateTime.now`?

adamruzicka: It would be best to let this trigger the callbacks, instead of specifying the date explicitly here. Using `update` on a collection or `update_attribute` on an instance might be the way to go

MariaAga: changed to update as is uses validations as well

MariaAga: changed to arrays so using length

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,33 @@ def bulk_cancel
...
+      param :search, String, :desc => N_('Stop tasks matching search string')
+      param :task_ids, Array, :desc => N_('Stop specific tasks by ID')
+      def bulk_stop
+        if params[:search].nil? && params[:task_ids].nil?
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+        end
+
+        filtered_scope = bulk_scope
+        skipped, to_stop = filtered_scope.partition { |task| task.state == "stopped" }
+        to_stop.each { |task| task.update({ state: :stopped }) }
```
adamruzicka, line 154: You should be able to drop the curly braces around `state: :stopped`

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,33 @@ def bulk_cancel
...
+      api :POST, '/tasks/bulk_stop', N_('Stop all stoppable tasks')
+      param :search, String, :desc => N_('Stop tasks matching search string')
+      param :task_ids, Array, :desc => N_('Stop specific tasks by ID')
+      def bulk_stop
+        if params[:search].nil? && params[:task_ids].nil?
+          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+        end
+
+        filtered_scope = bulk_scope
+        skipped, to_stop = filtered_scope.partition { |task| task.state == "stopped" }
```
adamruzicka, line 153: I know we did it this way bulk cancel, but the situation there was a bit different. In here we can tell if a task can be "force cancelled" or not before we load it. That means we can be more efficient and load only the records we will actually use and just count the others.

app/controllers/foreman_tasks/api/tasks_controller.rb
adamruzicka, line 154: Since we're actually loading the records, we should use something like [find_each](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) so we don't eat too much memory if there's a lot of tasks to be stopped.

MariaAga: So should we only send a number instead of all the records at the end?

adamruzicka: Yes, unless there's a reason why we would want all the details that I'm not seeing right now

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: Changed to 'where', thanks

app/controllers/foreman_tasks/api/tasks_controller.rb
MariaAga: removed, thanks

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -141,6 +141,36 @@ def bulk_cancel
...
+
+        filtered_scope = bulk_scope
+        total_length = filtered_scope.count
+        to_stop = filtered_scope.where.not(state: :stopped)
+        to_stop_length = to_stop.count
+        skipped_length = total_length - to_stop_length
+
+        to_stop.find_each do |task|
+          task.update(state: :stopped)
+        end
```
adamruzicka, line 160: By the curly braces I meant only the ones inside the parentheses. There is not reason to spread this over several lines

app/services/ui_notifications/tasks/task_bulk_stop.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      def create
+        Notification.create!(
+          initiator: initiator,
+          audience: audience,
+          subject: subject,
+          notification_blueprint: blueprint,
+          message: message,
+          notification_recipients: [NotificationRecipient.create({ :user => User.current })]
```
adamruzicka, line 17: Same with the curly brackets here

MariaAga: I just really like them :)

PR #544, opened 2020-05-13
Fixes #29736 - Add bulk force unlock button in tasks table

MariaAga:
Adds a button to bulk stop selected tasks in the tasks table.
ignores "allow dangerous actions" setting
Depends on: https://github.com/theforeman/foreman-tasks/pull/543, https://github.com/theforeman/foreman-tasks/pull/541

### DIFF
--- a/webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
+++ b/webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
@@ -28,7 +28,7 @@ export const ForceUnlockModal = ({ onClick, id, selectedRowsLen }) => (
     title={__('Force Unlock')}
     body={sprintf(
       __(
-        `Resources for %s task(s) will be unlocked and will not prevent other tasks from being run. As the task might be still running, it should be avoided to use this unless you are really sure the task got stuck.`
+        `Resources for %s task(s) will be unlocked and will not prevent other tasks from being run. As the task(s) might be still running, it should be avoided to use this unless you are really sure the task(s) got stuck.`
       ),
       selectedRowsLen
     )}
--- a/webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
@@ -24,7 +24,6 @@ import {
   resumeToastInfo,
   cancelToastInfo,
   toastDispatch,
-  forceCancelToastInfo,
 } from '../TaskActions/TaskActionHelpers';
 
 export const bulkByIdRequest = (resumeTasks, path) => {
@@ -196,31 +195,33 @@ export const bulkForceCancelById = ({
         ),
       })
     );
-  if (stopTasks.length) {
+  if (stopTasks.length > 0) {
     dispatch({ type: TASKS_FORCE_CANCEL_REQUEST });
     try {
       const { data } = await bulkByIdRequest(stopTasks, BULK_FORCE_CANCEL_PATH);
       dispatch({ type: TASKS_FORCE_CANCEL_SUCCESS });
-      if (data.stopped) {
-        data.stopped.forEach(task => {
-          toastDispatch({
-            type: 'forceCancelled',
-            name: task.action,
-            toastInfo: forceCancelToastInfo,
-            dispatch,
-          });
-        });
+      if (data.stopped_length) {
+        dispatch(
+          addToast({
+            type: TOAST_TYPES.SUCCESS,
+            message: sprintf(
+              '%s task(s) cancelled with force',
+              data.stopped_length
+            ),
+          })
+        );
       }
-      dispatch(
-        addToast({
-          type: TOAST_TYPES.WARNING,
-          message: sprintf(
-            '%s task(s) are already stopped',
-            data.skipped_length
-          ),
-        })
-      );
-      if (data.stopped) {
+      if (data.skipped_length > 0)
+        dispatch(
+          addToast({
+            type: TOAST_TYPES.WARNING,
+            message: sprintf(
+              '%s task(s) are already stopped',
+              data.skipped_length
+            ),
+          })
+        );
+      if (data.stopped_length > 0) {
         reloadPage(url, parentTaskID, dispatch);
       }
     } catch (error) {
--- a/webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
@@ -1,8 +1,12 @@
 import API from 'foremanReact/API';
 import { addToast } from 'foremanReact/redux/actions/toasts';
-import { translate as __ } from 'foremanReact/common/I18n';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
 import { TOAST_TYPES } from '../common/ToastsHelpers/ToastTypesConstants';
-import { BULK_CANCEL_PATH, BULK_RESUME_PATH } from './TasksTableConstants';
+import {
+  BULK_CANCEL_PATH,
+  BULK_RESUME_PATH,
+  BULK_FORCE_CANCEL_PATH,
+} from './TasksTableConstants';
 import {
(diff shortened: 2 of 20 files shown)


adamruzicka:
Needs a rebase

laviro:
@MariaAga can you do another rebase?

MariaAga:
Yeah need to make some changes after the API changed

MariaAga:
Rebased, added a small commit with the api change

webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
```diff
@@ -161,3 +168,84 @@ export const bulkCancelById = ({
...
+  if (stopTasks.length < selected.length)
+    dispatch(
+      addToast({
+        type: TOAST_TYPES.WARNING,
+        message: sprintf(
+          '%s task(s) are already stopped',
+          selected.length - stopTasks.length
+        ),
+      })
+    );
```
adamruzicka, line 197: I think this can show up even if there were no already stopped tasks. In that case, warning saying that 0 tasks were already stopped doesn't bring much value

webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
```diff
@@ -28,7 +28,7 @@ export const ForceUnlockModal = ({ onClick, id, selectedRowsLen }) => (
     title={__('Force Unlock')}
     body={sprintf(
       __(
-        `Resources for %s task(s) will be unlocked and will not prevent other tasks from being run. As the task might be still running, it should be avoided to use this unless you are really sure the task got stuck.`
+        `Resources for %s task(s) will be unlocked and will not prevent other tasks from being run. As the task(s) might be still running, it should be avoided to use this unless you are really sure the task(s) got stuck.`
```
adamruzicka, line 31: The button which leads here says "force cancel", maybe not confusing people with unlocking would be a good idea?

webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
MariaAga: selected.length should be more than stoppped.length so it cant be 0

webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
MariaAga: This would mean that force unlock button would open force cancel modal in task details, should we rename it in task details?

webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
adamruzicka: Maybe it wasn't from here, but I was getting a warning that 0 tasks are already stopped

MariaAga: I'll take a look :)

MariaAga: Should be fixed, added `> 0` for `data.skipped_length > 0`

webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
```diff
@@ -161,3 +168,84 @@ export const bulkCancelById = ({
...
+};
+
+export const bulkForceCancelBySearch = ({
+  query,
+  parentTaskID,
+}) => async dispatch => {
+  dispatch({ type: TASKS_FORCE_CANCEL_REQUEST });
+  dispatch(
+    addToast({
+      type: 'info',
```
laviro, line 240: can you reuse the `TOASTS_TYPES.INFO` constant ?
and btw, didn't you created some helper to make the `addToast` action shorter? e.g: `addInfoToast(message)` ?

laviro:
Selection_321

something that I just noticed, can you add a one liner space between the last one and the modal body text?
can you also align the checkbox with the line?

laviro:
It didn't actually forced cancel "Clean up StoredValues" scheduled task,
are there still some tasks that it can not force cancel?

Untitled_ May 26, 2020 5_10 PM

webpack/ForemanTasks/Components/TasksTable/TasksBulkActions.js
MariaAga: Yeah, changed the helper so it would be more generic

MariaAga:
@laviro I can't see the screenshot very well, force cancel should set the state of a task to stopped

laviro:
@MariaAga, here is the link to the original recording: https://drive.google.com/file/d/17rraDthcKqL30fbie-C837861fPkAMeE/view
@adamruzicka any idea what it can be?

adamruzicka:
Do you have logs from that time?

laviro:
@adamruzicka here are the logs: https://pastebin.com/2hZs28Sn

MariaAga:
https://github.com/theforeman/foreman-tasks/pull/547
This pr might solve it?

laviro:
Thanks @MariaAga,
Setting the `dynflow_allow_dangerous_actions` to `true` enabled the process,
I wonder if we should merge #547 first for users like me that had it on `false`, else it will fail.

after force cancelling the task, its `result` info is still stuck on `pending` while `cancelled` is expected:

Selection_323

MariaAga:
I think this is the expected behavior but I'm not sure, @adamruzicka what do you think?

adamruzicka:
I'd say that is expected. In foreman tasks (and dynflow) we distinguish between a graceful cancel, when we tell the task to cancel and it is up to the task to somehow perform the cancellation. When this happens, the state and result are updated accordingly.

Force cancel as implemented here (originally force unlock) is not a cancellation per-se. It is more of a "dont run this task anymore" kind of thing, so it only turns the task over to stopped state, without touching the result.

webpack/ForemanTasks/Components/TaskActions/UnlockModals.js
adamruzicka: Yeah, it would be good to be consistent

laviro:
@adamruzicka thanks for explaining,
To make it clearer in the UI, I have two suggestions:
- renaming it back to `force unlock`
- add buttons in the modal for `Unlock Task` (current implementation) and `Cancel Task Permanently` for cancelling that task for good.

MariaAga:
@laviro I think adding the buttons should be done in a separate pr as  this is only for the bulk action.
Force cancel is already merged

webpack/ForemanTasks/Components/common/ClickConfirmation/index.js
```diff
@@ -29,16 +30,18 @@ export const ClickConfirmation = ({
         <span className={`glyphicon glyphicon-${icon}-sign`} />
         {` ${title}`}
       </ForemanModal.Header>
-      {body}
-      <div>
+      <span>{body}</span>
+      <br />
+      <br />
```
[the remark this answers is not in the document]
MariaAga: I wanted to add `<br>` so it will be consistently one empty line and not margin pixels/precents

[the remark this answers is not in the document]
laviro: According to HTML5 [docs:](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-br-element)

- br elements must be used only for line breaks that are actually part of the content, as in poems or addresses.

- br elements must not be used for separating thematic groups in a paragraph.

IMO it's also better to add a css rule than adding actual elements to the DOM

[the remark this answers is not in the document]
MariaAga: Thanks for the clarification, added css instead

laviro:
Thanks, can you add a test for the last action? Coverage decreased

MariaAga:
Coverage increased :tada:

adamruzicka:
Waiting for Mr. J. to confirm, will merge afterwards

adamruzicka:

PR #545, opened 2020-05-14
Fixes #29594 - Fix wrong bookmark controller

laviro:
The bookmarks controller was changed to `foreman_tasks/tasks`
it means that users that had saved bookmarks on the original controller (`foreman_tasks_tasks`)
can't see all of their original bookmarks..
This change in the UI and the DB migration, changes to the right bookmarks controller.

old bookmark with the same name as the new bookmark
will be renamed to `#{token}_#{name}`..

### DIFF
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTableConstants.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTableConstants.js
@@ -19,5 +19,5 @@ export const UPDATE_CLICKED = 'UPDATE_CLICKED';
 
 export const TASKS_SEARCH_PROPS = {
   ...getControllerSearchProps('tasks'),
-  controller: 'foreman_tasks/tasks',
+  controller: 'foreman_tasks_tasks',
 };
--- a/db/migrate/20200517215015_rename_bookmarks_controller.rb
+++ b/db/migrate/20200517215015_rename_bookmarks_controller.rb
@@ -1,0 +1,35 @@
+class RenameBookmarksController < ActiveRecord::Migration[5.2]
+  def up
+    original_controller = 'foreman_tasks_tasks'
+    original_bookmarks = Bookmark.where(controller: original_controller)
+    original_bookmarks_names = Hash[original_bookmarks.pluck(:name, :id)]
+    
+    new_controller = 'foreman_tasks/tasks'
+    new_bookmarks = Bookmark.where(controller: new_controller)
+    new_bookmarks.find_each do |new_bookmark|
+      name = new_bookmark.name
+      is_name_taken = original_bookmarks_names.key? name
+
+      if is_name_taken
+        original_bookmark = original_bookmarks.find(original_bookmarks_names[name])
+        is_duplicated = original_bookmark.query == new_bookmark.query &&
+                        original_bookmark.owner_id == new_bookmark.owner_id &&
+                        original_bookmark.owner_type == new_bookmark.owner_type &&
+                        original_bookmark.public == new_bookmark.public
+
+        if is_duplicated
+          original_bookmark.destroy     
+        else
+          modified_name = "#{name}_#{generate_token}"
+          original_bookmark.update(name: modified_name)
+        end
+      end
+      # Revert to the original controller name
+      new_bookmark.update(controller: original_controller)
+    end
+  end
+
+  def generate_token
+    SecureRandom.base64(5).gsub(/[^0-9a-z ]/i, '')
+  end
+end
--- a/webpack/ForemanTasks/Components/TasksTable/__tests__/__snapshots__/TasksTablePage.test.js.snap
+++ b/webpack/ForemanTasks/Components/TasksTable/__tests__/__snapshots__/TasksTablePage.test.js.snap
@@ -85,7 +85,7 @@ exports[`TasksTablePage rendering render with Breadcrubs 1`] = `
           "documentationUrl": "4.1.5Searching",
           "url": "/api/bookmarks",
         },
-        "controller": "foreman_tasks/tasks",
+        "controller": "foreman_tasks_tasks",
       }
     }
     searchQuery="a=b"
@@ -255,7 +255,7 @@ exports[`TasksTablePage rendering render with minimal props 1`] = `
           "documentationUrl": "4.1.5Searching",
           "url": "/api/bookmarks",
         },
-        "controller": "foreman_tasks/tasks",
+        "controller": "foreman_tasks_tasks",
       }
     }
     searchQuery="a=b"


db/migrate/20200514214910_rename_bookmarks_controller.rb
```diff
@@ -0,0 +1,16 @@
+class RenameBookmarksController < ActiveRecord::Migration[5.2]
+  def up
+    original_controller = 'foreman_tasks_tasks'
+    original_bookmarks = Bookmark.where({ :controller => original_controller })
+    original_bookmarks_names = original_bookmarks.map { |b| b.name }
+    
+    mistake_controller = 'foreman_tasks/tasks'
+    mistake_bookmarks = Bookmark.where({ :controller => mistake_controller })
```
ShimShtein, line 8:
```suggestion
    mistake_bookmarks = Bookmark.where(controller: mistake_controller)
```

db/migrate/20200514214910_rename_bookmarks_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+    mistake_bookmarks.each do |b|
+      name = b.name
+      name_is_taken = original_bookmarks_names.include? name
+      name = name_is_taken ? "migrated_#{name}" : name
```
laviro, line 12: Should we rename the _`mistake`_ bookmark or the _`original`_ bookmark?

db/migrate/20200514214910_rename_bookmarks_controller.rb
```diff
@@ -0,0 +1,16 @@
+class RenameBookmarksController < ActiveRecord::Migration[5.2]
+  def up
+    original_controller = 'foreman_tasks_tasks'
+    original_bookmarks = Bookmark.where({ :controller => original_controller })
+    original_bookmarks_names = original_bookmarks.map { |b| b.name }
+    
+    mistake_controller = 'foreman_tasks/tasks'
+    mistake_bookmarks = Bookmark.where({ :controller => mistake_controller })
+    mistake_bookmarks.each do |b|
```
ShimShtein, line 9: I know it's not important here, since you won't have too much records, but it makes sense to use [`#find_each`](
https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) method for iteration over database results:
```suggestion
    mistake_bookmarks.find_each do |b|
```

db/migrate/20200514214910_rename_bookmarks_controller.rb
ShimShtein: I would vote for `changed_` or `lost_`.

ShimShtein: Maybe even `recovered_` for both the variable and the prefix

adamruzicka: I would rename the original one (the one with target name). This way the users will get what they created last

Also I would use a similar approach as in https://github.com/theforeman/foreman/pull/7532 and used a short random string instead of the `migrated_` prefix

ShimShtein: I like the random part. We can also try and remove one of them (the one we're iterating over sounds more reasonable) in case the `query` is the same. If they are different - we will need to prefix.

laviro:
Thanks @ShimShtein @adamruzicka, updated
now the old bookmark will be renamed,
or removed in case of equal `queries` and `names`

adamruzicka:
Shouldn't this take other attributes of the bookmark (such as `owner_id`) into consideration when deciding whether it is a duplicate or not?

laviro:
> Shouldn't this take other attributes of the bookmark (such as owner_id) into consideration when deciding whether it is a duplicate or not?

Thanks, added also `owner_id`, `owner_type` & `public`

webpack/ForemanTasks/Components/TasksTable/TasksTableConstants.js
```diff
@@ -19,5 +19,5 @@ export const UPDATE_CLICKED = 'UPDATE_CLICKED';
 
 export const TASKS_SEARCH_PROPS = {
   ...getControllerSearchProps('tasks'),
-  controller: 'foreman_tasks/tasks',
+  controller: 'foreman_tasks_tasks',
```
laviro, line 22: @MariaAga do we count on this controller for other things except bookmarks? thanks

MariaAga: only `SearchBar` uses this const, and it uses this for autocomplete and bookmarks

laviro: cool, I saw no regression with the searchbar autocomplete,
this prop is for info and a unique key for the redux store, though the API calls are not affected.

db/migrate/20200514214910_rename_bookmarks_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+        original_bookmark = original_bookmarks.where(name: name).first
+        is_duplicated = original_bookmark.query == new_bookmark.query &&
+                        original_bookmark.owner_id == new_bookmark.owner_id &&
+                        original_bookmark.owner_type == new_bookmark.owner_type &&
+                        original_bookmark.public == new_bookmark.public
+
+        if is_duplicated
+          original_bookmark.destroy     
+        else
+          modified_name = "#{SecureRandom.base64(5)}_#{name}"
```
laviro, line 23: should we add the letter 'Z' before the token to move all of the migrated values to the bottom of the dropdown?
and should it be the same token for all of them? for example all of the migrated values will look like `z_10k3n_value`
Selection_318

db/migrate/20200514214910_rename_bookmarks_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+    original_bookmarks_names = original_bookmarks.pluck(:name)
+    
+    new_controller = 'foreman_tasks/tasks'
+    new_bookmarks = Bookmark.where(controller: new_controller)
+    new_bookmarks.find_each do |new_bookmark|
+      name = new_bookmark.name
+      is_name_taken = original_bookmarks_names.include? name
+
+      if is_name_taken
+        original_bookmark = original_bookmarks.where(name: name).first
```
ShimShtein, line 14: To make the search more efficient, it makes sense to have the `id` of the old bookmark already in hand:
When getting original bookmarks, we can also pluck the `id` and store both in hash structure:
``` ruby
    original_bookmarks_names = Hash[original_bookmarks.pluck(:name, :id)]
```
Later, when you want to check if the name already exists, you can use the `key?` method:
``` ruby
       is_name_taken = original_bookmarks_names.key? name
```
and at last when you want to select the whole record, you can use `find` method with the `id` from hash:
``` ruby
        original_bookmark = original_bookmarks.find(original_bookmarks_names[name])
```

db/migrate/20200514214910_rename_bookmarks_controller.rb
ShimShtein: We should probably remove the non-letters from the hash for better readability.

Personally I don't have a strong opinion for the `z_` prefix, I think it would be better to have it near the original one:
```suggestion
          modified_name = "#{name}_#{cleanup(SecureRandom.base64(5))}"
```

db/migrate/20200514214910_rename_bookmarks_controller.rb
laviro: that's awesome, thanks!

db/migrate/20200514214910_rename_bookmarks_controller.rb
laviro: Agree, I will update.. thanks

ShimShtein:
LGTM

adamruzicka:
Thank you @laviro & @ShimShtein !

PR #546, opened 2020-05-18
Refs #29423 - Enforce the polling multiplier to be >=1

adamruzicka:


adamruzicka:
Thank you @pondrejk !

PR #547, opened 2020-05-19
Fixes #29875  - Remove allowDangerousActions setting

MariaAga:
The user already have to confirm (twice) the actions that are behind this setting.(checkbox confirm + button confirm)
And this will result in the new bulk stop being open for all users


MariaAga:
Redmine issue here: https://projects.theforeman.org/issues/29875

MariaAga:
Fixed, thanks!

MariaAga:
@adamruzicka Can you take a look? This also fixes a bug where users cant use force cancel in the table if their setting is set to `false`

PR #550, opened 2020-06-03
Fixes #29494 - task details show spinner instead of buttons

MariaAga:
When loading buttons are not shown, and the other tabs are disabled:
image
I had to change the integration test so recommended to look at the snapshot file as a whole and not the differences


adamruzicka:
Needs a rebase

MariaAga:
rebased

adamruzicka:
It works well when everything is fine, but if the server for some reason fails to give the task data (for example hits an ISE), then the page stays in the loading state. Shouldn't it show that the request could not be completed and disable the loading state?

MariaAga:
Good suggestion, I've changed the logic so if the reload fails instead of refreshing the page it will stop the reloading and show the error

adamruzicka:
Now the page just goes blank when the server gives a 500
image

adamruzicka:
Nevermind, the page does not go blank. It was caused by an overly long task label, which made the page too wide (which I haven't noticed) and the error message got displayed off screen.

adamruzicka:
One more rebase please

adamruzicka:
@laviro could you take a look please?

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsActions.js
```diff
@@ -38,29 +38,32 @@ export const taskReloadStart = (
   loading = false
 ) => {
   if (!timeoutId) {
-    timeoutId = setInterval(() => refetchTaskDetails(id, loading), 5000);
+    timeoutId = setInterval(
+      () => refetchTaskDetails(id, loading, timeoutId),
```
laviro, line 42: what do you think about using the Interval middleware from core?
you can get the interval ID from the store, instead of passing it around from action to action

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsReducer.js
```diff
@@ -1,32 +1,37 @@
...
       return state.merge({
-        loading: false,
+        status: STATUS.RESOLVED,
+        isData: true,
         ...payload,
         taskReload,
         timeoutId,
       });
+    case FOREMAN_TASK_DETAILS_FETCH_TASK_FAILURE:
+      return state.merge({ status: STATUS.ERROR, error: payload });
```
laviro, line 34: you can use the API middleware where all of this logic already made for you,
and then just select the status and error from the state.

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
```diff
@@ -76,13 +80,13 @@ class TaskDetails extends Component {
...
-          <Tab eventKey={3} title={__('Errors')}>
+          <Tab eventKey={3} disabled={loading} title={__('Errors')}>
             <Errors executionPlan={executionPlan} failedSteps={failedSteps} />
           </Tab>
-          <Tab eventKey={4} title={__('Locks')}>
+          <Tab eventKey={4} disabled={loading} title={__('Locks')}>
             <Locks locks={locks} />
           </Tab>
-          <Tab eventKey={5} title={__('Raw')}>
+          <Tab eventKey={5} disabled={loading} title={__('Raw')}>
```
laviro, line 89: how about showing a tabs loading skeleton or a large spinner until the data is loaded,
and only then show the real tabs? feels like a better UX

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsActions.js
MariaAga: Yeah I'll add it to redmine and do another PR

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
MariaAga: What do you mean by tabs loading skeleton? AFAIK we only have rows loading skeleton

laviro: it's the "skeleton" of the page, we can mimic the way the page looks and give it a loading feeling instead of using a large spinner, e.g:
1_yk5bG1DBaOU_KiXASE9dsA

MariaAga: We don't have this implemented in PF or foreman, and adding a skeleton to each information line in the code will look messy. Also the spinner in the screenshot is the largest one we have

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsActions.js
MariaAga: https://github.com/theforeman/foreman-tasks/pull/564

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
laviro: we spoke about it offline back than, the library that we are using for `skeleton`
in foreman is [react-loading-skeleton](https://github.com/dvtng/react-loading-skeleton#react-loading-skeleton). with this library you can customize even full pages loading, instead of spinners.. which result in a better UX.

see also [foreman usage](https://github.com/theforeman/foreman/blob/7553eea533daa7de9f2f6f7ee1a827d15d3c714e/webpack/assets/javascripts/react_app/routes/Audits/AuditsPage/components/AuditsLoadingPage.js#L2)

adamruzicka:
Needs a rebase

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
MariaAga: Screenshot from 2020-07-09 17-23-54
We decided to use one big skeleton, what do you think of this one?

laviro: Does it have animation?

MariaAga: yeah, the video quality is not the best :sweat_smile: 
Screencast from 07-09-2020 05_20_05 PM

laviro: it feels a bit heavy as a one big piece,
I would suggest replacing the existing buttons/text with the current layout, it feels lighter:
Untitled_ Jul 12, 2020 3_13 PM

you can do the same for the rest of the tabs, and it will be a better UX than disabling/enabling each tab

MariaAga: I don't think we should do it for each tab as the loading time won't be long enough for it to matter.. This loading state is only when the page opens. I'll add the split skeleton to this PR

webpack/ForemanTasks/Components/TaskDetails/Components/Task.js
```diff
@@ -47,96 +52,109 @@ const Task = props => {
...
+              <Button
+                className="force-unlock-button"
+                bsSize="small"
+                disabled={state === 'stopped'}
+                onClick={forceUnlockModalActions.setModalOpen}
+              >
+                {__('Force Unlock')}
+              </Button>
+            </Col>
+          )}
```
laviro, line 157: since this section grew really big, can you move all buttons into another component?

MariaAga: Yeah good idea

adamruzicka:
Mr. T is unhappy

MariaAga:
He's happy now :)

adamruzicka:

PR #551, opened 2020-06-04
Fixes #30009 - v1.2.0 should require foreman >= 2.1.0

laviro:
With the addition of `ActionButtons` which depends on foreman > 2.1.0
while running webpack compile with a bit older foreman, I got the error:
```bash
ModuleNotFoundError: Module not found: Error: Can't resolve 'foremanReact/components/common/ActionButtons/ActionButtons' in '/home/travis/.rvm/gems/ruby-2.5.8/gems/foreman-tasks-1.2.0/webpack/ForemanTasks/Components/common/ActionButtons'
```

cc @adamruzicka @MariaAga @ShimShtein


adamruzicka:
Thank you for spotting that, I completely missed it.

PR #552, opened 2020-06-04
Fixes #30011 - Remove internal use of deprecated JS API

MariaAga:
The API object was moved from 'foremanReact/API' to 'foremanReact/redux/API'


adamruzicka:
I haven't hit any issues messing around with this, but I'll leave the final decision to @laviro

adamruzicka:
Thank you both!

PR #554, opened 2020-06-04
Fixes #30018 - Add refresh button to tasks table

MariaAga:
Adds ability to refresh the task states by clicking on a refresh button rather than refreshing the whole page\session.
Not sure where to place the button:
A: image
B: image

### DIFF
--- a/webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
+++ b/webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { getURIsearch } from 'foremanReact/common/urlHelpers';
-import { Spinner } from 'patternfly-react';
+import { Spinner, Button, Icon } from 'patternfly-react';
 import PageLayout from 'foremanReact/routes/common/PageLayout/PageLayout';
 import { translate as __ } from 'foremanReact/common/I18n';
 import { getURIQuery } from 'foremanReact/common/helpers';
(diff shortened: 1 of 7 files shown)


adamruzicka:
I would probably go with B. @rohoover opinions?

adamruzicka:
Does this refresh only the table or does it affect the dashboard as well?

When clicking "Refresh Data", a spinner appears at the left side of the buttons and (possibly because it is taller than the buttons) the entire buttons row gets shifted a fex pixels down. I think the buttons should stay where they are.

Apart from the two above, it seems to work well

pondrejk:
Voting for B as it leaves the other 2 buttons in their original places, +1 to Adam's comments

MariaAga:
Thanks for the feedback! I changed the spinner size to fit the buttons so it won't jump:
Before:
image
After:
image

Refresh data refreshes the table and dashboard.

webpack/ForemanTasks/Components/TasksTable/TasksTablePage.js
```diff
@@ -63,7 +63,10 @@ const TasksTablePage = ({
         toastNotifications="foreman-tasks-cancel"
         toolbarButtons={
           <React.Fragment>
-            {props.status === STATUS.PENDING && <Spinner size="lg" loading />}
+            <Button onClick={() => props.reloadPage(url, props.parentTaskID)}>
+              {__('Refresh Data')}
```
laviro, line 67: can you add a refresh icon too? e.g: https://fontawesome.com/v4.7.0/icon/refresh
or maybe it can be just a refresh icon without the text.. saw that in several places of our app

MariaAga:
Added an icon to the button 
Screenshot from 2020-06-24 11-33-53

adamruzicka:

PR #557, opened 2020-06-11
Fixes #30103 - Add index to tasks_locks table

tbrisker:
This index leads to significant improvement in speed of certain queries
looking for locks related to a specific task and resource type.

see https://community.theforeman.org/t/slow-products-page-on-2-0/19142/ for details.

### DIFF
--- a/db/migrate/20200611090846_add_task_lock_index_on_resource_type_and_task_id.rb
+++ b/db/migrate/20200611090846_add_task_lock_index_on_resource_type_and_task_id.rb
@@ -1,0 +1,9 @@
+class AddTaskLockIndexOnResourceTypeAndTaskId < ActiveRecord::Migration[6.0]
+  def change
+    add_index :foreman_tasks_locks, [:task_id, :resource_type, :resource_id], name: 'index_tasks_locks_on_task_id_resource_type_and_resource_id'
+    # These indexes are not needed as they can be gained from partial index lookups
+    remove_index :foreman_tasks_locks, :task_id
+    remove_index :foreman_tasks_locks, :name
+    remove_index :foreman_tasks_locks, :resource_type
+  end
+end


db/migrate/20200611090846_add_task_lock_index_on_resource_type_and_task_id.rb
```diff
@@ -0,0 +1,5 @@
+class AddTaskLockIndexOnResourceTypeAndTaskId < ActiveRecord::Migration[6.0]
+  def change
+    add_index :foreman_tasks_locks, [:resource_type, :task_id]
```
adamruzicka, line 3: Should we add `:resource_id` as well?

tbrisker: updated, and cleaned up a couple of indexes that are no longer required as they are covered by other indexes (as postgres can use leading columns in an index for searching as well)

db/migrate/20200611090846_add_task_lock_index_on_resource_type_and_task_id.rb
```diff
@@ -0,0 +1,8 @@
+class AddTaskLockIndexOnResourceTypeAndTaskId < ActiveRecord::Migration[6.0]
+  def change
+    add_index :foreman_tasks_locks, [:task_id, :resource_type, :resource_id], name: 'index_tasks_locks_on_task_id_resource_type_and_resource_id'
+    # These indexes are not needed as they can be gained from partial index lookups
```
adamruzicka, line 4: On my machine I have the following indexes:
```
Indexes:
    "foreman_tasks_locks_pkey" PRIMARY KEY, btree (id)
    "index_foreman_tasks_locks_name_resource_type_resource_id" btree (name, resource_type, resource_id)
    "index_foreman_tasks_locks_on_exclusive" btree (exclusive)
    "index_foreman_tasks_locks_on_name" btree (name)
    "index_foreman_tasks_locks_on_resource_type" btree (resource_type)
    "index_foreman_tasks_locks_on_resource_type_and_resource_id" btree (resource_type, resource_id)
    "index_foreman_tasks_locks_on_task_id" btree (task_id)
```

Index on `task_id` is covered by index on `[:task_id, :resource_type, :resource_id]` and I assume index on `name` is covered by index on `[name, resource_type, resource_id]`, correct?

Following the example of `name`, could we then drop an index on `resource_type` because it should be covered by `[:task_id, :resource_type, :resource_id]`?

Do partial index lookups work on a single column or any subset? Is `[:resource_type, :resource_id]` covered by `[:task_id, :resource_type, :resource_id]`?

adamruzicka: Oh, missed your previous comment. Nevermind then

adamruzicka: But we should still be able to drop the index on `resource_type` as it should be covered by index on `[resource_type, resource_id]`, no?

tbrisker: No, this only works for the left-most part of the column list (i.e., `[:task_id, :resource_type, :resource_id]` will cover searches by `:task_id` and `:task_id, :resource_type` but not `:resource_type, :resource_id` or `:resource_id` or `:task_id, :resource_id`)

adamruzicka: If it works on the left-most part of the column list, then we should be able to drop
```
"index_foreman_tasks_locks_on_resource_type" btree (resource_type)
```
in favor of 
```
"index_foreman_tasks_locks_on_resource_type_and_resource_id" btree (resource_type, resource_id)
```

or am I still missing something?

tbrisker: Oh, that's interesting, for some reason I'm not seeing that index on my schema.

tbrisker: nor do i see where it was created by going over the db/migrate folder, could this be some leftover you have in your database from some experiment?

adamruzicka: It is a fresh 2.1rc2 and I haven't done anything strange to it. The index should be created from here https://github.com/theforeman/foreman-tasks/blob/master/db/migrate/20131209122644_create_foreman_tasks_locks.rb#L10

tbrisker: huh, i guess it is my db that is wrong :) i missed the index:true there. Updated.

tbrisker:
Pretty sure test failure is unrelated: 

`DashboardIntegrationTest::with origin::out of sync disabled.test_0001_has no out of sync link`

adamruzicka:
Thank you @tbrisker!

PR #558, opened 2020-06-11
 Fixes #30049 - cancel step in tasks details use correct url

MariaAga:
In task details > Running steps tab, the url leads to /tasks/undefined/cancel_step


app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -31,9 +31,12 @@ def sub_tasks
 
     def cancel_step
       task = find_dynflow_task
-      flash[:info] = _('Trying to cancel step %s') % params[:step_id]
-      ForemanTasks.dynflow.world.event(task.external_id, params[:step_id].to_i, ::Dynflow::Action::Cancellable::Cancel).wait
-      redirect_to foreman_tasks_task_path(task)
+      result = ForemanTasks.dynflow.world.event(task.external_id, params[:step_id].to_i, ::Dynflow::Action::Cancellable::Cancel).wait
```
adamruzicka, line 34: It could happen that delivery of the event takes too long and before `ForemanTasks.dynflow....wait` returns, the http request times out. Will the frontend be able to deal with it?

MariaAga: no, one min =>failure ):
So we can just show 'trying' toast or also add a notification like in the bulk action, what do you think is better?

adamruzicka: I'd say it is fine the way it is now

adamruzicka:
Thank you @MariaAga !

PR #560, opened 2020-06-17
Fixes #30145 - Make sure rake tasks run as Dynflow clients

adamruzicka:
(cherry picked from commit b192911)


ezr-ondrej:
Hmm rebasing resulted in different commit, so now we should CP https://github.com/theforeman/foreman-tasks/commit/b1929119a77179a634c8bda3434c7830ea95ec34

adamruzicka:
Right, I'll redo the CPs

PR #566, opened 2020-06-25
Fixes #30210 - Fix task progress dp discrepancy

ianballou:
There is a task progress decimal precision discrepancy between the UI, Hammer, and the API.  Hammer shows no decimals, the API shows many, and the UI shows at most 2.  This PR updates the UI to have the progress be an integer (no more decimals) and updates the API progress fraction to have the same precision as the UI's value.  E.g., if the task value in the UI and CLI is 83%, the progress from the API will be 0.83.

### DIFF
--- a/webpack/ForemanTasks/Components/TaskDetails/TaskDetailsSelectors.js
+++ b/webpack/ForemanTasks/Components/TaskDetails/TaskDetailsSelectors.js
@@ -30,7 +30,7 @@ export const selectErrors = state => {
 
 export const selectProgress = state =>
   selectTaskDetails(state).progress
-    ? parseFloat((selectTaskDetails(state).progress * 100).toFixed(2))
+    ? Math.trunc(selectTaskDetails(state).progress * 100)
     : 0;
 
 export const selectUsername = state =>
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -46,7 +46,8 @@ module ForemanTasks
     end
 
     def progress
-      execution_plan.try(:progress) || 0
+      progress_raw = execution_plan.try(:progress) || 0
+      progress_raw.round(2)
     end
 
     def execution_plan(silence_exception = true)


ianballou:
Not seeing the connection between my PR and those test failures.

adamruzicka:
> This PR updates the UI to have the progress be an integer (no more decimals)
Why? To fix what hammer shows?

ianballou:
@adamruzicka this PR is intended to make the UI show the same style percentage as hammer. The other route would've been to give hammer 2 decimals, but it looks like that would've required patching PowerBar. Plus, it's questionable whether or not we need decimal precision for our task progress.

ianballou:
@adamruzicka do you think this PR is in the right direction for addressing the issue? Also looking for opinions on if the issue is worth addressing since it isn't a high-profile problem.

https://projects.theforeman.org/issues/30210

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsSelectors.js
```diff
@@ -30,7 +30,7 @@ export const selectErrors = state => {
 
 export const selectProgress = state =>
   selectTaskDetails(state).progress
-    ? parseFloat((selectTaskDetails(state).progress * 100).toFixed(2))
+    ? parseInt(selectTaskDetails(state).progress * 100, 10)
```
adamruzicka, line 33: The `parseInt` doesn't feel right (shouldn't it do `String -> Integer`?).

The number here should be 0 <= number <= 1 with precision to two decimals so multiplying it by 100 should yield an integer already. If we want to be sure we really get an integer percentage, wouldn't something like `Math.trunc` be better?

@laviro opinions on this?

laviro: Yeah I agree, `Math.trunc` is exactly for that, to cut off the dot and the digits to the right of the number.
although multiplying by 100 should do the work in this case, it will "protect" us in cases where there's a decimal.

ianballou:
@adamruzicka, @laviro I've addressed your comments.

adamruzicka:
Thank you @ianballou !

PR #568, opened 2020-07-01
Fixes #30272 - Add foreman-js link script

laviro:


laviro:
@adamruzicka can we merge it?
it is already being tested with [foreman-js GH actions](https://github.com/theforeman/foreman-js/pull/143) and works well

adamruzicka:
Thank you @laviro !

PR #567, opened 2020-06-30
Fixes #30248 - upgrade vendor to v4.10.0

sharvit:

PR #569, opened 2020-07-01
Fixes #30274 - Fix tasks table page test snapshot

laviro:
oh, seem that it will only fail if we upgrade `theforeman` npm packages.
found by [foreman-js GH action PR](https://github.com/theforeman/foreman-js/pull/143)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+cache:
+  npm: false
 node_js:
   - '10'
   - '12'
--- a/script/travis_run_js_tests.sh
+++ b/script/travis_run_js_tests.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -ev
 if [[ $( git diff --name-only HEAD~1..HEAD webpack/ .travis.yml babel.config.js .eslintrc package.json | wc -l ) -ne 0 ]]; then
+  npm ls @theforeman/test;
   npm run test;
   npm run publish-coverage;
   npm run lint;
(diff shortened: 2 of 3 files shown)


MariaAga:
Looks like this is fixed with the foreman-js update here: https://github.com/theforeman/foreman-tasks/pull/567

laviro:
This issue is happening now, unless we pin `@theforeman` packages to <4.10.0
by default npm installs 4.10.1 on my machine, this fail is already visible also in Foreman core tests.

MariaAga:
But doesn't it mean that it will fail in other machines after adding this pr?

laviro:
on travis [log](https://travis-ci.org/github/theforeman/foreman-tasks/jobs/703900435#L239) you can see travis still uses an old version: `4.8.0` while if you re-install it locally you will get `v4.10.1`

laviro:
@adamruzicka do we have caching in travis? maybe we should disable it somewhere to get always the updated packages.
trying to run tests again.

adamruzicka:
According to the [docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#caching-with-npm) npm deps are being cached. Although I'm not a huge fan of downloading all the npm deps every time the tests run, but let's try disabling it for now and see if it helps

laviro:
Thanks, because this error is visible on local install and in react-tests on `foreman` & `foreman-js`, only in travis it isn't updated yet.

adamruzicka:
Wouldn't it be better to update package.json to have the newer version? As far as I understand travis runs `npm ci` which pulls the exact versions which are described in `package.json`

laviro:
interesting, now it's [without cache](https://travis-ci.org/github/theforeman/foreman-tasks/jobs/704234707#L360) and original snapshot passes, but foreman-js updated from yesterday to `v4.10.2`

laviro:
tested also locally, with `@theforeman` packages `v4.10.2` the snapshot is back to what it was.
closing this for now :)
but we definitely should disable npm cache to be inline with those libraries, opened a PR in https://github.com/theforeman/foreman-tasks/pull/570

PR #570, opened 2020-07-02
Fixes #30289 - disable travis npm cache

laviro:
the npm cache prevent using updated versions of libraries such as `@theforeman` packages.


laviro:
@adamruzicka  @MariaAga can we merge this?
it is very important to stay aligned with foreman-js and other library changes

adamruzicka:
Thank you @laviro !

PR #572, opened 2020-07-09
Consume foreman in js tests

laviro:
instead of using mocks for JS tests,
in this POC I am trying to use the actual code from foreman.

**works only with a foreman-js link from foreman


laviro:
@sharvit should I close this POC?
what are the near-future plans with foreman-js toolkit for plugins + foreman JS tests?

sharvit:
@laviro I would like to have it in and encourage other plugins to use mocks wisely instead of the "mock all foremanReact" policy we have now.

As for foreman-js, we have https://github.com/theforeman/foreman-js/pull/194 and https://github.com/theforeman/foreman-js/pull/192 which related to the topic, it would be nice to test against them and help to improve them.

As for this PR, I am wondering if there are mocks that are necessary for plugins to use (API?) and if we should provide them for plugins to reuse?

adamruzicka:
What's the status here?

adamruzicka:
@laviro ping

adamruzicka:
@laviro ping

laviro:
Worth another try, but I might start from scratch, closing for now, thanks @adamruzicka

PR #573, opened 2020-07-14
Fixes #30397 - Add DSL docs for tasks

ofedoren:


app/models/foreman_tasks/task.rb
```diff
@@ -70,6 +70,17 @@ def check_permissions_after_save
            end)
     scope :for_action_types, (->(action_types) { where('foreman_tasks_tasks.label IN (?)', Array(action_types)) })
 
+    apipie :class, "A class representing #{model_name.human} object" do
+      name 'Task'
+      refs 'Task'
+      sections only: %w[all additional]
+      property :main_action, Class, desc: 'Returns the main action of the task, e.g. Actions::RemoteExecution::RunHostJob'
```
adamruzicka, line 77: This should probably state that the return value isn't a class, but rather the instance

ofedoren: Thanks, @adamruzicka, updated.

app/models/foreman_tasks/task.rb
```diff
@@ -70,6 +70,17 @@ def check_permissions_after_save
            end)
     scope :for_action_types, (->(action_types) { where('foreman_tasks_tasks.label IN (?)', Array(action_types)) })
 
+    apipie :class, "A class representing #{model_name.human} object" do
+      name 'Task'
+      refs 'Task'
+      sections only: %w[all additional]
+      property :main_action, object_of: Class, desc: 'Returns the main action of the task, e.g. Actions::RemoteExecution::RunHostJob'
```
adamruzicka, line 77: This says that the returned thing is an instance of class `Class`? If so, could we change it to `Dynflow::Action`?

ofedoren: Makes more sense, thank you. Updated.

adamruzicka:
Thank you @ofedoren !

PR #574, opened 2020-07-17
Add GH actions tests

ezr-ondrej:
Add GH action testing.
Here it ran: https://github.com/ezr-ondrej/foreman-tasks/actions/runs/172770987
We are incompatible with 2.1 by the way because of the `apipie` dsl not being present (so I've left only testing with `develop`)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+      - uses: actions/checkout@v2
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby-version }}
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version:  ${{ matrix.node-version }}
+      - name: Nmp install
```
laviro, line 92:
```suggestion
      - name: Npm install
```

ezr-ondrej: great :eye:

ezr-ondrej:
Dropped Travis

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+          bundle install --jobs=3 --retry=3
+          bundle exec rake db:create
+          bundle exec rake db:migrate
+      - name: Run plugin tests
+        run: |
+          bundle exec rake test:foreman-tasks
+          bundle exec rake test TEST="test/unit/foreman/access_permissions_test.rb"
+  test_js:
+    runs-on: ubuntu-latest
+    needs: rubocop
```
laviro, line 76: can you add paths here, so like in [core](https://github.com/theforeman/foreman/blob/develop/.github/workflows/plugins_react_tests.yml#L6) it will run only when the `webpack` folder etc.. have changed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+          node-version:  ${{ matrix.node-version }}
+      - name: Npm install
+        run: |
+          npm install
+      - name: Run plugin linter
+        run: |
+          npm run lint
+      - name: Run plugin tests
+        run: |
+          npm run test
```
laviro, line 100: I think that for one-liner commands we don't need the pipes
```suggestion
      - name: Npm install
        run: npm install
      - name: Run plugin linter
        run: npm run lint
      - name: Run plugin tests
        run: npm run test
```

lib/foreman_tasks/tasks/dynflow.rake
```diff
@@ -1,5 +1,5 @@
 namespace :dynflow do
-  task :client do
+  task :client => :environment do
```
laviro, line 2: @adamruzicka any thoughts about this?

.github/workflows/ci.yml
ezr-ondrej: not as easy to do it on the jobs level. I would need to create another workflow for UI testing, or use something like https://github.com/dorny/paths-filter

lib/foreman_tasks/tasks/dynflow.rake
ezr-ondrej: Droping it and regenerating `.rubocop_todo.yml`.
It doesn't belong to this PR. :)

.github/workflows/ci.yml
laviro: Right, want to move it to a `js_ci.yml` workflow?

ezr-ondrej: not really xD now we have to manage two matrixes of versions to test, but done so. Take a look and let me know what you think :)

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,37 @@
+name: JavaScript Testing
+on:
+  pull_request:
+    paths:
+      - 'webpack/**'
+      - 'package.json'
+      - 'config/webpack.config.js'
```
laviro, line 7: no need for this one here :)

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,37 @@
...
+      - '.github/workflows/js_tests.yml'
+env:
+  RAILS_ENV: test
+  DATABASE_URL: postgresql://postgres:@localhost/test
+  DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
```
laviro, line 12: I think we can remove this section

.github/workflows/js_tests.yml
```diff
@@ -0,0 +1,37 @@
...
+      fail-fast: false
+      matrix:
+        ruby-version: [2.6]
+        node-version: [10, 12]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby-version }}
```
laviro, line 27: ditto

ezr-ondrej:
Here the last two runs are for the current changes: https://github.com/ezr-ondrej/foreman-tasks/actions?query=branch%3Agh_actions_local+event%3Apush

adamruzicka:
I'm fine with this, @ezr-ondrej could you also open a PR to foreman-infra to prevent PR tests from being triggered on jenkins?

ezr-ondrej:
Not sure, hope https://github.com/theforeman/foreman-infra/pull/1397 is correct :)

adamruzicka:
Thank you @ezr-ondrej & @laviro !

PR #575, opened 2020-08-10
Fixes #30622 - Add ID to recurring logics table

MariaAga:
Makes the ID clickable to the logic instead on the cron line
Screenshot from 2020-08-10 17-36-13


adamruzicka:
Thank you @MariaAga !

PR #577, opened 2020-08-10
Fixes #30621 - add toast to recurring logic state change

MariaAga:
When user enables/disables a recurring logic they will see a toast informing them of a successful action.


adamruzicka:
Thank you @MariaAga !

PR #578, opened 2020-08-11
Fixes #30633 - Stopped chart doesnt count cancelled

MariaAga:
in tasks dashboard the stopped chart only had counts for 'warning', 'error' and 'success', and was missing 'cancelled' and 'pending' result counts.
Added an 'other' counter for the total of 'cancelled' and 'pending'
Screenshot from 2020-08-11 16-45-20
Design was approved :smile:


adamruzicka:
Could we allow searching for "others" by clicking the number just like with other numbers in the table? Also, why not make it a part of the table?

MariaAga:
Good idea, I'll add it as a filter. because we'll have to add 2 more rows ( 'cancelled' and 'pending') making the card too big, just adding the 'other' made the card bigger

adamruzicka:
I meant doing it the same way as it is now, just adding it to the table instead of putting it below it. So you'd add just one row called "other"

MariaAga:
I think that its a bit meaningless to count "Other total" and "other in the last 24 hours"

adamruzicka:
> I think that its a bit meaningless

Ok :+1:

app/services/foreman_tasks/dashboard_table_filter.rb
```diff
@@ -24,7 +24,11 @@ def scope
     private
 
     def scope_by(field)
-      @new_scope = @new_scope.where(field => @params[field]) if @params[field].present?
+      if (field == :result) && (@params[field] == 'other')
+        @new_scope = @new_scope.where(:result => ['cancelled', 'pending'])
+      elsif @params[field].present?
+        @new_scope = @new_scope.where(field => @params[field])
+      end
     end
```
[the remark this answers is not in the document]
MariaAga: I thought this was better so the frontend labels won't have to convert an array to "other" and mess with prop types for label expecting a string but getting an array

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
```diff
@@ -26,6 +27,8 @@ const StoppedTasksCard = ({
   ...props
 }) => {
   const { STOPPED } = TASKS_DASHBOARD_AVAILABLE_QUERY_STATES;
+  const otherCount = data.other || 0;
+  delete data.other;
```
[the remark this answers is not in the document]
MariaAga: To not send it to the stopped table mapper, I can also filter it out so it will be softer..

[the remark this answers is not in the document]
adamruzicka: It is fine, I was just curious what was the reasoning behind it

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardSelectors.js
```diff
@@ -12,6 +12,12 @@ export const selectTime = state =>
 
 export const selectQuery = state => selectTasksDashboard(state).query || {};
 
+export const calcStoppedOther = data =>
+  (data?.cancelled?.recent || 0) +
+  (data?.cancelled?.total || 0) +
+  (data?.pedning?.recent || 0) +
+  (data?.pedning?.total || 0);
```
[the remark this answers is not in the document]
MariaAga: Yes and yes, fixed, thanks!

MariaAga:
@laviro can you take a look?

laviro:
Nice work @MariaAga,

> I meant doing it the same way as it is now, just adding it to the table instead of putting it below it. So you'd add just one row called "other"

I agree with @adamruzicka,
it would fit in more logically to the table.
Also I think that the users don't really want to know about the all history of "other" stopped tasks,
IMO it would be more relevant to show them also the 24th column.

thinking about it - should we also add in the future a column for "last week" / "month" / "year" ?

should we rename it from "Other" to "unknown" ? since the task is ended but its result is neither "success" / "warning" / "error"
for me it "unknown" sounds a bit better.

MariaAga:
Thanks @laviro for the detailed feedback! this design and text is approved by Rox and Melanie.

> Also I think that the users don't really want to know about the all history of "other" stopped tasks 

This PR is for a user wanting to know how many cancelled tasks there are

> IMO it would be more relevant to show them also the 24th column.

Because other includes different result types I don't think that a user will get any value from knowing how many 'other' result were there in the last 24h, but I do think the user should know that there are more result types than whats shown in the table.

> thinking about it - should we also add in the future a column for "last week" / "month" / "year" ?

last week already exists in the table, and I think that the number of tasks in a year for a user will be too big for a dashboard

> should we rename it from "Other" to "unknown" ? since the task is ended but its result is neither "success" / "warning" / "error"
for me it "unknown" sounds a bit better.

I don't think we should use unknown because we do know what the result is.

pondrejk (approved):
Checked from the user's pov, looks good

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
laviro, line 31: IMO it would be better to send the data correctly from the server, instead of manipulating it here.

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
```diff
@@ -44,16 +47,32 @@ const StoppedTasksCard = ({
...
+          <table className="table table-bordered table-striped stopped-table">
+            <thead>
+              <tr>
+                <th />
+                <th>{__('Total')}</th>
+                <th>{getQueryValueText(time)}</th>
+              </tr>
+            </thead>
+            <tbody>{StoppedTable(data, query, time, updateQuery)}</tbody>
+          </table>
```
laviro, line 60: This could go into its own component file.

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
```diff
@@ -44,16 +47,32 @@ const StoppedTasksCard = ({
...
+              onClick={() =>
+                updateQuery({
+                  state: TASKS_DASHBOARD_AVAILABLE_QUERY_STATES.STOPPED,
+                  result: TASKS_DASHBOARD_AVAILABLE_QUERY_RESULTS.OTHER,
+                })
+              }
+            >
+              {otherCount}
+            </Button>
+          </span>
```
laviro, line 74: Also could go into its own component file.

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardSelectors.js
```diff
@@ -12,6 +12,9 @@ export const selectTime = state =>
 
 export const selectQuery = state => selectTasksDashboard(state).query || {};
 
+export const calcStoppedOther = data =>
+  (data?.cancelled?.total || 0) + (data?.pending?.total || 0);
+
```
laviro, line 17: Can we get this data from the server?

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
MariaAga: What do you mean by getting it correctly? I can move the default `0` to the selector/ default props

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardSelectors.js
MariaAga: I think it will complicate the server more that this complicated the front end

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
laviro: the `data` object looks something like:
`data = { error, success, warning, other }`
instead of manipulating it here and later also using `calcStoppedOther` helper,
we could return from the server something like:
`data = { results: { error, success, warning }, otherCount: 0 }`

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardSelectors.js
laviro: commented about it also here: https://github.com/theforeman/foreman-tasks/pull/578#discussion_r477113509

MariaAga:
After moving the components around this became a bit of a refactor PR as well

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
```diff
@@ -44,16 +44,22 @@ const StoppedTasksCard = ({
...
+            query={query}
+            time={time}
+            updateQuery={updateQuery}
+          />
+          <OtherInfo
+            updateQuery={updateQuery}
+            otherCount={data.other}
+            active={
+              query.state === STOPPED &&
+              query.result === TASKS_DASHBOARD_AVAILABLE_QUERY_RESULTS.OTHER
```
laviro, line 59: Can you just pass the `query` to the component as you did on line 50? And calculate `active` there?

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardSelectors.js
laviro: We don't need this anymore :)

MariaAga: Why not?

laviro: Don't you pass it as otherCount from server?

MariaAga: Nope, the server returns {[result]:count} for every result in the DB

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/StoppedTasksCard.js
MariaAga: Done

webpack/ForemanTasks/Components/TasksDashboard/TasksDashboardSelectors.js
laviro: can you add a small test for `calcStoppedOther`?

laviro (commented):
Almost there !

Left some inline comments,
also, looks like the font is not on the same size?
Screenshot (6)

while at it, can you add a bit margin near the rest of the results icons?
to have the same space between the icon and the text as in `others`:

Screenshot (5)

MariaAga:
* Font is 12px in both.
* Added margin to the icons 
* Added test for `calcStoppedOther`

webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/OtherInfo.js
```diff
@@ -0,0 +1,53 @@
...
+      <Button
+        bsStyle="link"
+        onClick={() =>
+          updateQuery({
+            state: STOPPED,
+            result: OTHER,
+          })
+        }
+      >
+        {otherCount}
```
laviro, line 44: The span font-size is using `--pf-global--FontSize--sm` which is 0.875rem (14px)
while the button count text is 12px

webpack/ForemanTasks/Components/TasksDashboard/__tests__/TasksDashboardSelectors.test.js
```diff
@@ -55,6 +64,10 @@ const fixtures = {
   'should select tasks-summary': () => selectTasksSummary(state),
   'should select tasks-summary when state is empty': () =>
     selectTasksSummary({}),
+  'should calcStoppedOther': () =>
+    calcStoppedOther(
+      state.foremanTasks.tasksDashboard.tasksSummary.stopped.by_result
+    ),
```
laviro, line 70: I guess it would be fine to use:
```suggestion
    calcStoppedOther(selectTasksSummary(state)stopped.by_result)),
```

laviro: or just passing `by_result` from fixtures :)

MariaAga:
Same font for everyone :tada: 
Screenshot from 2020-09-03 13-16-41

laviro (approved):
Thanks for all of the work @MariaAga !
Works great :)

adamruzicka (approved):
Works as expected, ACK

adamruzicka:

PR #579, opened 2020-08-23
Fixes #30407 - removes unmaintained react-html-parser

MariaAga:
'react-html-parser' was causing issues when opening task details, and as it's not maintained any more we can use `dangerouslySetInnerHTML` instead.
The html should be safe as it comes from the server

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
   "dependencies": {
     "c3": "^0.4.11",
     "humanize-duration": "^3.20.1",
-    "react-html-parser": "^2.0.2",
     "react-intl": "^2.8.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


adamruzicka:
Do you recall why we went with `react-html-parser` in the first place here https://github.com/theforeman/foreman-tasks/pull/508 ?

MariaAga:
Because I wanted to go the extra careful way. But other libraries that do similar things are too bug imho

MariaAga:
ruby test fails are not relevant to the PR

adamruzicka:
ruby tests should be fixed by https://github.com/theforeman/foreman-tasks/pull/580

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @MariaAga !

PR #584, opened 2020-09-08
Fixes #30797 - tasks not taxable in filters

ezr-ondrej:
Tasks are falsely recognized as taxable by filters.

### DIFF
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -198,6 +198,10 @@ module ForemanTasks
       'ForemanTasks::Task'
     end
 
+    def self.allows_taxonomy_filtering?(_taxonomy)
+      false
+    end
+
     def add_missing_task_groups(groups)
       groups = [groups] unless groups.is_a? Array
       (groups - task_groups).each { |group| task_groups << group }
--- a/db/migrate/20200908120905_remove_taxonomy_from_task_filters.rb
+++ b/db/migrate/20200908120905_remove_taxonomy_from_task_filters.rb
@@ -1,0 +1,8 @@
+class RemoveTaxonomyFromTaskFilters < ActiveRecord::Migration[6.0]
+  def up
+    Permission.where(name: %w[view_foreman_tasks edit_foreman_tasks]).each do |perm|
+      TaxableTaxonomy.where(taxable_type: 'Filter', taxable_id: perm.filter_ids).delete_all
+      perm.filters.update_all(taxonomy_search: nil)
+    end
+  end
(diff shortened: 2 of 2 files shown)


adamruzicka:
Tasks may not be taxable directly, but we try to tax them using associated resources (mostly katello stuff), we even try to hide/show tasks on the index page based on current org & loc. It is on a best effort basis, but it is at least something. Could this have any impact on that?

Couldn't we (maybe not instead of this, but in addition to this) reshape the query it generates to turn it from `(type = something and id = 1) or (type = something and id = 2) or (type = something_else and id = 5)` to something like `(type = something and id in (1,2)) or (type = something_else and id ^ (5))` to cut the number of joins we have to make?

ezr-ondrej:
> Tasks may not be taxable directly, but we try to tax them using associated resources (mostly katello stuff), we even try to hide/show tasks on the index page based on current org & loc. It is on a best effort basis, but it is at least something. Could this have any impact on that?

Well the taxing we do now is definitelly wrong. Taxing it through the locks is far from good solution IMHO, but I agree, that this prevents the ability of other plugins to do so in better manner.
Although I would say if they wish to do so, I think, they can be explicit about it and override this method as well.

> Couldn't we (maybe not instead of this, but in addition to this) reshape the query it generates to turn it from `(type = something and id = 1) or (type = something and id = 2) or (type = something_else and id = 5)` to something like `(type = something and id in (1,2)) or (type = something_else and id ^ (5))` to cut the number of joins we have to make?

That's the right solution here, I'll take a look into it and hopefully do so in core ^_^

ezr-ondrej:
> Couldn't we (maybe not instead of this, but in addition to this) reshape the query it generates to turn it from `(type = something and id = 1) or (type = something and id = 2) or (type = something_else and id = 5)` to something like `(type = something and id in (1,2)) or (type = something_else and id ^ (5))` to cut the number of joins we have to make?

Oh, you're goofing with me, we(/you)'ve already done so in https://github.com/theforeman/foreman/pull/7553 so all what it takes is rebuild the taxonomy search for all filters.

adamruzicka:
I vaguely remember doing that, but since we were still hitting this I thought there was some other place where this needed to be done. Guess I missed the filter rebuild

ezr-ondrej:
So now the question is, should we disable the taxing of this anyway? As taxing it through the lock (on "lock is set on organization/location") doesn't make any sense to me. I'd even guess it can't work properly :thinking:

ezr-ondrej:
Closing in favor of #7995

PR #585, opened 2020-09-08
Fixes #30800 - Upgrades js tests to support jest 26

johnpmitsch:
Related to https://github.com/theforeman/foreman-js/pull/212

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "url": "http://projects.theforeman.org/projects/foreman-tasks/issues"
   },
   "peerDependencies": {
-    "@theforeman/vendor": ">= 4.0.2"
+    "@theforeman/vendor": ">= 6.0.0"
   },
   "dependencies": {
     "c3": "^0.4.11",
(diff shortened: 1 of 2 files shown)


adamruzicka:
Am I reading it right it requires theforeman/foreman-js#212 to make the tests green?

johnpmitsch:
> Am I reading it right it requires [theforeman/foreman-js#212](https://github.com/theforeman/foreman-js/pull/212) to make the tests green?

correct :+1: Once that is merged, the Foreman tests will break and this fixes it. One test may still need to be fixed, but looking into that.

adamruzicka:
The dependency in foreman was merged, could you please rebase so we get the tests running on latest and greatest?

johnpmitsch:
> The dependency in foreman was merged, could you please rebase so we get the tests running on latest and greatest?

Done!

adamruzicka:
I'm a bit foggy on how things work with foreman-js. Does the merge of the foreman-js dependency mean that PR tests will fail against latest foreman develop until this gets merged? Does merging this mean foreman-tasks master will become compatible with develop/2.4 only?

johnpmitsch:
> I'm a bit foggy on how things work with foreman-js. Does the merge of the foreman-js dependency mean that PR tests will fail against latest foreman develop until this gets merged? Does merging this mean foreman-tasks master will become compatible with develop/2.4 only?

It looks like foreman-tasks is on it's own foreman-js dependencies, so I believe in development it can move independently. With packaging, I'm not sure - since the foreman-js 6.0.0 packages are in packaging now, I think foreman-tasks has to update as these are the only builds available?

Since they are meant to be updated together, I can update everything to 6.0.0, there were no breaking changes other than the jest tests, does that sound ok?

adamruzicka:
Sounds good to me

johnpmitsch:
@adamruzicka done!

adamruzicka (approved):

adamruzicka:
Thank you @johnpmitsch !

PR #588, opened 2020-09-18
Allow export_tasks to skip tasks that fail to export

phess:
Allow foreman_tasks:export_tasks to keep going and skip any tasks that fail to export.
Also print a task.inspect of any failed task. (should this be done only if VERBOSE or DEBUG is set to true?)
Add the SKIP_FAILED env var with a default value of false to preserve the usual behavior of exploding when a task is broken.
Issue a suggestion to re-run with SKIP_FAILED=true in case it's not set and a task fails to export.
No redmine issue created for this; I hope this can be forgiven.

### DIFF
--- a/lib/foreman_tasks/tasks/export_tasks.rake
+++ b/lib/foreman_tasks/tasks/export_tasks.rake
@@ -14,6 +14,7 @@ namespace :foreman_tasks do
       * TASK_FILE       : file to export to
       * TASK_FORMAT     : format to use for the export (either html or csv)
       * TASK_DAYS       : number of days to go back
+      * SKIP_FAILED     : skip tasks that fail to export (true or false[default])
 
     If TASK_SEARCH is not defined, it defaults to all tasks in the past 7 days and
     all unsuccessful tasks in the past 60 days. The default TASK_FORMAT is html
@@ -26,6 +27,8 @@ namespace :foreman_tasks do
     deprecated_options.each do |option, new_option|
       raise "The #{option} option is deprecated. Please use #{new_option} instead" if ENV.include?(option.to_s)
     end
+
+    skip_errors = ENV['SKIP_FAILED'] == "true"
 
     class TaskRender
       def initialize
@@ -221,6 +224,15 @@ namespace :foreman_tasks do
         tasks.order('started_at desc').all.each do |task|
           html << "<tr><td><a href=\"#{task.id}.html\">#{task.label}</a></td><td>#{task.started_at}</td>\
                    <td>#{task.state}</td><td>#{task.result}</td></tr>"
+        rescue StandardError => e
+          if skip_errors
+            puts _('Skipping indexing a task that failed to export because SKIP_FAILED is set to "true"')
+          else
+            puts _('Indexing tasks failed because the task below failed to be indexed.')
+            puts _('Re-run with SKIP_FAILED=true if you wish to prevent this problem from halting task export altogether.')
+            puts task.inspect
+            raise e
+          end
         end
         html << '</table></div>'
       end
@@ -242,6 +254,7 @@ namespace :foreman_tasks do
     export_filename = ENV['TASK_FILE'] || "/tmp/task-export-#{Time.now.to_i}.#{format == 'csv' ? 'csv' : 'tar.gz'}"
 
     tasks = ForemanTasks::Task.search_for(filter)
+    exported_tasks = []
 
     puts _("Exporting all tasks matching filter #{filter}")
     puts _("Gathering #{tasks.count} tasks.")
@@ -254,10 +267,21 @@ namespace :foreman_tasks do
 
         tasks.each_with_index do |task, count|
           File.open(File.join(tmp_dir, "#{task.id}.html"), 'w') { |file| file.write(PageHelper.pagify(renderer.render_task(task))) }
+          exported_tasks << task
           puts "#{count + 1}/#{total}"
+        rescue StandardError => e
+          puts _("WARNING: task failed to export. Additional task details below.")
+          puts task.inspect
+          if skip_errors
+            # Displaying task.id may result in another failure in case the task is severely broken (as seen on the field).
+            puts _("WARNING: Skipping task #{count + 1} because it failed to export.")
+          else
+            puts _("Re-run with SKIP_FAILED=true if you want to simply skip task #{count + 1} and any other tasks that fail to export.")
+            raise e
+          end
         end
 
-        File.open(File.join(tmp_dir, 'index.html'), 'w') { |file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks))) }
+        File.open(File.join(tmp_dir, 'index.html'), 'w') { |file| file.write(PageHelper.pagify(PageHelper.generate_index(exported_tasks))) }
 
         system("tar", "czf", export_filename, tmp_dir)
       end
@@ -267,6 +291,16 @@ namespace :foreman_tasks do
         tasks.each do |task|
           csv << [task.id, task.state, task.type, task.label, task.result,
                   task.parent_task_id, task.started_at, task.ended_at]
+        rescue StandardError => e
+          puts _("WARNING: task failed to export. Additional task details below.")
+          puts task.inspect
+          if skip_errors
+            # Displaying task.id may result in another failure in case the task is severely broken (as seen on the field).
+            puts _("WARNING: Skipping a task because it failed to export.")
+          else
+            puts _("Re-run with SKIP_FAILED=true if you want to simply skip any tasks that fail to export.")
+            raise e
+          end
         end
       end
     end


lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -253,8 +255,20 @@ namespace :foreman_tasks do
...
 
         tasks.each_with_index do |task, count|
-          File.open(File.join(tmp_dir, "#{task.id}.html"), 'w') { |file| file.write(PageHelper.pagify(renderer.render_task(task))) }
-          puts "#{count + 1}/#{total}"
+          begin
+            File.open(File.join(tmp_dir, "#{task.id}.html"), 'w') { |file| file.write(PageHelper.pagify(renderer.render_task(task))) }
+            puts "#{count + 1}/#{total}"
+          rescue StandardError => e
+            puts "WARNING: task failed to export. Additional task details below."
+            puts "#{task.inspect}"
```
adamruzicka, line 263: `whatever.inspect` already returns a string, so wrapping it in another string and then using string interpolation is a bit redundant

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -253,8 +255,20 @@ namespace :foreman_tasks do
         total = tasks.count
 
         tasks.each_with_index do |task, count|
-          File.open(File.join(tmp_dir, "#{task.id}.html"), 'w') { |file| file.write(PageHelper.pagify(renderer.render_task(task))) }
-          puts "#{count + 1}/#{total}"
+          begin
```
adamruzicka, line 258: You can drop this `begin` and matching `end` and indent the block body one level less

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -265,8 +279,20 @@ namespace :foreman_tasks do
       CSV.open(export_filename, 'wb') do |csv|
         csv << %w[id state type label result parent_task_id started_at ended_at]
         tasks.each do |task|
-          csv << [task.id, task.state, task.type, task.label, task.result,
-                  task.parent_task_id, task.started_at, task.ended_at]
+          begin
```
adamruzicka, line 282: See previous comment about blocks

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -265,8 +279,20 @@ namespace :foreman_tasks do
...
+            csv << [task.id, task.state, task.type, task.label, task.result,
+                    task.parent_task_id, task.started_at, task.ended_at]
+          rescue StandardError => e
+            puts "WARNING: task failed to export. Additional task details below."
+            puts "#{task.inspect}"
```
adamruzicka, line 287: See previous comment about `.inspect`

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -240,6 +241,7 @@ namespace :foreman_tasks do
 
     format = ENV['TASK_FORMAT'] || 'html'
     export_filename = ENV['TASK_FILE'] || "/tmp/task-export-#{Time.now.to_i}.#{format == 'csv' ? 'csv' : 'tar.gz'}"
+    skip_errors = ENV['SKIP_FAILED']
```
adamruzicka, line 244: How about you compare it with `"true"` here so `skip_errors` is always a boolean and not `nil` or `String`?

adamruzicka (commented):
Haven't tested yet, left some nitpicks

lib/foreman_tasks/tasks/export_tasks.rake
phess: Thank you so much for the "nitpicks" @adamruzicka !   I appreciate them.  I think they're all resolved i.e. applied now.

phess:
@adamruzicka , please let me know if anything else is needed. I believe all issues you mentioned are resolved. Thanks!

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -255,6 +257,16 @@ namespace :foreman_tasks do
         tasks.each_with_index do |task, count|
           File.open(File.join(tmp_dir, "#{task.id}.html"), 'w') { |file| file.write(PageHelper.pagify(renderer.render_task(task))) }
           puts "#{count + 1}/#{total}"
+        rescue StandardError => e
+          puts "WARNING: task failed to export. Additional task details below."
+          puts task.inspect
+          if skip_errors?
```
adamruzicka, line 263: the variable is just `skip_errors` without the question mark

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -267,6 +279,16 @@ namespace :foreman_tasks do
         tasks.each do |task|
           csv << [task.id, task.state, task.type, task.label, task.result,
                   task.parent_task_id, task.started_at, task.ended_at]
+        rescue StandardError => e
+          puts "WARNING: task failed to export. Additional task details below."
+          puts task.inspect
+          if skip_errors?
```
adamruzicka, line 285: the variable is just `skip_errors` without the question mark

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -255,6 +257,16 @@ namespace :foreman_tasks do
...
+            # Displaying task.id may result in another failure in case the task is severely broken (as seen on the field).
+            puts "WARNING: Skipping task #{count + 1} because it failed to export."
+          else
+            puts "Re-run with SKIP_FAILED=true if you want to simply skip task #{count + 1} and any other tasks that fail to export."
+            raise e
+          end
         end
 
         File.open(File.join(tmp_dir, 'index.html'), 'w') { |file| file.write(PageHelper.pagify(PageHelper.generate_index(tasks))) }
```
adamruzicka, line 272: Even if we skip the failed tasks, they appear in the index generated here, but the index may link to files which are not there because they failed to export

adamruzicka (changes requested):
After some minor adjustments it works well, however there is an issue with html index.

Also, could you wrap the messages you're printing with `_()` so those strings can get extracted for translation?

lib/foreman_tasks/tasks/export_tasks.rake
adamruzicka: And possibly if even `task.id` would raise an exception as suggested here https://github.com/theforeman/foreman-tasks/pull/588/files#diff-8898a6eafc1346188e475c0a876c1aa6R264 then the index generation would fail

phess:
Thanks for your latest notes about implementing _() and the correct way to test skip_errors, @adamruzicka . They are now implemented.

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -221,6 +222,15 @@ namespace :foreman_tasks do
         tasks.order('started_at desc').all.each do |task|
           html << "<tr><td><a href=\"#{task.id}.html\">#{task.label}</a></td><td>#{task.started_at}</td>\
                    <td>#{task.state}</td><td>#{task.result}</td></tr>"
+        rescue StandardError => e
+          if skip_errors
```
adamruzicka, line 226: This runs in a different context than where `skip_errors` is defined

```
NameError: undefined local variable or method `skip_errors' for PageHelper:Class

/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:227:in `rescue in block in generate_index'
/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:223:in `block in generate_index'
/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:222:in `generate_index'
/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:282:in `block (4 levels) in <main>'
/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:282:in `open'
/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:282:in `block (3 levels) in <main>'
/home/aruzicka/projects/foreman/foreman-tasks/lib/foreman_tasks/tasks/export_tasks.rake:261:in `block (2 levels) in <main>'
```

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -221,6 +222,15 @@ namespace :foreman_tasks do
...
+            puts _('Skipping indexing a task that failed to export because SKIP_FAILED is set to "true"')
+          else
+            puts _('Indexing tasks failed because the task below failed to be indexed.')
+            puts _('Re-run with SKIP_FAILED=true if you wish to prevent this problem from halting task export altogether.')
+            puts task.inspect
```
adamruzicka, line 231: Do we expect that there could be a task which would pass the regular export but would fail when generating the index (or the other way around)? If not, then we would be logging basically the same failure twice

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -255,6 +266,16 @@ namespace :foreman_tasks do
         tasks.each_with_index do |task, count|
```
adamruzicka, line 266: How about we collect the tasks which were successfully exported here and then pass those to the function which builds the index?

adamruzicka (changes requested):

lib/foreman_tasks/tasks/export_tasks.rake
phess: That's a great idea. I'll work on this now. Thanks!

phess: Done. I added an exported_tasks array where I pushed each successfully exported task into, and I'm now passing generate_index() the new exported_tasks array instead of the original tasks array (or array-like structure).

Looking forward to your review. Thanks again for this, @adamruzicka !

lib/foreman_tasks/tasks/export_tasks.rake
phess: Hopefully fixed this with the latest commit.

lib/foreman_tasks/tasks/export_tasks.rake
phess: This is resolved by creating an exported_tasks array that is appended only the tasks that are successfully exported, then passing generate_index() this new array instead of the original one.
I expect this to fix the issue of generate_index() failing.

lib/foreman_tasks/tasks/export_tasks.rake
phess: Now that generate_index() will only receive successfully exported tasks, I think the `rescue` clause here is OK, too.

lib/foreman_tasks/tasks/export_tasks.rake
phess: Resolved.

lib/foreman_tasks/tasks/export_tasks.rake
phess: resolved.

lib/foreman_tasks/tasks/export_tasks.rake
phess: resolved.

lib/foreman_tasks/tasks/export_tasks.rake
phess: resolved.

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -221,6 +224,15 @@ namespace :foreman_tasks do
         tasks.order('started_at desc').all.each do |task|
```
adamruzicka, line 224: Before this change `tasks` was a `ForemanTasks::Task::ActiveRecord_Relation`, now it is an `Array` of `ForemanTasks::Task`, which does not implement `order`

```
NoMethodError: undefined method `order' for #<Array:0x000000000b043480>
lib/foreman_tasks/tasks/export_tasks.rake:224:in `generate_index'
lib/foreman_tasks/tasks/export_tasks.rake:284:in `block (4 levels) in <main>'
lib/foreman_tasks/tasks/export_tasks.rake:284:in `open'
lib/foreman_tasks/tasks/export_tasks.rake:284:in `block (3 levels) in <main>'
lib/foreman_tasks/tasks/export_tasks.rake:262:in `block (2 levels) in <main>'
```

adamruzicka (changes requested):
This isn't going to fly. Even if we solved the ordering issue, we would still accumulate *a lot* of tasks, keeping them in memory until the index is generated.

How about we:
1) Open a file for the index, write in the header
2) Order all the tasks
3) Export them one by one and also write out an index entry while exporting the task
4) Once all the tasks are exported, finish up the index and close the file.

This way we wouldn't have to keep tasks around and we'd do everything in a more stream-like fashion

adamruzicka:
I made some changes which should make https://github.com/theforeman/foreman-tasks/pull/588#pullrequestreview-561784073 easier here https://github.com/theforeman/foreman-tasks/pull/602

adamruzicka:
https://github.com/theforeman/foreman-tasks/pull/602 finally went in so this should be easier, but will need some more changes. @phess do you want to continue with this or should I take over?

phess:
> #602 finally went in so this should be easier, but will need some more changes. @phess do you want to continue with this or should I take over?

Thanks for your offer, @adamruzicka. Please do take over. Let me know if any action required from me in order for you to take over.

adamruzicka:
Closing in favor of https://github.com/theforeman/foreman-tasks/pull/645 . Thank you @phess

PR #589, opened 2020-10-05
Fixes #30969 - replace deprecated mount_react_component

MariaAga:
https://github.com/theforeman/foreman/pull/7874
`mount_react_component` should be replaced by `react_component`

### DIFF
--- a/app/views/foreman_tasks/tasks/show.html.erb
+++ b/app/views/foreman_tasks/tasks/show.html.erb
@@ -15,9 +15,4 @@
   switcher_item_url: foreman_tasks_task_path(:id => ':id')
 ) %>
 
-
-<div class="task-details" id="foremanTaskDetails">
-</div>
-
-
-<%= mount_react_component('TaskDetails', '#foremanTaskDetails')  %>
+<%= react_component('TaskDetails')  %>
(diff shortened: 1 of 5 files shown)


app/views/foreman_tasks/tasks/show.html.erb
```diff
@@ -15,9 +15,4 @@
   switcher_item_url: foreman_tasks_task_path(:id => ':id')
 ) %>
 
-
-<div class="task-details" id="foremanTaskDetails">
```
ezr-ondrej, line 19: We've got some css styles on `task-details` namely `.task-details pre { ... }`.
I'd choose an easy option and just leave the div, without the `id=""` part and put the mount inside.
But we can change css, or add the class into the component itself.

ezr-ondrej (changes requested):
Thanks @MariaAga ! I believe we need to keep the `task-details` class around though.

app/views/foreman_tasks/tasks/show.html.erb
MariaAga: I changed the css so it will be more specific, thanks for the suggestion!

ezr-ondrej (approved):
Thanks @MariaAga :+1:  Nice css cleanup ^_^

ezr-ondrej:
@adamruzicka could you press the button please? This works nicely and I believe this is ready to get in :)

adamruzicka:
With pleasure

adamruzicka (approved):

adamruzicka:
Thank you @MariaAga & @ezr-ondrej !

PR #590, opened 2020-10-12
fixes #31457 - Adding method to Task jail

domitea:
This PR is consist of adding `action_output` to Task model and also is added to jail for usage in reports. This methods acts as proxy method for `to_json_output` in Continuous output class. These methods are requested by another PR in REX (https://github.com/theforeman/foreman_remote_execution/pull/453).

### DIFF
--- a/app/models/foreman_tasks/task.rb
+++ b/app/models/foreman_tasks/task.rb
@@ -82,7 +82,7 @@ module ForemanTasks
       property :ended_at, ActiveSupport::TimeWithZone, desc: 'Returns date with time the task ended at'
     end
     class Jail < Safemode::Jail
-      allow :started_at, :ended_at, :result, :state, :label, :main_action
+      allow :started_at, :ended_at, :result, :state, :label, :main_action, :action_output
     end
 
     def input
@@ -240,6 +240,12 @@ module ForemanTasks
         end
       end.join('; ')
       parts.join(' ').strip
+    end
+
+    def action_output
+      return unless main_action.is_a?(Actions::Helpers::WithContinuousOutput)
+      main_action.continuous_output.sort!
+      main_action.continuous_output.raw_outputs
     end
 
     protected


lib/foreman_tasks_core/continuous_output.rb
```diff
@@ -46,5 +46,14 @@ def self.format_output(message, type = 'debug', timestamp = Time.now.getlocal)
         'output' => message,
         'timestamp' => timestamp.to_f }
     end
+
+    def to_json_output
+      sort!
+      outputs_json = { 'outputs': [] }
+      raw_outputs.map do |output|
```
adamruzicka, line 53: `Enumerable.map` returns an array, using it for side effects isn't exactly best practice.

I'd go with either
```ruby
outputs = raw_outputs.map { |output| { type: output[:output_type], value: output[:message] } }
{ 'outputs': outputs }
```

or just swap `map` for `each` if you want to go the mutating way

domitea: Thanks for the tip! I've updated that.

domitea:
I've made some improvements there. Also I've created a separate PR for template (https://github.com/theforeman/foreman/pull/8178) that is pointing to core.

ezr-ondrej (changes requested):
The referenced RM issue is from another project, can we please use: https://projects.theforeman.org/issues/31457 ?
Also, the new method feels somewhat uneccessary now, isn't it?

app/models/foreman_tasks/task.rb
```diff
@@ -242,6 +242,10 @@ def to_label
       parts.join(' ').strip
     end
 
+    def action_output
+      main_action.continuous_output.raw_outputs.sort! if main_action.is_a?(Actions::Helpers::WithContinuousOutput)
```
ezr-ondrej, line 246:
```suggestion
      return unless main_action.is_a?(Actions::Helpers::WithContinuousOutput)
      main_action.continuous_output.sort!
      main_action.continuous_output.raw_outputs
```

ezr-ondrej (approved):
Thanks @domitea ! Looks ok to me :+1:

adamruzicka (approved):
ACK

adamruzicka:
Thank you @domitea & @ezr-ondrej !

PR #592, opened 2020-10-22
Fixes #30161 - Make KeepCurrentUser cover entire transaction in finalize

adamruzicka:


adamruzicka:
@ezr-ondrej opinions? We're trying to solve this https://github.com/Katello/katello/pull/8783#issuecomment-713494329

app/lib/actions/middleware/keep_current_user.rb
```diff
@@ -19,6 +19,10 @@ def finalize
         restore_curent_user { pass }
       end
 
+      def finalize_phase(execution_plan, *args)
+        restore_curent_user(execution_plan.entry_action) { pass(execution_plan, *args) }
```
ezr-ondrej, line 23: Why are we passing a different action here?

lib/foreman_tasks/engine.rb
```diff
@@ -108,7 +108,7 @@ class Engine < ::Rails::Engine
       ForemanTasks.dynflow.require!
       ::ForemanTasks.dynflow.config.on_init(false) do |world|
         world.middleware.use Actions::Middleware::KeepCurrentTaxonomies
-        world.middleware.use Actions::Middleware::KeepCurrentUser
+        world.middleware.use Actions::Middleware::KeepCurrentUser, :before => ::Dynflow::Middleware::Common::Transaction
```
ezr-ondrej, line 111: This should solve the referenced Katello issue, right? or am I missing somethin?

app/lib/actions/middleware/keep_current_user.rb
adamruzicka: The `finalize_phase` method of the middleware is run only once around the entire phase, while regular `finalize` is run around each step in finalize. Because it is run around the entire phase, it doesn't get a specific action as an argument so we just pick the entry action of the execution plan and hope for the best. There are two issues with this approach (entry action not having current user set, each action setting its own current user), but the first one shouldn't happen and the best we can do about the other is strongly discourage people from doing so.

lib/foreman_tasks/engine.rb
adamruzicka: Yes, together with the extended period when current user is set

app/lib/actions/middleware/keep_current_user.rb
ezr-ondrej: What about deprecating it? :shrug:
If we can detect it, we can deprecate and remove in few versions, if noone complains? :thinking:

adamruzicka: Deprecating what exactly?

ezr-ondrej: `user_id` in this middleware. now it is being used to set user for every step, but we can deprecated that, right?

ezr-ondrej: But if we keep it as is, it will just override the user for the specific action, right? and if there is no :current_user_id, no user is set, that seems reasonable to me :shrug:

adamruzicka: > But if we keep it as is, it will just override the user for the specific action, right?

It will, but then the katello issue could get a new twist
1) Finalize phase and transaction starts, user A is set
2) nth step in finalize phase sets its user to B and saves an ActiveRecord object
3) Finalize phase ends, transaction ends, after_commit hooks are executed, but with user A

This is a hypothetical scenario, I don't think it could happen with the action code we currently have.

ezr-ondrej: I think it would be enough to document what can happen, but still we can try to detect such scenarios and warn about them, right?

jturel:
Please link this PR to https://bugzilla.redhat.com/show_bug.cgi?id=1796770 via redmine when it's ready to go :)

app/lib/actions/middleware/keep_current_user.rb
adamruzicka: In the end I went with a deprecation warning

adamruzicka:
@ezr-ondrej anything to add here?

ezr-ondrej (approved):
I'm good, I didn't hit the original issue, but this seems to be a reasonable fix :+1:

adamruzicka:
Thank you @ezr-ondrej !

PR #596, opened 2020-11-11
Fixes #27295 - Task widgets don't show time correctly

MariaAga:
date_time_relative is a function that loads a react component so translating it causes showing the html instead of rendering the component


adamruzicka (approved):
Fixes the issue for me. Since `date_time_relative` mounts a component, it is that component's responsibility to do the translation.

adamruzicka:
Thank you @MariaAga !

PR #597, opened 2020-11-24
Fixes #31388 - fix disable on rec logic with deleted task

jlsherrill:


jlsherrill:
The only way i could reproduce the user's issue was by manually deleting the ForemanTask associated with the recurring logic.  

The result with this change is that you can now disable the recurring logic, but any scheduled dynflow tasks will still be scheduled.  I'm not sure there's anything we can do about that?

chris1984 (approved):
ACK, looks sane, since you could only reproduce it that way, I assume this will be an edge case?

adamruzicka:
> but any scheduled dynflow tasks will still be scheduled. 

There should be no scheduled tasks associated with the recurring logic in the case you're addressing. Or was there a context switch that I missed?

jlsherrill:
@adamruzicka i think the method to reproduce was:

1.  create a sync plan (which creates a recurring logic and foreman-task and dynflow action)
2.  Manually delete the ForemanTask::Task object, now we have a recurring logic, and a dynflow task
3.  try to disable the recurring logic


This would error.  With this fix it will no longer error, but because the 'link' between the recurring logic and the dynflow task has been deleted, the dynflow task will still hang around.  I don't think there's any way to correct that here

adamruzicka:
Ahh, ok, I got confused by the terminology. Fine by me then

jlsherrill:
@adamruzicka can i get a merge? :)

adamruzicka (approved):

adamruzicka:
Thank you @jlsherrill & @chris1984 !

PR #599, opened 2020-12-14
Fixes #31508 - Restore changed bulk resume api

adamruzicka:
We changed an existing API to start require parameters it previously didn't require. This reverts the API to the previous state.

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -79,7 +79,7 @@ module ForemanTasks
       param :task_ids, Array, :desc => N_('Resume specific tasks by ID')
       def bulk_resume
         if params[:search].nil? && params[:task_ids].nil?
-          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+          params[:search] = 'state = paused and result = error'
         end
         resumed = []
         failed = []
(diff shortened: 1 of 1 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -79,7 +79,7 @@ def bulk_search
       param :task_ids, Array, :desc => N_('Resume specific tasks by ID')
       def bulk_resume
         if params[:search].nil? && params[:task_ids].nil?
-          raise BadRequest, _('Please provide at least one of search or task_ids parameters in the request')
+          params[:search] = 'state = paused and result = error'
```
ezr-ondrej, line 82: I don't like writing to params, but for the simplicity let's roll with it :+1:

ezr-ondrej (commented):
Should we do something similar for bulk_cancel/stop? I'd say no, as it is dangerous in comparison to resume everything, so just asking.
If not, **should we state something about the default in the enpoint/param documentation**?
If you read the bulk documentation they seem to suggest it will be on all cancellable, all resumable, all stoppable, but only resume will actually work like that now.

adamruzicka:
The change which broke the original behavior aimed to make all the `bulk_*` behave similarly by requiring either a search query or explicit list of task ids. Even though this fix breaks the "all endpoints behave the same way" idea, I'd say it is still better than making all endpoints behave the old way (and breaking the established behavior for new endpoints).

Added some more apipie docs, hope it is better now.

ezr-ondrej (approved):
Great! Thanks @adamruzicka ! :+1:

adamruzicka:
Thank you @ezr-ondrej !

PR #600, opened 2020-12-20
Fixes #31534 - tasks search button should always reload data

MariaAga:
On the "Monitor" -> "Tasks" page, if the user searches for a word, and then clicks the "search" button a second time to update the results, it does nothing.


adamruzicka (changes requested):
Changing the search now does not update the url so if you do a full refresh you'll lose what you were searching for

adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @MariaAga !

PR #601, opened 2021-01-05
Rake task interface scripts

adamruzicka:


adamruzicka:
@ezr-ondrej , @ares : opinions?

ezr-ondrej:
I don't get this, why is this better then the rake itself? it's just one more script to maintain :thinking:
Generally I don't like the odds of this being maintained over time, as I dont like plain shell scripts, but if there is good reason someone might want that, I'm not against.
We are just saying hey we have rake, that is cool cmd tool, but here is another script, because we know the rake is unusable IRL xD

adamruzicka:
This is just me scratching my own itch. What I don't like about the rake task we have:
- getting help is hard. Is it `foreman-rake foreman_tasks:cleanup --help`? `foreman-rake --help foreman_tasks:cleanup`? `foreman-rake help foreman_tasks:cleanup`? No, it is `foreman-rake -D foreman_tasks:cleanup`
- getting help takes ages, especially on production machines with `foreman-rake`, up to the point where looking into the source is faster.
- the env variables are not consistent (is it `STATES` or `TASK_STATES`? `SEARCH` or `TASK_SEARCH`?)
- afaik different shells deal with environment variables differently (although this might be something foreman-rake causes by sudo/su+scl magic). In some shells `TASK_SEARCH=something foreman-rake ...` works, in some the env vars have to come at the end

I've watched people struggle with using tasks cleanup way too many times. The rake task is not really intuitive and is hard to use right. 

> We are just saying hey we have rake, that is cool cmd tool, but here is another script, because we know the rake is unusable IRL

It is more of a "we have rake, which does great things, but we are aware it has really poor interface, so here's another tool, which is just a sane interface to rake task"

While we could address some of the pain points in the rake tasks themselves (and in foreman-rake possibly), it would take much more time and I'm afraid it would never reach the same ease of use.

ezr-ondrej (commented):
I'm basically fine with this, no strong opinion eighter way, if you believe this helps us and want to maintain this, lets get it in. I just don't want to maintain it much, that is my main issue here xD
So are you ready to maintain it if we pronounce it supported?

ezr-ondrej:
@ares care to weigh in? code-wise I'm ok with this, it's just another interface to the same thing and we'll need to maintain both. I've not done enough support to be able to say if it's worth the overhead to have nicer interface for export/cleanup of tasks

adamruzicka:
For the sake of completeness, a long time ago I put together https://github.com/theforeman/foreman_maintain/pull/182 to have this functionality included in foreman-maintain. It didn't gain much traction so some years after that I opened this one, which didn't gain much traction either. If we were to pick one of the two, I'd like to keep this one. It works on all platforms, is faster and is not tied to any specific configuration/installed packages of the machine you run it on. Otoh fm is the de facto go-to tool for all maintenance operations so maybe having it there makes more sense.

ezr-ondrej:
Well I'd be more inclined to have it in FM, as here we would have additional issue of getting in in `$PATH`, so it would require some packaging mojo as without it for most users it would be burried in gem folder, without much use.

adamruzicka:
The changes in f-m add only interface for cleanup, not for export.

upadhyeammit:
@ezr-ondrej the changes in f-m are only for cleanup and not the export, I am inclined to merge this instead. Accordingly should I close the f-m pr?

ezr-ondrej (approved):
I'm not bloking this from getting in, I just don't want to be taking care of it, but @adamruzicka if you want like it more, I can see why (rake tasks are not great) and just hit the button :)

adamruzicka:
Rebased on latest master, adding `html-dir` to supported export formats

adamruzicka:
I'm fine with carrying this, let's get it in

PR #602, opened 2021-01-06
Tweaks to task export

adamruzicka:
HTML export used to be a two pass process. First all the tasks would be exported one by one, then all the tasks would be ordered differently, loaded again and the index would be generated. With this change, the index is being generated as tasks are being exported, removing the need for a second pass.

Also introduces a `html-dir` export format, which is the same as `html`, but produces a directory instead of a `.tar.gz`. If `$TASK_FILE` is set, it is used as a directory for the export. Adding it as a separate output format is a bit unfortunate, but it felt like the most sensible way without introducing a breaking change.

### DIFF
--- a/lib/foreman_tasks/tasks/export_tasks.rake
+++ b/lib/foreman_tasks/tasks/export_tasks.rake
@@ -239,6 +239,40 @@ namespace :foreman_tasks do
       end
     end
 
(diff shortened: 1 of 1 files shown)


ezr-ondrej (approved):
Works smoothly, thanks @adamruzicka !

PR #605, opened 2021-01-19
Fixes #31677 - Emit events when tasks stop 

adamruzicka:
CC @lzap


app/lib/actions/observable_action.rb
```diff
@@ -0,0 +1,37 @@
...
+  module ObservableAction
+    def self.included(base)
+      base.include ::Foreman::Observable
+      base.execution_plan_hooks.use :emit_event, :on => :stopped
+
+      def base.event_name
+        # actions/remote_execution/run_host_job_stopped
+        ::Foreman::Observable.event_name_for(to_s.underscore + '_stopped')
+      end
+    end
```
lzap, line 11: I understand that this means a task was stopped which can mean it was a success or it is in stopped state until someone unpause it. I was wondering what events are available and if we should also emit some of the other ones?

adamruzicka: So far it emits the event only when the task goes into stopped state (stopped for good, no pausing, resuming or anything). It doesn't necessarily mean it was a success, it just means the task will not run anymore. If it was a success or not can be read from the payload.

We could emit an event each time a task goes into a different state (one of pending, scheduled, planning, planned, running, paused, stopped) or on success (state stopped, result success) or on failure (state stopped, result error). I can add whatever you feel is useful, I just didn't want to overdo it from the start.

lzap (approved):
Works fine! I tested with ReX patch you sent me, please open a PR for ReX.

One nitpick: The event name is a bit weird it contain Ruby namespace.

    2021-01-26T17:45:01 [I|app|] ForemanWebhooks::EventSubscriber: actions/remote_execution/run_hosts_job_stopped.event.foreman.event.foreman event received

I guess it's the same for ActiveRecord STI (host), so not something we should be solving here but in core. But maybe if you could convert `/` to `.` that would look nice from the day one. I will do a patch into core to do the same conversion but I won't backport that.

I just noticed the "event" is present twice, maybe to remove the "event" in here? It's weird.

lzap:
Tests are good, RM issue is a wrong project but I don't mind. Don't have push permissions tho.

app/lib/actions/observable_action.rb
lzap: I wonder if makes sense to emit _processed or _finished when it was a success. This way all webhook users will need to do if-statement in the ERB, it could be cleaner if we'd do this for them here. I am not sure how much work is this to implement, e.g. for ActiveRecord it's possible to pass options in the form of `:if => { # condition }`.

Otherwise let's just merge.

lzap: I understand, I just think that requiring users to do processing of the payload everytime when they just want to subscribe to "success" tasks would be nice.

Would you mind doing `_finished` or something like that and then `_stopped` for all the others?

adamruzicka:
> Would you mind backporting this into 2.3?

Not at all, I was counting on it.

> What else are we missing for Katello? The same change as for ReX needs to be done there?

Exactly, just pick the interesting actions and include `Actions::ObservableAction` into them

app/lib/actions/observable_action.rb
adamruzicka: Went with `${action}_succeeded` for now

adamruzicka:
> But maybe if you could convert / to . that would look nice from the day one.

Done

> I just noticed the "event" is present twice, maybe to remove the "event" in here?

That was an oversight on my side. Fixed

app/lib/actions/observable_action.rb
```diff
@@ -0,0 +1,82 @@
...
+    #   "ended_at"=>Tue, 15 Dec 2020 12:14:48 UTC +00:00,
+    #   "state"=>"stopped",
+    #   "result"=>"error",
+    #   "external_id"=>"66e2781a-379f-430b-8a2b-31a5035ec026",
+    #   "parent_task_id"=>"21a4539a-76a1-404c-b98b-0ee86b169e74",
+    #   "start_at"=>Tue, 15 Dec 2020 12:14:48 UTC +00:00,
+    #   "start_before"=>nil,
+    #   "action"=>"Remote action: Run ls -la / on allen-touhy.logan-minteer.rene-henandez.example.net",
+    #   "user_id"=>4,
+    #   "state_updated_at"=>Tue, 15 Dec 2020 12:14:48 UTC +00:00 }
```
lzap, line 77: I've pulled this whole example into the documentation PR in core. What are possible result values? I'd rather show successful example but unsure if that's `"result"=>"success"` or something else.

app/lib/actions/observable_action.rb
```diff
@@ -0,0 +1,82 @@
...
+      def event_names
+        [event_name_base + '_' + event_name_suffix(:success)]
+      end
+
+      def namespaced_event_names
+        event_names.map { |e| ::Foreman::Observable.event_name_for(e) }
+      end
+
+      def event_name_base
+        to_s.underscore.tr('/', '.')
```
lzap, line 47: Thanks, I actually opened a PR into core doing the same thing but that one might not get backported. Let's see.

lzap:
Looks good, let me try this.

lib/foreman_tasks/event_subscribers_extensions.rb
```diff
@@ -0,0 +1,11 @@
...
+  module EventSubscribersExtensions
+    def all_observable_events
+      @all_observable_events ||=
+        begin
+          actions = ::Dynflow::Action.descendants.select { |klass| klass <= ::Actions::ObservableAction }.map(&:namespaced_event_names)
+          (super + actions).flatten.uniq
+        end
+    end
+  end
+end
```
lzap, line 11: Can you move this to `engine.rb` and use the newly created `extend_observable_events`? You will need to checkout https://github.com/theforeman/foreman/pull/8284 for that.

app/lib/actions/observable_action.rb
adamruzicka: The result should always be one of `cancelled`, `error`, `warning`, `success` or `pending`. For the docs, `"result" => "success"` sounds like the way to go

lzap (approved):
There is an extra empty space in `engine.rb` but nothing important.

LGTM I am going to test this today/tomorrow.

lzap:
FYI there are some test failures, feel free to fix them before I test this.

adamruzicka:
And this in turn depends on https://github.com/theforeman/foreman/pull/8284 , tests won't turn green before that is merged.

ezr-ondrej:
> And this in turn depends on [theforeman/foreman#8284](https://github.com/theforeman/foreman/pull/8284) , tests won't turn green before that is merged.

So let's merge it? :)

lzap:
Let me test this first, on it.

lzap (changes requested):
In order to test this, I need to see _failed events but for some reason I only see _succeeded one. Looks like it does not register failure case.

adamruzicka:
> In order to test this, I need to see _failed events but for some reason I only see _succeeded one. Looks like it does not register failure case.

Wait, didn't we decide to emit events only on success by default?

lzap (approved):
It works fine. I am going to file a PR with an example template:

```erb
# Task main action <%= @object.task.main_action %>
# Run job host name <%= @object.host_name %>
# Run job host id <%= @object.host_id %>
# Run job host <%= @object.host %>
# Run job job_invocation_id <%= @object.job_invocation_id %>
# Run job job_invocation <%= @object.job_invocation %>
# Task label <%= @object.task.label %>
# Task started at <%= @object.task.started_at %>
# Task ended at <%= @object.task.ended_at %>
# Task resulted with <%= @object.task.result %>
# Task state <%= @object.task.state %>
# Task action output <%= @object.task.action_output %>
```

Output (ignore the empty lines - a bug in shellhooks):

```
2021-02-09T10:53:58 b9579e97 [I] Finished POST /shellhook/print_body with 200 (1.09 ms)
2021-02-09T10:53:58  [D] close: 127.0.0.1:47438
2021-02-09T10:53:58 b9579e97 [I] [18535] Started task ["/home/lzap/work/smart_proxy_shellhooks/examples/print_body"]
2021-02-09T10:53:58 b9579e97 [D] [18535] # Task main action #<Actions::RemoteExecution::RunHostJob:0x00007fff8aac5f80>

2021-02-09T10:53:58 b9579e97 [D] [18535] # Run job host name centos7-default.dummy.lan

2021-02-09T10:53:58 b9579e97 [D] [18535] # Run job host id 271

2021-02-09T10:53:58 b9579e97 [D] [18535] # Run job host centos7-default.dummy.lan

2021-02-09T10:53:58 b9579e97 [D] [18535] # Run job job_invocation_id 22

2021-02-09T10:53:58 b9579e97 [D] [18535] # Run job job_invocation Run echo hello

2021-02-09T10:53:58 b9579e97 [D] [18535] # Task label Actions::RemoteExecution::RunHostJob

2021-02-09T10:53:58 b9579e97 [D] [18535] # Task started at 2021-02-09 10:53:55 +0100

2021-02-09T10:53:58 b9579e97 [D] [18535] # Task ended at 2021-02-09 10:53:58 +0100

2021-02-09T10:53:58 b9579e97 [D] [18535] # Task resulted with success

2021-02-09T10:53:58 b9579e97 [D] [18535] # Task state stopped

2021-02-09T10:53:58 b9579e97 [D] [18535] # Task action output [{"output_type"=>"stdout", "output"=>"hello\r\n", "timestamp"=>1612864438.0971014}, {"output_type"=>"stdout", "output"=>"Exit status: 0", "timestamp"=>1612864439.0971014}]
```

lzap:
Some tests failed, also RM issue problem.

app/lib/actions/observable_action.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end
+
+    def self.included(base)
+      base.extend ClassMethods
+      base.include ::Foreman::Observable
+      base.execution_plan_hooks.use :emit_event, :on => :success
```
ezr-ondrej, line 54: If se use the hook only on success, will the `*_failed` events ever be emitted?

adamruzicka: Not by default, We decided to emit events on success by default, however action authors can decide to emit events also on different execution plan state transitions. Some parts of the code are there just in case someone does that

ezr-ondrej (approved):
Make sense :+1:

app/lib/actions/observable_action.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      trigger_hook "#{self.class.event_name_base}_#{self.class.event_name_suffix(hook)}",
+                   payload: event_payload(execution_plan)
+    end
+
+    def event_payload(_execution_plan)
+      { object: self }
+    end
+
+    class Jail < Safemode::Jail
```
ofedoren, line 68: Actually it would be nice to add DSL docs here as we do for all `jail` methods.

I'd recommend to use `webhooks` section with the name of a class such as `Action` or the actual class which accepts `task` method.

ares: sounds good, @adamruzicka do you plan to add this in some separate PR? or who could add it?

adamruzicka: It is on my todo list, but not at the top. Created https://projects.theforeman.org/issues/31953 so it doesn't get lost. I definitely wouldn't mind if anyone got to it before I do :)

PR #606, opened 2021-02-01
Fixes #31773 - Fix pg syntax errors when searching by taxonomy

adamruzicka:


test/unit/task_test.rb
```diff
@@ -129,6 +129,16 @@ class TasksTest < ActiveSupport::TestCase
         _ { proc { ForemanTasks::Task.search_for('duration = "25 potatoes"') } }.must_raise ScopedSearch::QueryNotSupported
       end
     end
+
+    context 'by taxonomies' do
+      test 'can search by taxonomies using IN' do
+        # These raise an exception if it does not work
+        # ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near "'1'"
+        # LINE 3: ...foreman_tasks_links_taxonomy7a7295.resource_id IN '1' OR for...
```
ezr-ondrej, line 137: Should we use `assert_nothing_raised` instead of the comment here?
We can have details of the exception in commit description, but I don't think it belongs to a comment :thinking:

ezr-ondrej (commented):
Just a test nit, but the change works :+1:

app/models/foreman_tasks/task/search.rb
```diff
@@ -20,7 +20,7 @@ def search_by_taxonomy(key, operator, value)
             foreman_tasks_links_taxonomy#{uniq_suffix}.resource_type = '#{resource_type}')
         SQL
         # Select only those tasks which either have the correct taxonomy or are not related to any
-        sql = "foreman_tasks_links_taxonomy#{uniq_suffix}.resource_id #{operator} ? OR foreman_tasks_links_taxonomy#{uniq_suffix}.resource_id IS NULL"
```
ezr-ondrej, line 23: wait, can't the operator be something else then `IN` would this work with `=` ? Could we add test for that as well?

adamruzicka: It can, but then the parentheses just group the expression within them, even if it is just a single value

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #608, opened 2021-02-15
Fixes #31843 - Make sure logging context doesn't get wiped too early

adamruzicka:
The logging context storing middleware was being run after first round of execution plan hooks, which already wiped the logging context.


jjeffers:
@jlsherrill @adamruzicka Test this patch on katello 4.0/master development environment and seems to work.

jlsherrill (approved):

adamruzicka:
I hear no objections, merging

adamruzicka:
Thank you @jjeffers & @jlsherrill !

PR #609, opened 2021-02-18
fixes #31913 - Introducing find_job_invocation_by_id macro

domitea:
This PR is consist of adding `find_job_invocation_by_id` macro for get job invocation instance by id. This macro are requested by another PR in REX (https://github.com/theforeman/foreman_remote_execution/pull/453).

### DIFF
--- a/app/services/renderer_methods.rb
+++ b/app/services/renderer_methods.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# macros to fetch information about invoked jobs
+module RendererMethods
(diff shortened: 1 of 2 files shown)


adamruzicka:
Why add a macro in tasks, which only touches objects from REX to use it from REX?

domitea:
That's a good point. I thorught that `JobInvocation` model is in Tasks, so it would makes sense to move this macro to Tasks. However there is no model at all... :( I'll close it...

PR #610, opened 2021-03-01
Fixes #30043 - Add duration column to exported tasks

pheeeter:


lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -216,11 +216,24 @@ namespace :foreman_tasks do
...
+      def self.count_duration(task)
+        duration = nil
+        if task.started_at && task.ended_at
+          duration = (task.ended_at - task.started_at).floor
+        elsif task.started_at && task.ended_at.nil?
+          duration = (Time.now - task.started_at).floor
+        else
+          duration = "N/A"
+        end
+        return duration
```
adamruzicka, line 228: In ruby, everything is an expression. A function implicitly returns the result of evaluating last expression in the function, explicit returns are only needed if you need to exit early. The same goes for `if`, result of evaluating an `if` statement is whatever the selected branch evaluated to. With this in mind, you should be able to reduce the code a bit

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -217,15 +217,13 @@ namespace :foreman_tasks do
...
-        duration = nil
-        if task.started_at && task.ended_at
-          duration = (task.ended_at - task.started_at).floor
-        elsif task.started_at && task.ended_at.nil?
-          duration = (Time.now - task.started_at).floor
-        else
-          duration = "N/A"
-        end
-        return duration
+        duration = if task.started_at && task.ended_at
```
adamruzicka, line 220: Almost there. Why is this local variable needed?

pheeeter: Ah, my bad

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -216,11 +216,22 @@ namespace :foreman_tasks do
         end
       end
 
+      def self.count_duration(task)
+        if task.started_at && task.ended_at
+          (task.ended_at - task.started_at).floor
+        elsif task.started_at && task.ended_at.nil?
+          (Time.now.to_i - task.started_at).floor
```
adamruzicka, line 223: If both `started_at` and `ended_at` are non-nil, you just subtract them, but if the duration is open ended, you convert the time to integer first. Why?

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -216,11 +216,22 @@ namespace :foreman_tasks do
         end
       end
 
+      def self.count_duration(task)
```
adamruzicka, line 219: This is fine for the time being, but maybe having it as a method in the `Task` class would be better?

lib/foreman_tasks/tasks/export_tasks.rake
pheeeter: It should be `to_f` not `to_i`

adamruzicka: But the question remains. Why do you subtract two times in one branch, but time from integer/float in another?

pheeeter: I added there `to_i` then `to_f`, because of rubocop testing... so should be there `to_f` for each time?

adamruzicka: What exactly was rubocop complaining about?

pheeeter:
```
lib/foreman_tasks/tasks/export_tasks.rake:223:17: C: Rails/TimeZone: Do not use Time.now without zone. Use one of Time.zone.now, Time.current, Time.now.in_time_zone, Time.now.utc, Time.now.getlocal, Time.now.xmlschema, Time.now.iso8601, Time.now.jisx0301, Time.now.rfc3339, Time.now.httpdate, Time.now.to_i, Time.now.to_f instead.
```

app/models/foreman_tasks/task.rb
```diff
@@ -255,6 +256,10 @@ def action_output
       main_action.continuous_output.raw_outputs
     end
 
+    def duration
+      {}
+    end
```
adamruzicka, line 261: Why keep this around?

app/models/foreman_tasks/task.rb
```diff
@@ -76,6 +76,7 @@ def check_permissions_after_save
                                  :"foreman_tasks_links.resource_type" => resource.class.name)
            end)
     scope :for_action_types, (->(action_types) { where('foreman_tasks_tasks.label IN (?)', Array(action_types)) })
+    scope :duration, -> { select("DISTINCT foreman_tasks_tasks.*, coalesce(ended_at, current_timestamp) - coalesce(coalesce(started_at, ended_at), current_timestamp) as duration") }
```
adamruzicka, line 79: Since it returns a scope of tasks with calculated duration, maybe `with_duration` would be a better name?

adamruzicka (commented):
You still need to use the new scope in the non-api controller and in export rake task. A place just before the collection of tasks gets processed one-by-one would probably be a good place

app/models/foreman_tasks/task.rb
pheeeter: Sorry, I forgot to remove that

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -99,7 +99,8 @@ def resource_class
 
     def respond_with_tasks(scope)
       @tasks = filter(scope, paginate: false)
-      csv_response(@tasks, [:id, :action, :state, :result, 'started_at.in_time_zone', 'ended_at.in_time_zone', :username], ['Id', 'Action', 'State', 'Result', 'Started At', 'Ended At', 'User'])
+      @tasks = @tasks.merge(ForemanTasks::Task.with_duration)
```
adamruzicka, line 102: You should be able to simplify this with `@tasks = filter(scope, paginate: false).with_duration`

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -242,15 +242,16 @@ namespace :foreman_tasks do
     export_filename = ENV['TASK_FILE'] || "/tmp/task-export-#{Time.now.to_i}.#{format == 'csv' ? 'csv' : 'tar.gz'}"
 
     tasks = ForemanTasks::Task.search_for(filter)
+    tasks = tasks.merge(ForemanTasks::Task.with_duration.distinct)
```
adamruzicka, line 245: Here again `tasks = ForemanTasks::Task.search_for(filter).with_duration.distinct` should work

adamruzicka (changes requested):
Looks promising, but there are still some things to fix.

Show endpoint of the api controller does not return duration, which in turn makes `hammer task info` not show duration.

Also would be nice to get some test coverage

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -322,7 +322,7 @@ def resource_class
       end
 
       def find_task
-        @task = resource_scope.find(params[:id])
+        @task = resource_scope_for_index.find(params[:id])
```
adamruzicka, line 325: `resource_scope_for_index` behaves slightly differently than just `resource_scope`, wouldn't something like `resource_scope.with_index.find(...)` work?

test/controllers/api/tasks_controller_test.rb
```diff
@@ -76,6 +76,24 @@ class Api::TasksControllerTest < ActionController::TestCase
                      session: set_session_user(User.current)
           assert_response :not_found
         end
+
+        it 'shows duration column in exported task' do
```
adamruzicka, line 80: This applies to all the tests, hitting `show`, `index` or `sub_tasks` endpoints with a request does not do an export so mentioning "exported task" in the test name might be misleading

test/controllers/api/tasks_controller_test.rb
```diff
@@ -76,6 +76,24 @@ class Api::TasksControllerTest < ActionController::TestCase
...
+          assert_response :success
+          data = JSON.parse(response.body)
+          _(data['duration']).must_equal task.duration
+        end
+      end
+
+      describe 'GET /api/tasks/index' do
+        it 'shows duration column in exported task' do
+          task = ForemanTasks::Task.with_duration.find(FactoryBot.create(:dynflow_task).id)
+          get :index, params: { id: task.id }, session: set_session_user
```
adamruzicka, line 92: Index does not accept `id` parameter

test/controllers/tasks_controller_test.rb
```diff
@@ -89,6 +89,20 @@ def in_taxonomy_scope(organization, location = nil)
                      session: set_session_user(User.current)
           assert_response :not_found
         end
+
+        it 'shows duration column in exported task' do
+          task = ForemanTasks::Task.with_duration.find(FactoryBot.create(:some_task).id)
+          get(:show, params: { id: task.id, duration: task.duration }, session: set_session_user)
```
adamruzicka, line 95: Show ignores `duration` parameter

test/controllers/tasks_controller_test.rb
```diff
@@ -89,6 +89,20 @@ def in_taxonomy_scope(organization, location = nil)
...
+          assert_response :success
```
adamruzicka, line 96: This test currently tests that a request succeeds, but not that correct data is returned.

test/controllers/tasks_controller_test.rb
```diff
@@ -89,6 +89,20 @@ def in_taxonomy_scope(organization, location = nil)
...
+        end
+      end
+
+      describe 'index' do
+        it 'shows duration column in exported task' do
+          task = ForemanTasks::Task.with_duration.find(FactoryBot.create(:some_task).id)
+          get(:index, params: { id: task.id, duration: task.duration }, session: set_session_user)
```
adamruzicka, line 103: Index ignores both `id` and `duration` parameters

test/controllers/tasks_controller_test.rb
```diff
@@ -89,6 +89,20 @@ def in_taxonomy_scope(organization, location = nil)
...
+          assert_response :success
```
adamruzicka, line 104: This test currently tests that a request succeeds, but not that correct data is returned.

test/controllers/tasks_controller_test.rb
```diff
@@ -109,6 +123,12 @@ def in_taxonomy_scope(organization, location = nil)
           assert_equal 2, response.body.lines.size
           assert_include response.body.lines[1], 'Child action'
         end
+
+        it 'shows duration column in exported task' do
+          task = ForemanTasks::Task.with_duration.find(FactoryBot.create(:some_task).id)
+          get(:sub_tasks, params: { id: task.id, duration: task.duration }, session: set_session_user)
```
adamruzicka, line 129: Sub tasks lists tasks which are children of a task specified by `id` parameter. `duration` parameter is ignored.

test/controllers/tasks_controller_test.rb
```diff
@@ -109,6 +123,12 @@ def in_taxonomy_scope(organization, location = nil)
...
+          assert_response :success
```
adamruzicka, line 130: This test currently tests that a request succeeds, but not that correct data is returned. Additionally, you request sub tasks of a specific task, but that task has no sub tasks.

adamruzicka (changes requested):

test/controllers/tasks_controller_test.rb
```diff
@@ -124,10 +119,14 @@ def in_taxonomy_scope(organization, location = nil)
           assert_include response.body.lines[1], 'Child action'
         end
 
-        it 'shows duration column in exported task' do
-          task = ForemanTasks::Task.with_duration.find(FactoryBot.create(:some_task).id)
-          get(:sub_tasks, params: { id: task.id, duration: task.duration }, session: set_session_user)
+        it 'shows duration column' do
+          parent = ForemanTasks::Task.with_duration.find(FactoryBot.create(:some_task).id)
```
adamruzicka, line 123: The only thing we need to know about the parent task is the id, so we should be fine with just `FactoryBot.create(:some_task).id`. There's no need to reload it with duration

adamruzicka:
Could you please squash the commits so we have one commit for controller changes, one for the rake task and one for tests?

adamruzicka:
Needs a rebase

adamruzicka:
That didn't really help. Rebasing your commit on top of current master would be better approach than merging current master into your branch

pkoprda:
Thank you @adamruzicka for your review! Now I know that I should use rebasing instead of merging when something like this happens.

PR #611, opened 2021-03-01
Fixes #31953 - ObservableAction jail doc

adamruzicka:


adamruzicka:
CC @ofedoren

ofedoren (approved):
Thanks, @adamruzicka! LGTM for starters. We can add some improvements afterwards (probably will require a new version of `apipie_dsl`)

adamruzicka:
> We can add some improvements afterwards (probably will require a new version of apipie_dsl)

Could you please file in issue for it so it doesn't get lost?

adamruzicka:
Thank you @ofedoren !

ofedoren:
> > We can add some improvements afterwards (probably will require a new version of apipie_dsl)
> 
> Could you please file in issue for it so it doesn't get lost?

Actually I don't see what can be improved just yet. I meant that IF something occurres, then it can be improved later :)

PR #612, opened 2021-03-02
Refs #31773 - Fix pg syntax errors when searching by taxonomy

adamruzicka:
Previous fix allowed searching using IN, but only if there was a single value.
The value was treated as a string of comma-delimited numbers instead of an array
of numbers.

### DIFF
--- a/test/unit/task_test.rb
+++ b/test/unit/task_test.rb
@@ -134,6 +134,8 @@ class TasksTest < ActiveSupport::TestCase
       test 'can search by taxonomies using IN' do
         assert_nothing_raised(PG::SyntaxError) { ForemanTasks::Task.search_for('location_id ^ (1)').first }
         assert_nothing_raised(PG::SyntaxError) { ForemanTasks::Task.search_for('organization_id ^ (1)').first }
+        assert_nothing_raised(PG::SyntaxError) { ForemanTasks::Task.search_for('location_id ^ (1,2)').first }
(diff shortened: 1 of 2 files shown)


test/unit/task_test.rb
```diff
@@ -132,8 +132,8 @@ class TasksTest < ActiveSupport::TestCase
 
     context 'by taxonomies' do
       test 'can search by taxonomies using IN' do
-        assert_nothing_raised(PG::SyntaxError) { ForemanTasks::Task.search_for('location_id ^ (1)').first }
-        assert_nothing_raised(PG::SyntaxError) { ForemanTasks::Task.search_for('organization_id ^ (1)').first }
+        assert_nothing_raised(PG::SyntaxError) { ForemanTasks::Task.search_for('location_id ^ (1,2)').first }
```
ezr-ondrej, line 135: can we keep both tests? for a single value and for multiple values?

ezr-ondrej (commented):
We are implementing more and more of scoped_search funcionality in our code
Let's roll with this, but I'd imagine we can do something easier here :thinking:

adamruzicka:
> I'd imagine we can do something easier here 

Iirc scoped search would get confused when dealing with habtm associations with sti. I think I tried doing something about it in scoped search but I don't recall how it went. Definitely worth revisiting later

pondrejk (approved):
checked on patched downstream, ack from the functional point of view

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #613, opened 2021-03-04
Fixes #30248 - upgrade vendor to v8

MariaAga:
opened to replace: https://github.com/theforeman/foreman-tasks/pull/567
Tested the plugin and it works fine for me

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "url": "http://projects.theforeman.org/projects/foreman-tasks/issues"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


ezr-ondrej (commented):
This is foreman 2.4 incompatible as vendor 8.3.3 is only in develop packaging, should we bump the foreman version dependency?

MariaAga:
I now see that https://github.com/theforeman/foreman-tasks/pull/585 fixed #30248 so I'm not sure if we need to update js vendor. I'll close it for now.

PR #614, opened 2021-03-05
Fixes #32024 - Drop fallback to non-batch triggering API

adamruzicka:
The batch triggering API was introduced in 1.20, it should be available on all
releases which are not EOL.

### DIFF
--- a/app/lib/actions/proxy_action.rb
+++ b/app/lib/actions/proxy_action.rb
@@ -29,10 +29,10 @@ module Actions
       default_connection_options.each do |key, value|
         options[:connection_options][key] = value unless options[:connection_options].key?(key)
       end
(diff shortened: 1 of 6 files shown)


adamruzicka:
We need to hold this for a bit, do not merge please.

adamruzicka:
Ok, we don't need to hold this anymore

adamruzicka:
Requires https://github.com/theforeman/smart_proxy_dynflow/pull/93 on the proxy side

ezr-ondrej (commented):
We need to bump smart_proxy_dynflow dependency.

adamruzicka:
We were depending on smart_proxy_dynflow from tasks-core and I'd rather drop that completely.

ezr-ondrej (approved):
Oh I've missed this one as well, this is deffinitely what we want :+1:

PR #616, opened 2021-03-09
Fixes #32043 - allow action tweak proxy_batch_size

ezr-ondrej:
Ask action if it has any special value for `proxy_batch_size` and only
then set the default.


adamruzicka:
Will this still be needed after https://github.com/theforeman/foreman-tasks/pull/640 gets in?

adamruzicka:
@ezr-ondrej ping

PR #617, opened 2021-03-09
Fixes #32047 - Match proxy actions using task_id

adamruzicka:
When we're retrieving outputs of proxy actions we get back a serialized
execution plan, walk its actions and look for an action with either expected
action_class or operation name. This is somewhat flaky.

With this change we only check one field in the actions' input and that is
task_id. First action found with matching value is considered the source of
output for the current task.

### DIFF
--- a/app/lib/actions/proxy_action.rb
+++ b/app/lib/actions/proxy_action.rb
@@ -218,10 +218,8 @@ module Actions
     end
 
     def get_proxy_data(response)
-      proxy_data = response['actions'].detect do |action|
-        action['class'] == proxy_action_name || action.fetch('input', {})['proxy_operation_name'] == proxy_operation_name
-      end
-      proxy_data.fetch('output', {})
+      response['actions'].detect { |action| action.fetch('input', {})['task_id'] == task.id }
+                         .try(:fetch, 'output', {})
     end
 
     def proxy_version(proxy)


app/lib/actions/proxy_action.rb
```diff
@@ -218,10 +218,8 @@ def start_or_resume
     end
 
     def get_proxy_data(response)
-      proxy_data = response['actions'].detect do |action|
-        action['class'] == proxy_action_name || action.fetch('input', {})['proxy_operation_name'] == proxy_operation_name
-      end
-      proxy_data.fetch('output', {})
+      response['actions'].detect { |action| action['input']['task_id'] == task.id }
```
ares, line 221: given how `action.fetch('input', {})` was used defensively before, are we sure action['input'] is always a hash? or should we also write that more defensively?

ares (commented):
One comment, otherwise makes sense. I wonder why we didn't do that like this before? Is there any moment or use case, where we wouldn't have the task id available in the parsed response?

app/lib/actions/proxy_action.rb
adamruzicka: If there's an action there should be an input and if there's an input it is a hash. Otoh playing it safe is almost free so I can do it that way

adamruzicka: Fixed

adamruzicka:
FTR, I would consider this temporary. It will most likely change when we start retrieving outputs in batches

ares (approved):
Thanks @adamruzicka, merging!

ares:
This probably caused the regression that for Ansible, I see `Error loading data from proxy: NoMethodError - undefined method `fetch' for nil:NilClass` with every update of the console (1 sec interval) until it finishes. I'll see if reverting helps when I have a moment

adamruzicka:
To be honest I'd be surprised if this was the cause as this doesn't really change how we access the data.

adamruzicka:
https://github.com/theforeman/foreman-tasks/pull/629

PR #618, opened 2021-03-09
Fixes #32050 - Prevent BatchRunnerAction from collecting outputs

adamruzicka:


lib/foreman_tasks_core/runner/parent.rb
```diff
@@ -10,21 +10,23 @@ def initialize(targets = {}, suspended_action: nil)
       end
 
       def generate_updates
-        @outputs.reduce({}) do |acc, (key, value)|
+        base = {}
+        base[@suspended_action] = Runner::Update.new(ForemanTasksCore::ContinuousOutput.new, @exit_status) if @exit_status
```
ezr-ondrej, line 14: I'm sorry, but I can't find where the @suspended_action comes from for sure. Is it from dynflows `Action#suspended_action` ? isn't it an instance of `Action::Suspended`? In that case why are we using it as a key? :thinking:
What am I missing here?

adamruzicka: Each runner holds a reference to the suspended action which spawned the runner, that is `@suspended_action` in this context and it is indeed `Action::Suspended`. It gets passed to the runner on initialization and is stored in the top-level runner class here https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks_core/runner/base.rb#L10 .

The return value from `generate_updates` is a hash, where keys are `Action::Suspended`-like objects, which serve as targets when sending events with updates to the per-host actions, and values are the updates themselves. Usually we want to send updates only to the per-host actions, but in case `@exit_status` is set, we need to send an update to the runner's action as well so it can finish.

ezr-ondrej: OH! now I see it, we are generating `Action::Suspended` from the host names :facepalm:
It makes me bit uneasy to have complicated objects as hash keys, I'd feel better if `Runner::Dispatcher` would know the `host -> execution_plan_id` mapping and we would just send hostnames and nil for global, but we can't always get what we wish for :) :shrug:

Your changes make it bit more readable as we can clearly see what are the keys supposed to look like at the beggining of the method so :+1: and thanks for the explanation

lib/foreman_tasks_core/runner/parent.rb
```diff
@@ -10,21 +10,23 @@ def initialize(targets = {}, suspended_action: nil)
...
+        @outputs.reduce(base) do |acc, (key, value)|
           if value.empty? && @exit_status.nil?
             acc
           else
             identifier = key
             @outputs[identifier] = ForemanTasksCore::ContinuousOutput.new
-            key = host_action(identifier) unless identifier == @suspended_action
+            key = host_action(identifier)
             exit_status = @exit_statuses[identifier] || @exit_status if @exit_status
             acc.merge(key => Runner::Update.new(value, exit_status))
```
ezr-ondrej, line 23: could we rename the variable? it is confusing to redefine the key from `hostname<String>` to `Action::Suspended` I believe that identifier is now useless assignment and we could have:
```suggestion
            identifier = host_action(key)
            exit_status = @exit_statuses[identifier] || @exit_status if @exit_status
            acc.merge(identifier => Runner::Update.new(value, exit_status))
```

Or maybe the other way around `key -> identifier` being the hostname directly comming to the loop block and keeping the key as the calculated key as is

ezr-ondrej (commented):
just a naming nit, that could help future readers to resolve the confusion I had bit more easier, otherwise :+1:

ezr-ondrej (approved):
This definitely improves the code :+1:
Thanks @adamruzicka

PR #619, opened 2021-03-10
Implement retries for batch triggering

adamruzicka:


adamruzicka:
Now that we have async triggering this shouldn't be necessary. I'll keep the branch around in case we want to revisit this in the future, but I'll close this PR.

PR #620, opened 2021-03-15
Mop up after split from locks to locks and links

laviro:
looks like migration and API endpoint that rely on OWNER_LOCK_NAME are failing now,
reverting with a small fix

### DIFF
--- a/app/models/foreman_tasks/lock.rb
+++ b/app/models/foreman_tasks/lock.rb
@@ -1,5 +1,7 @@
 module ForemanTasks
   class Lock < ApplicationRecord
+    OWNER_LOCK_NAME = :task_owner
+
     class LockConflict < StandardError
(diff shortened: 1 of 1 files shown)


app/models/foreman_tasks/lock.rb
```diff
@@ -1,5 +1,7 @@
 module ForemanTasks
   class Lock < ApplicationRecord
+    OWNER_LOCK_NAME = :task_owner
```
adamruzicka, line 3: Originally tasks were linked to users using locks. Later we swapped this for explicitly storing user_id on tasks, that's why the constant was needed in the failing migration. Once that migration is done, there should be no locks using `OWNER_LOCK_NAME` as a resource_type, meaning the use of this constant in the api controller makes no sense.

laviro:
Closing in favor of the work you did @adamruzicka!

PR #622, opened 2021-03-19
Fixes #32132 - Make tasks counts match when using dashboard

adamruzicka:
We had two ways of filtering tasks in the controllers. One relied on the search
query passed as search parameter, the other relied on a couple of special
parameters. If any of the special parameters were provided, we performed the
search *before* we passed the scope to the helper which does the actual search
based searching. Because of this total and subtotal in the response did not
match what was shown.

This change throws away the special parameters handling from the backend and
makes it frontend's responsibility to assemble proper search query.


MariaAga:
I'm currently testing it, but this  seems to fix 30682 as well.
[tasks api has wrong subtotal with dashboard filters](https://projects.theforeman.org/issues/30682)

adamruzicka:
> but this seems to fix 30682 as well.

Is it a bad thing? :)

MariaAga:
In a good way! the pagination shows almost the exact number of filtered tasks now. not sure why not the exact number but I have 240 tasks, 155 success stopped, but pagination shows 163 (subtotal in api is 163)

adamruzicka:
total is before filtering, subtotal after filtering, correct? I'll try looking into it

MariaAga:
Yeah, it's working fine with the other filters but stopped success is not exact

webpack/ForemanTasks/Components/TasksTable/__tests__/__snapshots__/TasksTablePage.test.js.snap
```diff
@@ -227,7 +227,7 @@ exports[`TasksTablePage rendering render with minimal props 1`] = `
         />
         <ExportButton
           title="Export All"
-          url="/foreman_tasks/tasks.csv?action=%22some-name%22&state=stopped"
+          url="/foreman_tasks/tasks.csv?action=%22some-name%22&search=%28state%3Dstopped%29"
```
MariaAga, line 230: This doesn't look like the correct query, shouldn't it be `search=(state=stopped& action=some-name)?

MariaAga: Its happening because getURIQuery is mocked and ` history.location.search` is no longer used in `getCSVurl` so its ok

ezr-ondrej (approved):
Getting rid of DashboardTableFilter is definitelly a good thing, I still have some missmatches in counts though, but I'm ok with any improvement.
I'm ok to merge, agreed @adamruzicka ?

adamruzicka:
Yeah, sure. Could you file an issue about the mismatches you're still seeing?

ezr-ondrej:
> Yeah, sure. Could you file an issue about the mismatches you're still seeing?

Interestingly, playing with it, it always mismatches by 1 for me :thinking:

adamruzicka:
> Interestingly, playing with it, it always mismatches by 1 for me thinking

Huh, interesting, but I didn't manage to reproduce it on my machine. So let's get this in and I'll keep an eye out for it?

ezr-ondrej:
Thanks @adamruzicka ! :)

PR #625, opened 2021-03-29
Fix test rake task used in CI

adamruzicka:


test/factories/task_factory.rb
```diff
@@ -58,7 +81,7 @@
         after(:create) do |task, evaluator|
           create_list(
             :lock,
-            evaluator.locks_count,
+            1,
```
ezr-ondrej, line 84: list of 1? Why not use `create` then?

adamruzicka: The association requires it to be an array and while we could construct it differently, this worked. Also bumping the number doesn't make sense since locks have to be unique by `(task_id, resource_type, resource_id)` triplet, unless be make the count somehow touch some part of it

PR #628, opened 2021-04-26
Extract strings

adamruzicka:


adamruzicka:
@upadhyeammit Hi, to double check, should I take just the `.pot` file or all the changes, including `.po` files?

upadhyeammit:
> @upadhyeammit Hi, to double check, should I take just the `.pot` file or all the changes, including `.po` files?

We need both .pot and .po files. The .pot has source strings which are referred while creating the .po files for each language. I strongly recommend to do one more pull before we do the GA of 2.5. As I will hand over string for translation on branching for Foreman and all plugins, and then before branching I will merge those on transifex. I will ping you back/raise pr to merge it before 2.5 GA. This means whatever plugin version you are aiming for 2.5 should get the translation pull.

ezr-ondrej (approved):
Better late then never I guess, sorry it got stuck in wrong queue. Thanks @adamruzicka :+1:

PR #630, opened 2021-04-28
Fixes #32445 - Stub out settings in proxy action tests

adamruzicka:
Those settings used to be nil in tests, with the SettingsRegistry they are
initialized to a default value which messes up the tests.

### DIFF
--- a/test/unit/actions/proxy_action_test.rb
+++ b/test/unit/actions/proxy_action_test.rb
@@ -14,6 +14,9 @@ module ForemanTasks
         Support::DummyProxyAction.any_instance.stubs(:with_batch_triggering?).returns(batch_triggering)
         Support::DummyProxyAction.reset
         RemoteTask.any_instance.stubs(:proxy).returns(Support::DummyProxyAction.proxy)
+        Setting.stubs(:[]).with('foreman_tasks_proxy_action_retry_interval')
+        Setting.stubs(:[]).with('foreman_tasks_proxy_action_retry_count')
+        Setting.stubs(:[]).with('foreman_tasks_proxy_batch_trigger')
         @action = create_and_plan_action(Support::DummyProxyAction,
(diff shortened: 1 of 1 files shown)


test/unit/actions/proxy_action_test.rb
```diff
@@ -14,6 +14,9 @@ class ProxyActionTest <  ActiveSupport::TestCase
         Support::DummyProxyAction.any_instance.stubs(:with_batch_triggering?).returns(batch_triggering)
         Support::DummyProxyAction.reset
         RemoteTask.any_instance.stubs(:proxy).returns(Support::DummyProxyAction.proxy)
+        SettingRegistry.instance.stubs(:[]).with('foreman_tasks_proxy_action_retry_interval')
+        SettingRegistry.instance.stubs(:[]).with('foreman_tasks_proxy_action_retry_count')
+        SettingRegistry.instance.stubs(:[]).with('foreman_tasks_proxy_batch_trigger')
```
ezr-ondrej, line 19: Can we stub out `Setting` directly? we are reading it through `Setting` class, right?

ezr-ondrej (commented):
I'd rather stub the Setting, also, don't we want to ensure a value being returned?

test/unit/actions/proxy_action_test.rb
adamruzicka: Correct, stubbed `Setting` directly

adamruzicka:
In this case, there is a default value for when the setting is unset (and that's actually what the tests expected).

ezr-ondrej (approved):
I'd rather see us testing the default that comes from the default value in Setting, as that will be the default behaviour for most users, but lets get this in to fix the test regression and followup when there is time for that :+1:

ezr-ondrej:
The tests are very moody, but it's not caused by this PR.

PR #632, opened 2021-05-06
Refs #31457 - add output messages method alias

lzap:
We already have shipped webhooks with action_output and it is unfortunate that its missing, let's keep the alias as undocumented so it still works.


ezr-ondrej:
@adamruzicka didn't like this name, as it may suggest it is the output as the output hash of dynflow.

adamruzicka:
The examples containing the use of the old name were dropped, this is not needed anymore. Closing.

PR #635, opened 2021-06-04
Make tasks core only a compatibility layer to proxy plugin

adamruzicka:
requires:
- [ ] https://github.com/theforeman/smart_proxy_dynflow/pull/81


ezr-ondrej:
What about the `ContinuousOutput` class?

lib/foreman_tasks_core/continuous_output.rb
```diff
@@ -1,50 +0,0 @@
-module ForemanTasksCore
-  class ContinuousOutput
```
ezr-ondrej, line 2: Didn't we want to keep this?
Maybe transfering to a `ForemanTasks` namespace? :thinking:

ezr-ondrej (commented):
should we alias the actions ?
`OutputCollectorAction` and so on just in case we have some dynflow actions in the database still not finished, they would start failing immediatelly, right?

adamruzicka:
By default dynflow on the smart proxy uses an in memory database so we should start with a clean slate after each process restart. It could be an issue if someone had a persistent db though :thinking:

ezr-ondrej:
> It could be an issue if someone had a persistent db though

How often that happens, do we support such cases and how hard it is to recover from such situation?
I can't judge here but if it can give someone hard time, I'd prefer to have the aliases just in case

adamruzicka:
> How often that happens

No idea, but on-disk sqlite is borderline unusable with our access pattern.

> how hard it is to recover from such situation?

Remove the db and start over? Even if there was no issue with action classes, I'm not sure if any of the actions running on the proxy is actually resumable

ezr-ondrej (approved):
Works well, let's get this in :+1:

PR #636, opened 2021-06-08
Fixes #32729 - reload setting prior run

ezr-ondrej:
In Foreman we've switched to in-memory setting cache, that needs to be
reloaded on demand, Rails is doing that on every request, for dynflow
actions we need to do it on every phase.

### DIFF
--- a/app/lib/actions/middleware/load_setting_values.rb
+++ b/app/lib/actions/middleware/load_setting_values.rb
@@ -1,0 +1,36 @@
+module Actions
+  module Middleware
+    class LoadSettingValues < ::Dynflow::Middleware
+      # ::Actions::Middleware::LoadSettingValues
+      #
+      # A middleware to ensure we load current setting values
+
+      def delay(*args)
+        reload_setting_values
+        pass(*args)
+      end
+
+      def plan(*args)
+        reload_setting_values
+        pass(*args)
+      end
+
+      def run(*args)
+        reload_setting_values
+        pass(*args)
+      end
+
+      def finalize(*args)
+        reload_setting_values
+        pass(*args)
+      end
+
+      private
+
+      def reload_setting_values
+        return unless Gem::Version.new(::SETTINGS[:version]) >= Gem::Version.new('2.5')
+        ::Foreman.settings.load_values
+      end
+    end
+  end
+end
--- a/app/lib/actions/middleware/load_setting_values.rb
+++ b/app/lib/actions/middleware/load_setting_values.rb
@@ -28,7 +28,6 @@ module Actions
       private
 
       def reload_setting_values
-        return unless Gem::Version.new(::SETTINGS[:version]) >= Gem::Version.new('2.5')
         ::Foreman.settings.load_values
       end
     end
--- a/lib/foreman_tasks/engine.rb
+++ b/lib/foreman_tasks/engine.rb
@@ -113,6 +113,7 @@ module ForemanTasks
         world.middleware.use Actions::Middleware::KeepCurrentUser, :before => ::Dynflow::Middleware::Common::Transaction
         world.middleware.use Actions::Middleware::KeepCurrentTimezone
         world.middleware.use Actions::Middleware::KeepCurrentRequestID
+        world.middleware.use ::Actions::Middleware::LoadSettingValues if Gem::Version.new(::SETTINGS[:version]) >= Gem::Version.new('2.5')
       end
     end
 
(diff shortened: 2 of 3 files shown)


adamruzicka:
Looks good to me. @ShimShtein could you take this for a spin?

ShimShtein:
Tried it, but it looks like the settings are not reloaded.

adamruzicka:
Are you sure the code changes propagated to all the processes? It seems to work just fine for me

ekohl (commented):
Here it states it's 2.4+ compatible. Would this also work on Foreman 2.4 or should the minimum version be raised?
https://github.com/theforeman/foreman-tasks/blob/86c7e83c5b84d3cbce24791fdd1c2b49d91af590/lib/foreman_tasks/engine.rb#L36

ShimShtein:
@ekohl @ezr-ondrej, I have created another PR, since I couldn't change this one. I have modified the original PR so it would also work for ActiveJobs and modified the minimum requirement, since `Foreman.settings` is not available in older versions.

app/lib/actions/middleware/load_setting_values.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      def finalize(*args)
+        reload_setting_values
+        pass(*args)
+      end
+
+      private
+
+      def reload_setting_values
+        return unless Gem::Dependency.new('', ">= 2.5").match?('', ::SETTINGS[:version])
```
ekohl, line 31: I'd use `Gem::Version` instead.
```suggestion
        return unless Gem::Version.new(::SETTINGS[:version]) >= Gem::Version.new('2.5')
```

ezr-ondrej: Done

lib/foreman_tasks/engine.rb
```diff
@@ -113,6 +113,7 @@ class Engine < ::Rails::Engine
         world.middleware.use Actions::Middleware::KeepCurrentUser, :before => ::Dynflow::Middleware::Common::Transaction
         world.middleware.use Actions::Middleware::KeepCurrentTimezone
         world.middleware.use Actions::Middleware::KeepCurrentRequestID
+        world.middleware.use ::Actions::Middleware::LoadSettingValues
```
ekohl, line 116: Rather than a guard in the middleware, can we only load the middleware if Foreman 2.5+ is used?

app/lib/actions/middleware/load_setting_values.rb
```diff
@@ -0,0 +1,36 @@
+module Actions
+  module Middleware
+    class LoadSettingValues < ::Dynflow::Middleware
+      # ::Actions::Middleware::LoadSettingValues
+      #
+      # A middleware to ensure we load current setting values
+
+      def delay(*args)
```
ekohl, line 8: What about keyword argument (`**args`)?

lib/foreman_tasks/engine.rb
adamruzicka, line 116: Could we register the middleware only on 2.5 and newer instead of checking the version all the time?

app/lib/actions/middleware/load_setting_values.rb
adamruzicka: Dynflow doesn't really expect keyword arguments in there https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/middleware.rb#L28

lib/foreman_tasks/engine.rb
ezr-ondrej: Well, I feel it's bit cleaner checking it in the middleware, as

1. it will look ugly in code
2. it is less prone to version not being initialized yet in this phase

OTH doing it in initializer would be easier to spot and remove once we bump foreman requirement version

app/lib/actions/middleware/load_setting_values.rb
ezr-ondrej: there are non on the parent, are those expected to be added?

lib/foreman_tasks/engine.rb
ezr-ondrej: Done

adamruzicka:
Test failures on older rubies seem to be related to kwarg handling changes

ezr-ondrej:
I believe kwarks are not needed, so I droped them. We should not temper with the action arguments IMHO and we do not have kwarks.

adamruzicka:
@ezr-ondrej Forgot to push?

ezr-ondrej:
Now done... I was moving it from commit to commit and didn't do it properly :facepalm:

adamruzicka (approved):

adamruzicka:
Thank you everyone!

PR #638, opened 2021-06-10
Fixes #32771 - adds taxonomy scope to API results

ezr-ondrej:
Current taxonomies are not reflected within the API results, as the
search needs to be explicitly added for them, because of specifics of
Tasks taxonomy scoping - as it doesn't have direct association to
taxonomies.

### DIFF
--- a/app/controllers/foreman_tasks/api/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/api/tasks_controller.rb
@@ -1,6 +1,7 @@
 module ForemanTasks
   module Api
     class TasksController < ::Api::V2::BaseController
+      include ForemanTasks::FindTasksCommon
       include ::Foreman::Controller::SmartProxyAuth
       add_smart_proxy_filters :callback, :features => 'Dynflow'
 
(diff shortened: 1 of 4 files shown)


app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -245,6 +246,13 @@ def search_tasks(search_params)
         scope.all.map { |task| task_hash(task) }
       end
 
+      def current_taxonomy_search
+        conditions = []
+        conditions << "organization_id = #{Organization.current.id}" if Organization.current
+        conditions << "location_id = #{Location.current.id}" if Location.current
+        conditions.empty? ? '' : "(#{conditions.join(' AND ')})"
+      end
```
adamruzicka, line 254: We already have the exact same method in the non-api controller. Could we maybe unify the two?

ezr-ondrej: yup, I've extracted the method from there.
I've tried to unify, but keep the changes minimal.

adamruzicka (approved):
Works nicely, ACK

adamruzicka:
Thank you @ezr-ondrej !

PR #639, opened 2021-06-18
Fixes #32838 - Use AJ wrapper in case loading AJ fails

adamruzicka:


app/models/foreman_tasks/task/dynflow_task.rb
```diff
@@ -140,12 +140,18 @@ def cli_example
...
+      @main_action = execution_plan && execution_plan.root_plan_step.try(:action, execution_plan)
       if active_job?
         job_data = active_job_data
-        @main_action = active_job_action(job_data['job_class'], job_data['arguments'])
-      else
-        @main_action = execution_plan && execution_plan.root_plan_step.try(:action, execution_plan)
+        begin
+          @main_action = active_job_action(job_data['job_class'], job_data['arguments'])
+        # rubocop:disable Lint/SuppressedException
+        rescue
```
ezr-ondrej, line 150: Should we log the failure?

ezr-ondrej (approved):
Thanks @adamruzicka :+1:

PR #640, opened 2021-06-23
Fixes #32853 - send remote tasks in Action

ezr-ondrej:
Introduce an action that will take care of triggering the remote tasks
instead of middleware.

PR #641, opened 2021-07-08
Fixes #32960 - Async proxy batch triggering

adamruzicka:
When triggering tasks on the smart proxy before batch triggering, we always got back the uuid of the task on the smart proxy. With batch triggering this stayed the same, but we had to wait until all the tasks in the batch were planned, which could take some time.

Instead of waiting for the uuids, we make the proxy-side tasks be spawned with uuids matching the ones on foreman's side and therefore we don't have to wait for them to be fully planned.

TODO:
- [x] when checking if a task is on the smart proxy we should also check the parent task to be able to distinguish when the task disappeared and when it was not planned yet


adamruzicka:
Proxy part https://github.com/theforeman/smart_proxy_dynflow/pull/90

adamruzicka:
Currently waiting for https://github.com/theforeman/foreman-tasks/pull/640

adamruzicka:
rebased

ezr-ondrej (approved):
Thanks @adamruzicka ! :shipit:

PR #643, opened 2021-07-21
Fixes #33091 - Trigger recurring logic with a delay

xprazak2:
I plan to pass [start_at](https://github.com/theforeman/foreman_remote_execution/pull/619/files#diff-c7a804360df9b6b127e28adfdaa63daa5fcd5af31d82886dab5e4ff4d3708baaR150) on rex side.

### DIFF
--- a/app/models/foreman_tasks/triggering.rb
+++ b/app/models/foreman_tasks/triggering.rb
@@ -30,7 +30,7 @@ module ForemanTasks
     validates :input_type, :if => :recurring?,
                            :inclusion => { :in => ALLOWED_INPUT_TYPES,
                                            :message => _('%{value} is not allowed input type') }
-    validates :start_at_raw, format: { :with => TIME_REGEXP, :if => :future?,
+    validates :start_at_raw, format: { :with => TIME_REGEXP, :if => ->(triggering) { triggering.future? || (triggering.recurring? && triggering.start_at_raw) },
                                        :message => _('%{value} is wrong format') }
     validates :start_before_raw, format: { :with => TIME_REGEXP, :if => :future?,
                                            :message => _('%{value} is wrong format'), :allow_blank => true }
@@ -70,7 +70,7 @@ module ForemanTasks
                              delay_options,
                              *args
       when :recurring
-        recurring_logic.start(action, *args)
+        recurring_logic.start_after(action, delay_options[:start_at] || Time.zone.now, *args)
       end
     end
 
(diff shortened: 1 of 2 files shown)


adamruzicka:
Looks fine on first glance, will wait for the rex part

xprazak2:
The rex part: https://github.com/theforeman/foreman_remote_execution/pull/627

app/models/foreman_tasks/triggering.rb
```diff
@@ -104,6 +104,7 @@ def parse_start_before!
     private
 
     def can_start_recurring
+      parse_start_at!
```
adamruzicka, line 107: Could we add a validation message instead of just raising?

adamruzicka (commented):
One small nitpick.

@jlsherrill Could you check if this doesn't break some assumption katello makes when using sync plans? I don't expect really it, but better to be safe

app/models/foreman_tasks/triggering.rb
xprazak2: I do not think [parse_start_at!](https://github.com/theforeman/foreman-tasks/blob/master/app/models/foreman_tasks/triggering.rb#L96) raises:

```ruby
Time.zone.parse('foo') #=> nil
```

adamruzicka: It does ocasionally

```
> Time.zone.parse('2021-08-32')
ArgumentError: argument out of range
from /home/aruzicka/.rvm/gems/ruby-2.7.0/gems/activesupport-6.0.3.7/lib/active_support/values/time_zone.rb:548:in `initialize'

> Time.zone.parse('127.0.0.1')
ArgumentError: mon out of range
from /home/aruzicka/.rvm/gems/ruby-2.7.0/gems/activesupport-6.0.3.7/lib/active_support/values/time_zone.rb:548:in `initialize'
```

xprazak2: Updated.

adamruzicka:
Just to double check if I'm reading this right. If I schedule a job to start next monday at noon and run daily, then the first run will happen on monday at noon, next run at midnight between monday and tuesday and then continue happening at midnights?

xprazak2:
The schedule is always honored. If I schedule a job to run daily at midnight starting next monday at noon, then the first run will be at midnight between monday and tuesday.

run-schedule

run-detail

xprazak2:
Anything I can do to move this forward?

adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @xprazak2 !

PR #644, opened 2021-07-22
Fixes #33111 - Add purpose to recurring logic

xprazak2:


adamruzicka:
Could you also expose it here https://github.com/theforeman/foreman-tasks/blob/master/app/views/foreman_tasks/task_groups/recurring_logic_task_groups/_recurring_logic_task_group.html.erb ? Or should it be something that is intentionally hidden from the users?

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -189,5 +193,9 @@ def self.cronline_hash(recurring_type, time_hash, days, days_of_week_hash)
                      end
       hash.select { |key, _| allowed_keys.include? key }
     end
+
+    def valid_purpose?
+      !(purpose.present? && self.class.where(:purpose => purpose, :state => %w[active disabled]).any?)
```
adamruzicka, line 198: I'm a bit worried about concurrent writes here, especially if we will do something like "create a new RL with given purpose if it does not already exist" from an initializer.

xprazak2: Yes, concurrent cases will probably be tricky. Do we know about the places where they might occur?
We could go for ActiveRecord transactions or I can try to conjure a partial unique index for postgres, though I am not sure if that would help. Any ideas?

xprazak2:
Is there anything I can do to move this forward?

xprazak2:
@adamruzicka, what can I do to move this forward?

app/models/foreman_tasks/recurring_logic.rb
adamruzicka: Initializers will most likely be prime suspects. How about instead of the purpose being a property of a recurring logic itself, it would be a separate table where we could leverage unique index? The only downside is RLs would have to drop records from this table when they finish

adamruzicka (commented):
I'm still concerned about concurrent writes. While it might not be an immediate issue, it feels like a thing that will come back and bite us eventually

app/models/foreman_tasks/recurring_logic.rb
xprazak2: That might work, though we also need to retain information about finished RLs with a particular purpose - so we'd probably need another relationship for that.

I looked into the partial index which seems like a less involved approach and pushed a change, let me know what you think.

adamruzicka:
Needs a rebase

app/models/foreman_tasks/triggering.rb
```diff
@@ -105,6 +105,7 @@ def parse_start_before!
 
     def can_start_recurring
       errors.add(:input_type, _('No task could be started')) unless recurring_logic.valid?
+      errors.add(:purpose, _('already exists for active or disabled state')) unless recurring_logic.valid_purpose?
```
adamruzicka, line 108: Currently I wouldn't be able to make sense out of it. Could we reword it a bit? Something along the lines of "Active or disabled recurring logic with purpose '$purpose' already exists".

image

app/models/foreman_tasks/recurring_logic.rb
```diff
@@ -17,6 +17,9 @@ class RecurringLogic < ApplicationRecord
     scoped_search :on => :iteration, :complete_value => false
     scoped_search :on => :cron_line, :complete_value => true
     scoped_search :on => :state, :complete_value => true
+    scoped_search :on => :purpose, :complete_value => false
```
adamruzicka, line 20: Why `:complete_value => false`?

app/models/foreman_tasks/recurring_logic.rb
adamruzicka: Seems to work well :+1:

adamruzicka (commented):
Works well, although error wording could be improved a bit

xprazak2:
Updated.

app/views/foreman_tasks/recurring_logics/index.html.erb
```diff
@@ -62,6 +63,7 @@
       <td><%= format_recurring_logic_limit recurring_logic.end_time.try(:in_time_zone) %></td>
       <td><%= recurring_logic_state(recurring_logic) %></td>
       <td><%= recurring_logic_action_buttons(recurring_logic) %></td>
+      <td><%= recurring_logic.purpose %></td>
```
adamruzicka, line 66: This needs to be one line higher
image

adamruzicka (changes requested):
Noticed one more thing, apart from it it looks good

xprazak2:
Good catch, I swapped the columns.

PR #645, opened 2021-07-28
Allow export_tasks to skip tasks that fail to export

adamruzicka:
Spiritual successor of https://github.com/theforeman/foreman-tasks/pull/588

### DIFF
--- a/lib/foreman_tasks/tasks/export_tasks.rake
+++ b/lib/foreman_tasks/tasks/export_tasks.rake
@@ -14,6 +14,7 @@ namespace :foreman_tasks do
       * TASK_FILE       : file to export to
       * TASK_FORMAT     : format to use for the export (either html, html-dir or csv)
       * TASK_DAYS       : number of days to go back
+      * SKIP_FAILED     : skip tasks that fail to export (true or false[default])
 
     If TASK_SEARCH is not defined, it defaults to all tasks in the past 7 days and
     all unsuccessful tasks in the past 60 days. The default TASK_FORMAT is html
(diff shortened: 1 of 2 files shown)


adamruzicka:
@phess could you check if this works for you?

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -282,6 +288,20 @@ namespace :foreman_tasks do
       end
     end
 
+    def with_error_handling(task, what = 'task')
+      yield
+    rescue StandardError => e
+      resolution = SKIP_ERRORS ? ', skipping' : ''
+      puts _("WARNING: #{what} failed to export#{resolution}. Additional task details below.")
```
ezr-ondrej, line 295: This translation won't work.
Exported string do not match the string that will actually be translated.
Should we interpolate the string instead?

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -282,6 +288,20 @@ namespace :foreman_tasks do
       end
     end
 
+    def with_error_handling(task, what = 'task')
```
ezr-ondrej, line 291: `task` will not get translated

adamruzicka:
Fixed the translations

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -282,6 +288,20 @@ namespace :foreman_tasks do
       end
     end
 
+    def with_error_handling(task, what = _('task'))
```
ezr-ondrej, line 291: just FYI I THINK this will always be english I think, because i believe default params are evaluated on class evaluation, not sure and don't see the point of translation of task :)

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #607, opened 2021-02-15
Fixes #31872 - Disable bulk actions when unauthorized

adamruzicka:
This disables action dropdown and select all in tasks table for users without `edit_foreman_tasks` permission.

Requires:
- [ ] https://github.com/theforeman/foreman/pull/8326

PR #646, opened 2021-08-11
Fixes #31872 - Disable bulk actions when unauthorized

adamruzicka:
disables action dropdown and select all in tasks table for users without
`edit_foreman_tasks` permission.

This was already merged once as https://github.com/theforeman/foreman-tasks/pull/607, but then got reverted for some reason. IIRC the patch in core got reverted. However, the needed changes in core are present in both develop and 2.5-stable so this should be good to go once again.


ezr-ondrej (approved):
Nothing changed since I approved last time, just the dependecly is in :thinking: :shrug:

ezr-ondrej:
Just be aware it is only in 2.5 and up :)

PR #651, opened 2021-09-06
Refs #33091 - Do not add second field

xprazak2:


adamruzicka:
This fixes scheduling jobs to the future, but then there's no option in the ui to create a recurring job which will start later.

xprazak2:
Should be fixed now.

adamruzicka (approved):
Does the trick

adamruzicka:
Thank you @xprazak2 !

PR #653, opened 2021-09-29
Completely drop foreman-tasks-core

adamruzicka:


ezr-ondrej (approved):
Yup, we don't need it around :+1:

PR #656, opened 2021-10-08
Fixes #28475 - Show task count of RL details page

adamruzicka:


ezr-ondrej (approved):
The table is getting little bit crowded, but apart of that :+1:

PR #659, opened 2021-11-23
Fixes #32412 - use DSL to define settings

ezr-ondrej:
Uses the new DSL to define settings.


lib/foreman_tasks/engine.rb
```diff
@@ -65,6 +57,60 @@ class Engine < ::Rails::Engine
...
+                    type: :boolean,
+                    description: N_('Allow triggering tasks on the smart proxy in batches'),
+                    default: true,
+                    full_name: N_('Allow proxy batch tasks'))
+            setting('foreman_tasks_proxy_batch_size',
+                    type: :integer,
+                    description: N_('Number of tasks which should be sent to the smart proxy in one request, if foreman_tasks_proxy_batch_trigger is enabled'),
+                    default: 100,
+                    full_name: N_('Proxy tasks batch size'))
+            setting('foreman_tasks_troubleshooting_url',
```
tbrisker, line 97: This feels like it shouldn't have been a setting in the first place, but that's outside of the scope of this pr

tbrisker (approved):

lib/foreman_tasks/engine.rb
ezr-ondrej: agreed - [RM#33997](https://projects.theforeman.org/issues/33997)

PR #660, opened 2021-11-29
Fixes #34031 - Make cards on Locks tab link to things

adamruzicka:
image


MariaAga:
I couldn't find any tasks with locks but the code looks good :+1:

ezr-ondrej (approved):
And it works. Thanks @adamruzicka and @MariaAga ! :)

PR #662, opened 2021-12-02
Fixes #34065 - Replace deprecated toast actions

tbrisker:
The toast actions were moved to the toast component and the legacy
action has been dropped from core.


ezr-ondrej (approved):
Thanks @tbrisker ! :+1:

PR #665, opened 2022-01-06
Fixes #34227 - Properly reference top-level Rails module

adamruzicka:


ofedoren (commented):
Thanks, @adamruzicka, LGTM. Although it should be quite safe to just merge, is it worth to review more places, e.g. https://github.com/theforeman/foreman-tasks/blob/ef88735b5f0026cbb4f421cbf1492bf02b9d0e5c/lib/foreman_tasks/dynflow/configuration.rb#L25 (wouldn't it look for `Rails` under `Dynflow` namespace there as well?)

adamruzicka:
It seems to resolve correctly in all the other places, although I wouldn't mind being explicit (and safe) everywhere

adamruzicka:
Replaced all occurences of `Rails` with `::Rails` just to be safe

ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

lib/foreman_tasks/dynflow/configuration.rb
```diff
@@ -1,4 +1,4 @@
-require File.expand_path('lib/foreman/dynflow/configuration', Rails.root)
+require File.expand_path('lib/foreman/dynflow/configuration', ::Rails.root)
```
ezr-ondrej, line 1: FYI This actually should be `Rails.root.join('lib', 'foreman' ...)`. But it's unrelated to this PR :)

PR #666, opened 2022-01-17
Fixes #34267 - Use operations even when batch triggering is off

adamruzicka:


ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

ezr-ondrej:
The tests are still complaining :/

ezr-ondrej:
:green_apple: Shall we merge? This should not be a breaking change, right?

adamruzicka:
Yeah, don't quote me on this, but this shouldn't break anything :). Let's get it in

PR #667, opened 2022-01-19
Fixes for interface scripts

adamruzicka:
- Fixes #34288 - Do not crash the task export if SKIP_FAILED is unset
- Refs #27790 - Add option validation to cleanup interface script

### DIFF
--- a/extra/foreman-tasks-cleanup.sh
+++ b/extra/foreman-tasks-cleanup.sh
@@ -21,6 +21,21 @@ function build_rake() {
     echo
 }
 
+function incorrect_usage() {
+    echo "$1" >&2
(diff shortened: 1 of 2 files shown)


extra/foreman-tasks-cleanup.sh
```diff
@@ -119,6 +134,8 @@ while true; do
     shift
 done
 
+validate_options!
```
ezr-ondrej, line 137: I do not like the bang in the method name as it has special meaning in shell, tho it should work. I'm ok, just saying :)

ezr-ondrej (approved):
I'm ok here :shrug:

PR #668, opened 2022-01-28
Fixes #34343 - Unlock tasks on {force_,}unlock

adamruzicka:
After f505bb3 it is not enough for a task to be in stopped state, its locks have
to be really destroyed to count the task as unlocked.

### DIFF
--- a/app/controllers/foreman_tasks/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/tasks_controller.rb
@@ -4,6 +4,8 @@ module ForemanTasks
     include Foreman::Controller::CsvResponder
     include ForemanTasks::FindTasksCommon
 
+    before_action :find_dynflow_task, only: [:unlock, :force_unlock, :cancel, :cancel_step, :resume]
+
     def show
       @task = resource_base.find(params[:id])
       render :layout => !request.xhr?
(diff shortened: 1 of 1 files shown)


app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -72,17 +72,15 @@ def resume
...
-        render json: { statusText: 'OK' }
+        force_unlock(task)
       else
         render json: {}, status: :bad_request
       end
     end
 
-    def force_unlock
-      task       = find_dynflow_task
+    def force_unlock(task = find_dynflow_task)
```
ezr-ondrej, line 81: ugh... controller action with a param smells bit wrong to me. Would it be possible to use `@task` in unlock instead?
then the first line here would be `@task ||= find_dynflow_task` which smells bit better to me.

Or possibly just extract the unlocking into a separate method instead of calling one controller action from the other? it would be the cleanest way IMHO

adamruzicka: Extracted to a separate method + replaced all the calls to find_dynflow_task with a before_action

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -122,9 +122,11 @@ def resource_scope(_options = {})
       @resource_scope ||= ForemanTasks::Task.authorized("#{action_permission}_foreman_tasks")
     end
 
+    # rubocop:disable Naming/MemoizedInstanceVariableName
     def find_dynflow_task
-      resource_scope.where(:type => 'ForemanTasks::Task::DynflowTask').find(params[:id])
+      @dynflow_task ||= resource_scope.where(:type => 'ForemanTasks::Task::DynflowTask').find(params[:id])
```
ezr-ondrej, line 127: I believe you do not need the memoization here (you're always calling it just once, right?) and it would prevent rubocop from complaining.

ezr-ondrej (commented):
Thanks @adamruzicka nice cleanup :+1:

PR #669, opened 2022-01-28
Fixes #34344 - Add support for halting

adamruzicka:
Requires https://github.com/Dynflow/dynflow/pull/406

### DIFF
--- a/app/controllers/foreman_tasks/tasks_controller.rb
+++ b/app/controllers/foreman_tasks/tasks_controller.rb
@@ -69,7 +69,7 @@ module ForemanTasks
 
     def unlock
       if @dynflow_task.paused?
-        unlock_task(@dynflow_task)
+        @dynflow_task.halt
         render json: { statusText: 'OK' }
       else
         render json: {}, status: :bad_request
@@ -77,8 +77,13 @@ module ForemanTasks
     end
 
     def force_unlock
-      unlock_task(@dynflow_task)
-      render json: { statusText: 'OK' }
+      if @dynflow_task.pending?
+        @dynflow_task.halt
+        render json: { statusText: 'OK' }
+      else
+        # Cannot halt an already stopped task
+        render json: {}, status: :bad_request
+      end
     end
 
     # we need do this to make the Foreman helpers working properly
@@ -91,12 +96,6 @@ module ForemanTasks
     end
 
     private
-
-    def unlock_task(task)
-      task.state = :stopped
-      task.locks.destroy_all
-      task.save!
-    end
 
     def respond_with_tasks(scope)
       @tasks = filter(scope, paginate: false).with_duration
--- a/app/models/foreman_tasks/task/dynflow_task.rb
+++ b/app/models/foreman_tasks/task/dynflow_task.rb
@@ -251,6 +251,10 @@ module ForemanTasks
       superclass.model_name
     end
 
+    def halt
+      ::ForemanTasks.dynflow.world.halt(external_id)
+    end
+
     private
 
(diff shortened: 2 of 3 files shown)


app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -72,18 +72,14 @@ def resume
...
         render json: {}, status: :bad_request
       end
     end
 
-    def force_unlock
-      task       = find_dynflow_task
-      task.state = :stopped
-      task.save!
+    def force_unlock(task = find_dynflow_task)
+      task.halt
```
ofedoren, line 82: After description in https://github.com/Dynflow/dynflow/pull/406 I understand that this means:
>  for this task do halt the execution plan but please do finish any running steps.

Is this correct? If so, I still don't quite understand how the whole task will be marked as stopped since we remove here the `task.state = :stopped`? Won't it be `pending` even after the running steps are done?

Also, we kinda change here behaviour we had previously. Wouldn't it confuse the users even more?

adamruzicka: > Is this correct?

In general, yes, but if I were to be pedantic, I'd have to say almost. There might be different interpretations of "finish any running steps" with slight nuances, but yes. If an action's `#run` is being executed while the execution plan is halted, the execution plan moves over to stopped, but the action's `#run` will continue to run until it returns. This doesn't necessarily mean the step itself will be finished.

> I still don't quite understand how the whole task will be marked as stopped 

There is a [mechanism](https://github.com/theforeman/foreman-tasks/blob/master/app/lib/actions/base.rb#L5), which makes tasks to be updated based on the state of its corresponding execution plan as the execution plan changes states. Ideally, tasks should be read-only and the only changes should be done to them by this mechanism. Once we flip the execution plan over to stopped, this hook will trigger and update the task to stopped.

> we kinda change here behaviour we had previously. Wouldn't it confuse the users even more?

The previous behaviour was seriously broken. Dynflow doesn't care about tasks' attributes at all, so if we went in and updated the state of a task, it would not really do anything. The task would appear as stopped for some time, but if the underlying execution plan was still running, it would overwrite task's status again. It worked for things like locks which relied on "is a resource locked with a locked by a task which is not stopped", but we reworked that eventually.

ofedoren (approved):
Well, by reading all of your comments and the code, everything LGTM and makes sense now. Not sure if I'm missing any side effects and not sure how to test this all :/ Otherwise ACK from me.

It would be perfect if we somehow ensured we don't brake anything by that. Can our tests do that? If so, let's merge both PRs (after the conflicts are solved of course).

adamruzicka:
Ohh, I knew I was missing something. The new way does nothing when used against a paused execution plan.

Either we can:
- make the code here do different things based on task's state
- introduce halting as a completely new action that can only be used when the task is running
- somehow handle this in dynflow in a do-what-i-mean fashion

ofedoren:
> * introduce halting as a completely new action that can only be used when the task is running

I guess this is the safest way. We could start with this, let users to try this out. But if there will be a request to merge this feature into the current behaviour then we would still need to refactor this again :/

> * somehow handle this in dynflow in a do-what-i-mean fashion

This sounds just OK if there wasn't "somehow" :D Also, I'm kinda afraid to poke Dynflow if we can avoid that.

> * make the code here do different things based on task's state

What benefits it could bring instead of the first approach?

adamruzicka:
Went with the second option, the changes in dynflow should now work reasonably well in all possible scenarios. Now further changes should be needed here

PR #222, opened 2017-01-03
Improved speed of rake export task

lzap:
Find each should improve batch processing a bit, I can see small boost about
1-5 % in my environment. I was also thinking removing the counter which makes
export slower on slow terminals (ssh, encryption, thousands of lines). The
patch also changed the default format from html to csv, I can change that if
you don't like it.

PR #670, opened 2022-02-08
Fixes #34108 - Sort generated tasks by date

pkoprda:


adamruzicka:
This essentially undoes https://github.com/theforeman/foreman-tasks/pull/222 . While that is a valid approach, couldn't we be more smart about it? `#find_each` loads the records in batches, reducing overall memory consumption, but discards ordering. `#each` loads the whole collection (read every single task in existence) and then iterates over it in the right order which is less the ideal when people have millions of tasks.

`#find_each` discards the ordering to avoid issues where the ordering might change between batches, but we know that it cannot really happen when we're sorting by started at as new tasks cannot start in the past.

Could you take a stab at controlling the iteration explicitly?

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -317,7 +323,7 @@ namespace :foreman_tasks do
     format = ENV['TASK_FORMAT'] || 'html'
     export_filename = ENV['TASK_FILE'] || generate_filename(format)
 
-    tasks = ForemanTasks::Task.search_for(filter).order(:started_at => :desc).with_duration.distinct
+    tasks = ForemanTasks::Task.search_for(filter).with_duration.order(:started_at => :desc)
```
adamruzicka, line 326: If I recally correctly, certain search can cause tasks to be duplicated by using multiple left joins. We cannot really just drop `distinct`. Of course we can replace it with a different mechanism (`group by` should work well in this context),  but we cannot just drop it

pkoprda: But how should I use `group by`? Because I found out, that it only works when I group by each column, which is very slowly.

adamruzicka: Something like this could work
```
From 24537a495f44f49281bad45647a9229a783061e1 Mon Sep 17 00:00:00 2001
From: Adam Ruzicka <aruzicka@redhat.com>
Date: Wed, 16 Feb 2022 13:30:01 +0100
Subject: [PATCH] ...

---
 lib/foreman_tasks/tasks/export_tasks.rake | 20 +++++++++++---------
 1 file changed, 11 insertions(+), 9 deletions(-)

diff --git lib/foreman_tasks/tasks/export_tasks.rake lib/foreman_tasks/tasks/export_tasks.rake
index a48483f..76553c4 100644
--- lib/foreman_tasks/tasks/export_tasks.rake
+++ lib/foreman_tasks/tasks/export_tasks.rake
@@ -255,19 +255,20 @@ namespace :foreman_tasks do
       end
     end
 
-    def html_export(workdir, tasks)
+    def html_export(workdir, tasks, task_scope)
       PageHelper.copy_assets(workdir)
+      ids = tasks.pluck(:id)
 
       renderer = TaskRender.new
       count = 0
-      total = tasks.count(:all)
+      total = ids.count
       index = File.open(File.join(workdir, 'index.html'), 'w')
 
       File.open(File.join(workdir, 'index.html'), 'w') do |index|
         PageHelper.pagify(index) do |io|
           PageHelper.generate_with_index(io) do |index|
-            tasks.pluck(:id).in_groups_of(1000, false).each do |ids|
-              tasks.where(id: ids).each do |task|
+            ids.in_groups_of(1000, false).each do |ids|
+              task_scope.where(id: ids).each do |task|
                 content = with_error_handling(task) { renderer.render_task(task) }
                 if content
                   File.open(File.join(workdir, "#{task.id}.html"), 'w') { |file| PageHelper.pagify(file, content) }
@@ -323,21 +324,22 @@ namespace :foreman_tasks do
     format = ENV['TASK_FORMAT'] || 'html'
     export_filename = ENV['TASK_FILE'] || generate_filename(format)
 
-    tasks = ForemanTasks::Task.search_for(filter).with_duration.order(:started_at => :desc)
+    id_scope = ForemanTasks::Task.search_for(filter).order(:started_at => :desc).group(:id, :started_at)
+    task_scope = ForemanTasks::Task.search_for(filter).with_duration
 
     puts _("Exporting all tasks matching filter #{filter}")
-    puts _("Gathering #{tasks.count(:all)} tasks.")
+    puts _("Gathering #{id_scope.count(:all).count} tasks.")
     case format
     when 'html'
       Dir.mktmpdir('task-export') do |tmp_dir|
-        html_export(tmp_dir, tasks)
+        html_export(tmp_dir, id_scope, task_scope)
         system("tar", "czf", export_filename, tmp_dir)
       end
     when 'html-dir'
       FileUtils.mkdir_p(export_filename)
-      html_export(export_filename, tasks)
+      html_export(export_filename, id_scope)
     when 'csv'
-      csv_export(export_filename, tasks)
+      csv_export(export_filename, id_scope)
     else
       raise "Unkonwn export format '#{format}'"
     end
-- 
2.31.1
```

lib/foreman_tasks/tasks/export_tasks.rake
```diff
@@ -241,35 +241,42 @@ namespace :foreman_tasks do
...
 
-    def csv_export(export_filename, tasks)
+    def csv_export(export_filename, id_scope, task_scope)
       CSV.open(export_filename, 'wb') do |csv|
         csv << %w[id state type label result parent_task_id started_at ended_at duration]
-        tasks.find_each do |task|
-          with_error_handling(task) do
-            csv << [task.id, task.state, task.type, task.label, task.result,
-                    task.parent_task_id, task.started_at, task.ended_at, task.duration]
+        id_scope.pluck(:id).in_groups_of(1000, false).each do |ids|
```
adamruzicka, line 247: We don't really need anything that `in_groups_of` offers over `each_slice`, let's use that.

adamruzicka (commented):
One more nitpick, apart from it this is good to go

adamruzicka (approved):
Works well, ACK

adamruzicka:
Thank you @pkoprda !

PR #672, opened 2022-02-14
Fixes #34447 - Fix batch triggering with multiple proxies

adamruzicka:


ezr-ondrej:
I'd love to see some tests for this, as this is quite bad bug, but we can do it in some future refactoring :+1:

PR #673, opened 2022-02-14
Fixes #34447 - Fix batch triggering with multiple proxies

adamruzicka:


ezr-ondrej:
Should this be a proper cherry-pick? :thinking:
We have no way to track this commit connection to #672

adamruzicka:
Fixed the comit message to reference the original fix. 5.3.0 is in 3.2 repos so we'll need to pick it there as well

PR #675, opened 2022-02-28
let ruby/setup-ruby handle caching and bundle install

adamruzicka:


.github/workflows/ruby_tests.yml
```diff
@@ -4,24 +4,22 @@ env:
...
+  #       uses: ruby/setup-ruby@v1
+  #       with:
+  #         ruby-version: 2.7
+  #         bundler-cache: true
+  #     - name: Run rubocop
+  #       run: bundle exec rubocop
   test_ruby:
     runs-on: ubuntu-latest
-    needs: rubocop
+    # needs: rubocop
```
ezr-ondrej, line 22: Why disabling rubocop? :thinking:

adamruzicka: Primarily I was just trying to get to the bottom of an issue I was seeing in a different PR and rubocop failing in dynflow was blocking the issue from manifesting itself. Now that the issue has been resolve, I can put rubocop back and start working on fixing it properly

PR #676, opened 2022-03-03
Fixes #34447 - Fix batch triggering with multiple proxies, vol. 2

adamruzicka:
Leaving as separate commits, please squash before merge.


adamruzicka:
@ezr-ondrej thank you for wanting to have a test for this. While putting together the test I found the previous solution wasn't really fixing it properly

ezr-ondrej (approved):
Right, I did not realize eighter just by looking at the code :)
I'd prefer to have the proxy_url passed into the block, but we can do so once there is a need for it :+1:

PR #677, opened 2022-03-14
Rewrite to a shared GitHub Actions workflow

ekohl:
This is an effort to align all plugin testing on reusable workflows. For now this is mostly testing if I've missed something, so I'm going over the various plugins.

### DIFF
--- a/lib/foreman_tasks/tasks/test.rake
+++ b/lib/foreman_tasks/tasks/test.rake
@@ -9,6 +9,9 @@ namespace :test do
 
     Rake::Task[test_task.name].invoke
   end
+
+  desc 'Alias for test:foreman_tasks'
+  task :'foreman-tasks' => :foreman_tasks
 end
 
 Rake::Task[:test].enhance do
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -1,68 +1,20 @@
-name: Ruby Testing
-on: [pull_request]
-env:
-  RAILS_ENV: test
-  DATABASE_URL: postgresql://postgres:@localhost/test
-  DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
-  BUNDLE_WITHOUT: "journald:development:console:mysql2:sqlite:libvirt"
+---
+name: CI
+
+on: pull_request
+
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.ref_name }}-${{ github.workflow }}
   cancel-in-progress: true
+
 jobs:
   rubocop:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - name: Setup Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: 2.7
-          bundler-cache: true
-      - name: Run rubocop
-        run: bundle exec rubocop
-  test_ruby:
-    runs-on: ubuntu-latest
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Ruby
     needs: rubocop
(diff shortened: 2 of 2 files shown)


adamruzicka:
@ekohl any plans to move this forward?

ekohl:
Plans would be a big word. There certainly is a desire to move it forward, but right now I'm working hard to avoid picking up new work so I can focus on the upcoming conferences and branching.

I have transferred the repo to https://github.com/theforeman/actions a while back and rebased the PR to reflect that. https://github.com/theforeman/actions/issues/1 has the notes on what should be done to move forward. I now see that dropping Ruby 2.5 from the tests is something that should be added to the list.

If anyone wants to pick it up and take it over the finish line I'd be more than happy.

lib/foreman_tasks/tasks/test.rake
```diff
@@ -1,14 +1,15 @@
...
-      t.verbose = true
-      t.warning = false
-    end
-
-    Rake::Task[test_task.name].invoke
+  desc 'Foreman Tasks plugin tests'
+  Rake::TestTask.new(foreman_tasks: 'db:test:prepare') do |t|
+    t.libs << ForemanTasks::Engine.root.join('test')
+    t.pattern = ForemanTasks::Engine.root.join('test/**/*_test.rb')
+    t.test_files = [Rails.root.join('test/unit/foreman/access_permissions_test.rb')]
```
ekohl, line 7: In OpenSCAP I saw a dedicated task that ran the tests from core where it modified things. Especially when some controllers are extended I think that's a good pattern to keep, and it may help with setting up correct load paths. I'm now debating that approach as well.

lib/foreman_tasks/tasks/test.rake
```diff
@@ -1,14 +1,16 @@
...
-    test_task = Rake::TestTask.new('foreman_tasks_test_task') do |t|
-      t.libs << ['test', "#{ForemanTasks::Engine.root}/test"]
-      t.test_files = Dir["#{ForemanTasks::Engine.root}/test/**/*_test.rb"]
-      t.verbose = true
-      t.warning = false
-    end
-
-    Rake::Task[test_task.name].invoke
+  desc 'Foreman Tasks plugin tests'
+  Rake::TestTask.new(foreman_tasks: 'db:test:prepare') do |t|
```
ekohl, line 3: One downside of this is that it'll also be included in `foreman-rake` and executed, which slows things down. The previous approach only defined a very simple task and loaded things on demand.

We may want to exclude this file from the gem or RPMs/debs.

ekohl:
> This differs a little bit from what is suggested in https://github.com/theforeman/actions#foreman-plugin-tests

foreman-tasks is special, because of the plugin name (`foreman-tasks`) differing from the plugin id (`foreman_tasks`).

I've rebased this and made some small tweaks to make the differences to the reference a bit smaller.

lib/foreman_tasks/tasks/test.rake
```diff
@@ -1,14 +1,16 @@
...
+    t.libs << Rails.root.join('test')
+    t.libs << ForemanTasks::Engine.root.join('test')
+    t.pattern = ForemanTasks::Engine.root.join('test', '**', '*_test.rb')
+    t.test_files = [Rails.root.join('test', 'unit', 'foreman', 'access_permissions_test.rb')]
```
adamruzicka, line 7: Should this be dropped so that we're consistent with how the test task is configured in other plugins?

ekohl: If you prefer, I can drop it and then we can solve it consistently with all plugins.

adamruzicka: Please do, unless we'll fix it everywhere this way

adamruzicka (approved):
APT

adamruzicka:
Thank you @ekohl !

PR #678, opened 2022-03-21
test against ehelm's upgrade-rails-6.0.4.7 branch

evgeni:


evgeni:
it didn't the first time, which is confusing…

https://github.com/theforeman/foreman-tasks/runs/5632286413?check_suite_focus=true

ezr-ondrej:
Looks like quite "innocent" transient failure, the fulltext search found task that it was not supposed to find, but that might have quite a lot of unrelated reasons (it matched the uuid, description, or whatever else we use the fullsearch on). Disclaimer: I did not find the attribute it could have matched, so I'm unclear about the exact cause.
Tho I do not get why we test on Ruby 2.6 anyway :stuck_out_tongue:

ehelms:
> Tho I do not get why we test on Ruby 2.6 anyway

https://github.com/theforeman/foreman-tasks/pull/677 would drop that aspect

evgeni:
Closing this PR ;)

PR #679, opened 2022-04-05
Fixes #34806 - Replace pf3 OverlayTrigger

MariaAga:


webpack/ForemanTasks/Components/TasksDashboard/Components/TasksCardsGrid/Components/StoppedTasksCard/OtherInfo.js
```diff
@@ -2,38 +2,35 @@ import React from 'react';
...
+      <Tooltip
+        id="stopped-tooltip"
+        content={__(
+          'Other includes all stopped tasks that are cancelled or pending'
+        )}
+        position={TooltipPosition.bottom}
       >
         <span>
-          <Icon type="pf" name="info" />
+          <InfoCircleIcon className="pficon" />
```
MariaAga, line 28: Using the same classname as pf3 so the css we have for other icons in the page will apply

adamruzicka:
Probably unrelated to this PR per-se, but something odd I noticed. Sometimes (I think there is a click involved), the popups don't go away and you can end up with something like this 

image

Apart from this it seems to work just fine

MariaAga:
These tooltips come from the browser by using `title=...` in the html so that probably unrelated.
Also tried to reproduce it on chrome and firefox and couldn't

ofedoren (commented):
Thanks, @MariaAga! Nice change. The corresponding changes were merged in Foreman, so I guess we can merge this as well. The only thing I'd like to change is the issue ticket to be in ForemanTasks component rather than Foreman and just linked them, but that's just a nitpick...

Also, I've noticed a slight difference in alignment: the number of the "other" tasks is kinda off. Otherwise looks better :)

Before:
before
After:
after

MariaAga:
Thanks @ofedoren!
the number got a bit bigger as pf4 font size is 2px larger than pf3 so it looks different

ofedoren (approved):
I don't see any other reason not to get it in, thanks @MariaAga :)

P.S. PR's title differs from the commit name.

ezr-ondrej (approved):
Thanks @MariaAga and @ofedoren ! :+1:

PR #680, opened 2022-04-06
Fixes #34741 - Supply time in right time zone

adamruzicka:
when calculating next occurence time. CronParser is picky about this.


stejskalleos (approved):
:heavy_check_mark: **LGTM** 
Tested with time zone -10 hours in combinations `hourly`, `daily`, `weekly`, `monthly` and it works as expected

stejskalleos:
Tested also with `Asia/Ust-Nera (+10, +1000)` zone and it works as expected.

adamruzicka:
Thank you for the review and testing @stejskalleos !

PR #681, opened 2022-04-22
Fixes #34800 - Check data consistency on executor startup

adamruzicka:


ares (approved):
Thanks @adamruzicka, makes sense, merging! We can't really predict any side effects, so let's try this out :-)

PR #682, opened 2022-05-12
Fixes #34917 - do not load model constants too early

ezr-ondrej:
Rails do not handle well when namespaced models are loaded prior any unnamespaced one.
**EDIT**: please let me look smart and don't ask why :pray: 
Loading models too soon in initialization process will be imposible with Zeitwerk, so it is good practice anyway.

### DIFF
--- a/lib/foreman_tasks/engine.rb
+++ b/lib/foreman_tasks/engine.rb
@@ -42,17 +42,17 @@ module ForemanTasks
         security_block :foreman_tasks do |_map|
           permission :view_foreman_tasks, { :'foreman_tasks/tasks' => [:auto_complete_search, :sub_tasks, :index, :summary, :summary_sub_tasks, :show],
                                             :'foreman_tasks/react' => [:index],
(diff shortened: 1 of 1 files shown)


jeremylenz (approved):
I tested this and confirm that it un-breaks Rails server

* Foreman checked out https://github.com/theforeman/foreman/pull/9050
* Katello on master

Fixes the error `PG::UndefinedTable: ERROR:  relation "application_records" does not exist` when attempting to run migrations or start the rails server.

ACK :+1:

ezr-ondrej:
Thanks @jeremylenz for the debugging session that has lead to this PR, I'm bit worried there will be more instances of this and I still did not figured out **why** is this happening :(

adamruzicka (approved):
Not a huge fan of hardcoding class names as strings all over the place, but if that's what's needed then so be it.

adamruzicka:
Thank you @ezr-ondrej & @jeremylenz !

PR #684, opened 2022-05-24
Fixes #34954 - Render issue after Rails 6.1.6 upgrade

wbclark:
This fixes

```
 Showing /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/foreman-tasks-6.0.2/app/views/foreman_tasks/layouts/react.html.erb where line #13 raised:

`render file:` should be given the absolute path to a file. 'layouts/base' was given instead
```


wbclark:
I believe these test failures are unrelated

adamruzicka (approved):
Fixes the issue for me. Test failures are unrelated and should be fixed by https://github.com/theforeman/foreman-tasks/pull/685

PR #686, opened 2022-06-13
Fixes #35052 - Update js foreman packages to 10.1

MariaAga:
This is the current version that core uses and it makes js tests green


adamruzicka:
What about the coveralls? Are they supposed to be published?

MariaAga:
Can you re run it? it worked in one of the tests

adamruzicka:
Could you also bump required foreman version in here https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L28 ?

MariaAga:
Changed, thanks

adamruzicka (approved):

PR #690, opened 2022-07-20
Fixes #35250 - Run tests only on ruby 2.7 in CI

adamruzicka:


adamruzicka:
Thank you @ofedoren !

PR #691, opened 2022-08-03
Fixes #35328 - Propagate errors when remote parent task fails

adamruzicka:
Before we notified the local child actions that their remote counterpart
has failed, but we didn't include any details. This made the local
actions ask the smart proxy for status, get back a 404, which the local
actions did not expect.

If we know that the remote parent failed during triggering, it is clear
there will not be any remote tasks to ask about, so we can fail straight
away, without having to ask for status.

### DIFF
--- a/app/lib/actions/proxy_action.rb
+++ b/app/lib/actions/proxy_action.rb
@@ -102,6 +102,8 @@ module Actions
       else
         suspend
       end
+    rescue RestClient::NotFound
+      on_proxy_action_missing
     end
 
     def cancel_proxy_task
--- a/app/lib/actions/proxy_action.rb
+++ b/app/lib/actions/proxy_action.rb
@@ -22,7 +22,15 @@ module Actions
       end
     end
 
-    class ProxyActionStopped; end
+    class ProxyActionStopped < RuntimeError
+      def backtrace
+        []
+      end
+    end
+
+    ProxyActionStoppedEvent = ::Algebrick.type do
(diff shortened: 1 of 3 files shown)


ofedoren (approved):
Thanks, @adamruzicka, LGTM and seems working. Also, tried to reproduce the issue in the ticket, but without any luck :/

I'll approve for now without merging in case you want me to try to reproduce the issue and see if this fixes that. Otherwise, let's get this in :)

adamruzicka:
This approach in general should work, but I found an issue with this particular implementation. When sidekiq is used, the exception cannot be properly serialized, sent through redis as an event and the deserialized and instantiated again. I'll have to revisit this and use simpler types.

ofedoren:
I knew I was missing something :/

adamruzicka:
Well, that wasn't so bad. This should be ready again

ofedoren (approved):
Hmm... LGTM. Weird, but OK. I thought that somewhere we would use all the info provided from the exception, but if that's not the case then I'm OK with the workaround.

adamruzicka:
To elaborate on the workaround. Apparently the exception contains both the response and the original request through the response. If I recall correctly, the body of the response is read on demand and the actual read is synchronized so it is thread safe. The monitor used for this synchronization cannot be serialized, pass the boundaries of a process and then be instantiated again. It is ugly, but so far we only care about the exception's class and its backtrace.

adamruzicka:
Thank you @ofedoren !

PR #692, opened 2022-10-04
Fixes #35592 - Fix delegated proxy task watching

adamruzicka:
Consider tasks which were 'triggered' or 'parent-triggered' as running.

Steps to reproduce:
1) Start a rex job that runs for a while, `sleep 1000` works well
2) Wait for the job to get to the smart proxy
3) Kill smart proxy
4) Wait 10 minutes

Actual results: The job keeps pending, no checks to /dynflow/tasks/status going to the proxy

Expected results: One check to /dynflow/tasks/status on the proxy, the job then fails


adamruzicka (commented):
Seems to work well and I haven't noticed any undesirable side effects

PR #694, opened 2022-11-01
Fixes #35523 - translate all table titles

MariaAga:


adamruzicka (approved):
Right, makes sense

adamruzicka:
Thank you @MariaAga !

PR #695, opened 2022-11-25
i18n - pulling from tx

ekohl:
Also syncs the Makefile from the plugin template, with the difference that it still hardcodes the domain because it doesn't match the gem name (foreman_tasks vs foreman-tasks).


locale/action_names.rb
```diff
@@ -1,5 +1,4 @@
 # Autogenerated!
-_("Import Puppet classes")
-_("Import facts")
 _("Action with sub plans")
-_("Remote action:")
\ No newline at end of file
+_("Import Puppet classes")
+_("Import facts")
```
ekohl, line 4: Stable sorting of this in https://github.com/theforeman/foreman-tasks/pull/698.

PR #696, opened 2022-12-01
code of conduct

mikaila-martin:
created code_of_conduct file

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,13 @@
+Community Guidelines
+We have a variety of communication tools that the Foreman Community uses to get work done. The guidelines below apply in general when interacting with other Foreman users & developers:
+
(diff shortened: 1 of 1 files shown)


adamruzicka:
Hi, thank you for sending this in. We already discussed introducing a code of conduct into Foreman itself[1] in the past and eventually came to a conclusion that it is not really necessary. We already have community guidelines we adhere to, which are also reflected in the changes you are proposing.

If any such changes should be introduced, they should probably be introduced into the whole Foreman community and ecosystem, not just into selected plugins such as this one. Just out of curiosity, what piqued your interest in this repository?

[1] - https://github.com/theforeman/foreman/pull/4606

adamruzicka:
It has been two months already since this was opened without any response or activity. I'll go ahead and close this to reduce the number of stale PRs we have here, feel free to reopen.

PR #698, opened 2022-12-02
Store action names sorted

ekohl:
This makes the output stable and avoids moving lines around. That only creates noise in the git log.

### DIFF
--- a/lib/tasks/gettext.rake
+++ b/lib/tasks/gettext.rake
@@ -16,6 +16,7 @@ if gettext_find_task
                  .all_action_names
                  .uniq
                  .map { |n| %[_("#{n}")] }
+                 .sort
                  .join("\n")
     end
   end


adamruzicka (approved):
ack

adamruzicka:
Thank you @ekohl !

lib/tasks/gettext.rake
```diff
@@ -16,6 +16,7 @@ if gettext_find_task
                  .all_action_names
```
ekohl, line 16: I noticed in plugins that depend on other plugins with tasks (most notably, X depends on Katello) that it also lists all tasks from Katello. Is there a way this could be limited to one domain?

adamruzicka: Right now probably not or at least not cleanly. The action classes are not namespaced properly across projects nor do they have a reliable designator that we could use.

We could resort to something hacky like `Object.const_source_location(klass)` and deciding based on what it gives us, but I'm not sure I want to go down that route

ekohl: > but I'm not sure I want to go down that route

No. Perhaps a more explicit way to define it may be nice, or a way in the plugin DSL to define/register tasks.

But I'm sure you'll agree that it's annoying that every single plugin that defines tasks finds at least the 3 that are defined in `foreman-tasks` and those that also depend on Katello have way more.

PR #700, opened 2022-12-13
Fixes #35860 - Fix untranslatable strings

ekohl:
It is never valid to use ```__(``)``` since string extraction relies on static strings. String interpolation breaks this, so it's best to avoid it even if the string contained is static.

This also fixes a string where interpolation should happen. This allows translations to move around the word order. Ideally they would actually properly use pluralization so that it would be "select 1 task" instead of "select 1 tasks". That would also allow other languages to properly use plurals, but that's left as a future improvement.

### DIFF
--- a/webpack/ForemanTasks/Components/TaskDetails/Components/TaskButtons.js
+++ b/webpack/ForemanTasks/Components/TaskDetails/Components/TaskButtons.js
@@ -48,7 +48,7 @@ export const TaskButtons = ({
         <span
           className={`glyphicon glyphicon-refresh ${taskReload ? 'spin' : ''}`}
         />
-        {__(`${taskReload ? 'Stop' : 'Start'}  auto-reloading`)}
+        {taskReload ? __('Stop auto-reloading') : __('Start auto-reloading')}
       </Button>
       <Button
         className="dynflow-button"
--- a/webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
+++ b/webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
@@ -1,7 +1,7 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Tab, Tabs } from 'patternfly-react';
-import { translate as __ } from 'foremanReact/common/I18n';
+import { translate as __, sprintf } from 'foremanReact/common/I18n';
 import { STATUS } from 'foremanReact/constants';
 import MessageBox from 'foremanReact/components/common/MessageBox';
 import Task from './Components/Task';
@@ -49,7 +49,7 @@ const TaskDetails = ({
       <MessageBox
         key="task-details-error"
         icontype="error-circle-o"
-        msg={__(`Could not receive data: ${APIerror && APIerror.message}`)}
+        msg={sprintf(__('Could not receive data: %s'), APIerror?.message)}
       />
     );
   }
--- a/webpack/ForemanTasks/Components/TasksTable/Components/SelectAllAlert.js
+++ b/webpack/ForemanTasks/Components/TasksTable/Components/SelectAllAlert.js
@@ -1,3 +1,4 @@
+import { FormattedMessage } from 'react-intl';
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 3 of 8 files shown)


webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
```diff
@@ -49,7 +49,7 @@ const TaskDetails = ({
       <MessageBox
         key="task-details-error"
         icontype="error-circle-o"
-        msg={__(`Could not receive data: ${APIerror && APIerror.message}`)}
+        msg={sprintf(__('Could not receive data: %s'), APIerror?.message)}
```
ekohl, line 52: I think this is equivalent, but my JS knowledge is limited.

webpack/ForemanTasks/Components/TaskDetails/Components/TaskButtons.js
```diff
@@ -48,7 +48,7 @@ export const TaskButtons = ({
         <span
           className={`glyphicon glyphicon-refresh ${taskReload ? 'spin' : ''}`}
         />
-        {__(`${taskReload ? 'Stop' : 'Start'}  auto-reloading`)}
+        {taskReload ? __('Stop auto-reloading') : _('Start auto-reloading')}
```
ekohl, line 51: Introduced in 4efdfaabd0b420245c93f70dcf58e56c57167c29

webpack/ForemanTasks/Components/TasksTable/Components/SelectAllAlert.js
```diff
@@ -17,14 +17,13 @@ export const SelectAllAlert = ({
         Math.min(itemCount, perPage)
       )}
       <Button bsStyle="link" onClick={selectAllRows}>
-        {__('Select All')}
-        <b> {itemCount} </b> {__('tasks.')}
+        {sprintf(__('Select all %s tasks'), `<b>${itemCount}</b>`)}
```
ekohl, line 20: This is an example where proper pluralization should be used, but at least now there is a single sentence rather than a collection of words that may be moved around in other languages.

webpack/ForemanTasks/Components/TasksTable/Components/SelectAllAlert.js
```diff
@@ -17,14 +17,13 @@ export const SelectAllAlert = ({
...
       </Button>
     </React.Fragment>
   );
   const undoSelectText = (
     <React.Fragment>
-      {sprintf(__(`All %s tasks are selected. `), itemCount)}
+      {sprintf(__('All %s tasks are selected.'), itemCount)}
```
ekohl, line 26: This was translatable, but string interpolation is risky and it triggered my ``rg -l '__\(`'`` search. Also removes a trailing space from the translation. If JS needs the space, it should be done outside of the translation since a translator might do this for you and break the layout.

webpack/ForemanTasks/Components/TasksTable/TasksTable.js
```diff
@@ -68,7 +68,7 @@ const TasksTable = ({
       <MessageBox
         key="tasks-table-error"
         icontype="error-circle-o"
-        msg={__(`Could not receive data: ${error && error.message}`)}
+        msg={sprintf(__('Could not receive data: %s'), error?.message)}
```
ekohl, line 71: Same as above: please make sure this is valid JS.

ekohl (commented):
When reviewing this, please treat this like I don't know any JS at all because I rarely use it.

adamruzicka:
Test failures should be fixed by https://github.com/theforeman/foreman-tasks/pull/701

adamruzicka:
Right, now the failures are related.

webpack/ForemanTasks/Components/TasksTable/Components/SelectAllAlert.js
ekohl: Looks like `<b>` is getting escaped. How would I write this in a compatible way?

MariaAga (approved):
code is good and also tested and works as expected

adamruzicka:
Thank you @ekohl & @MariaAga !

PR #702, opened 2023-01-02
Fixes #35880 - Do not configure sidekiq sessions

adamruzicka:
Sidekiq >= 6.2.0 has reworked session handling, we shouldn't need to set anything and everything should just work out of the box.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -72,7 +72,6 @@ Foreman::Application.routes.draw do
         require 'sidekiq/web'
         redis_url = ENV['DYNFLOW_REDIS_URL'] || SETTINGS.dig(:dynflow, :redis_url)
         Sidekiq.redis = { url: redis_url }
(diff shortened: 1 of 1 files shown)


lhellebr:
I did this change on the installed Foreman nightly in `/usr/share/gems/gems/foreman-tasks-7.1.0/config/routes.rb` and the message is still there when restarting foreman services.

lhellebr:
I did the change on an already installed foreman nightly in `/usr/share/gems/gems/foreman-tasks-7.1.0/config/routes.rb` and the WARNING doesn't appear anymore.

ekohl (approved):
Since there is no explicit dependency on Sidekiq, what's the minimum version now?

adamruzicka:
6.2.0 to make the console mounted at /foreman_tasks/sidekiq work. For actual background processing, >= 6.y.z

PR #704, opened 2023-01-16
Fixes #35950 - Make delivery of ProxyActionStopped events optional

adamruzicka:
They are meant to catch cases when the remote task stops or disappears. If this happens but the local task is already stopped, it is not an error state.


ofedoren (approved):
Not sure how to reproduce, but this doesn't seem like breaking change, so LGTM.

PR #707, opened 2023-02-21
Add packit config

adamruzicka:


adamruzicka:
Maybe it can't be a draft?

ekohl:
Perhaps https://github.com/theforeman/foreman-packaging/pull/9073 helps. You could also consider dropping the other CI testing to reduce the load to iterate.

.packit.yaml
```diff
@@ -0,0 +1,39 @@
...
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: foreman-tasks
+# downstream (Fedora) RPM package name
+downstream_package_name: rubygem-foreman-tasks
+
+actions:
+  post-upstream-clone:
+    - "wget https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec -O rubygem-foreman-tasks.spec"
+    - "wget https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/plugins/rubygem-foreman-tasks/foreman-tasks.logrotate -O foreman-tasks.logrotate"
+  get-current-version: "grep -Po '\\d+\\.\\d+\\.\\d+' lib/foreman_tasks/version.rb"
```
evgeni, line 21:
```suggestion
  get-current-version: "ruby -I lib/ -r foreman_tasks/version -e 'puts ForemanTasks::VERSION'"
```

evgeni (approved):

.packit.yaml
```diff
@@ -0,0 +1,39 @@
...
+# name in upstream package repository or registry (e.g. in PyPI)
+upstream_package_name: foreman-tasks
+# downstream (Fedora) RPM package name
+downstream_package_name: rubygem-foreman-tasks
+
+actions:
+  post-upstream-clone:
+    - "wget https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/plugins/rubygem-foreman-tasks/rubygem-foreman-tasks.spec -O rubygem-foreman-tasks.spec"
+    - "wget https://raw.githubusercontent.com/theforeman/foreman-packaging/rpm/develop/packages/plugins/rubygem-foreman-tasks/foreman-tasks.logrotate -O foreman-tasks.logrotate"
+  get-current-version: "ruby -I lib/ -r foreman_tasks/version -e 'puts ForemanTasks::VERSION'"
```
ekohl, line 21: You could also consider a more generic approach of using the gemspec:
```suggestion
  get-current-version: "ruby -rrubygems -e 'puts Gem::Specification::load(Dir.glob("*.gemspec").first).version'"
```

That may make it easier if you want to copy it between various gems.

.packit.yaml
```diff
@@ -0,0 +1,39 @@
...
+  create-archive:
+    - gem build foreman-tasks.gemspec
+    - bash -c "ls -1t ./foreman-tasks-*.gem | head -n 1"
```
ekohl, line 24: gem build does output the filename so you can consider:
```suggestion
    - gem build foreman-tasks.gemspec | awk '/File: / { print $2 }'
```

But it may be better to explicitly determine the output filename:
```suggestion
    - gem build foreman-tasks.gemspec --output foreman-tasks.gem
    - echo foreman-tasks.gem
```

.packit.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      centos-stream-8:
+        additional_modules: "foreman:el8,ruby:2.7,nodejs:12"
```
ekohl, line 31: Keeping this in sync between repos is going to be a pain. Why do we need to list the modules that `foreman:el8` depends on?

evgeni: because `foreman:el8` does not depend on `nodejs:12`

evgeni: (correct for ruby tho)

.packit.yaml
evgeni: Ah nice, I didn't manage to do it like this.

.packit.yaml
adamruzicka: Doesn't the spec require the gem to be named like `%{gem_name}-%{version}`?

.packit.yaml
evgeni: (the quoting is off, tho)

.packit.yaml
evgeni: no, packit mangles the `Source0` line based on the output of this action :)

.packit.yaml
ekohl: So it doesn't install the ones we specify as build dependencies? It feels like it should so RFE time? :)

evgeni: My understanding is that the build dependencies we specify only apply in the context of "build inside the module", but we're not doing that.

ekohl: It kind of feels like we should. The target package is built inside the module, and we want to replicate that. Essentially I'm looking for the equivalent of `BuildRequires` for modularity. So something like `foreman:el8;with-build`. It feels like we're once again hitting an area of modularity where few have gone before.

evgeni: No, the target package is not built *inside* a module. It's built with modules enabled in the buildroot.

If we'd want to have those build *inside* a module we'd need MBS, which we neither have nor want.

ekohl: I meant that we want to replicate the build environment in terms of packages available.

PR #709, opened 2023-02-22
simplify packit config

evgeni:
- no need to depend on the ruby module, foreman does already for us
- get current version using rubygems directly


evgeni:
for now I did not do the filename part from https://github.com/theforeman/foreman-tasks/pull/707#discussion_r1114136812 as I didn't like to make the filename completely static, and the awk pipe didn't work out inside python

adamruzicka:
@ekohl could we get your blessing here?

ekohl (approved):

adamruzicka:
Thank you @evgeni & @ekohl !

PR #710, opened 2023-03-01
Replace parse-cron with fugit

adamruzicka:
`parse-cron` seems to be unmaintained since 2016, `fugit` seems to be still active. In addition to that, it should also allow expressing intervals such as "first Monday each month" which wasn't possible with `parse-cron`.

TODO:
- [x] packaging

### DIFF
--- a/foreman-tasks.gemspec
+++ b/foreman-tasks.gemspec
@@ -27,8 +27,8 @@ same resource. It also optionally provides Dynflow infrastructure for using it f
   s.extra_rdoc_files = Dir['README*', 'LICENSE']
 
   s.add_dependency "dynflow", '>= 1.6.0'
+  s.add_dependency 'fugit', '~> 1.8'
   s.add_dependency "get_process_mem" # for memory polling
-  s.add_dependency "parse-cron", '~> 0.1.4'
   s.add_dependency "sinatra" # for Dynflow web console
 
   s.add_development_dependency 'factory_bot_rails', '~> 4.8.0'
--- a/app/models/foreman_tasks/recurring_logic.rb
+++ b/app/models/foreman_tasks/recurring_logic.rb
@@ -1,5 +1,5 @@
 module ForemanTasks
-  require 'parse-cron'
+  require 'fugit'
 
   class RecurringLogic < ApplicationRecord
     include Authorizable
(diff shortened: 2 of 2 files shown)


foreman-tasks.gemspec
```diff
@@ -27,8 +27,8 @@ same resource. It also optionally provides Dynflow infrastructure for using it f
   s.extra_rdoc_files = Dir['README*', 'LICENSE']
 
   s.add_dependency "dynflow", '>= 1.6.0'
+  s.add_dependency 'fugit', '~> 1.8.1'
```
ekohl, line 30: This pins to 1.8.x. Would it be better to use this pin, or do they not respect semver?
```suggestion
  s.add_dependency 'fugit', '>= 1.8.1', '< 2'
```

ekohl (approved):
Dependency should be available: https://github.com/theforeman/foreman-packaging/pull/9208. I suspect packit fails because fugit is not listed as a BuildRequires, but we can't deal with that on EL8. For EL9 we may have https://github.com/theforeman/foreman-packaging/pull/8414, but for now we'll have to accept the limitation.

@evgeni perhaps we can patch the spec file based on the gem by autoregenerating the deps, similar to what we do in `bump_rpm.sh`.

evgeni:
@ekohl we certainly can (and [packit has a hook for that](https://packit.dev/docs/actions/#fix-spec-file)), just nobody sat down and did *innocent whistle*

adamruzicka:
Thank you @ekohl !

sjha4:
I am seeing some change in behavior with fugit.

```
[1] pry(main)> require 'fugit'
=> false
[2] pry(main)> @parser ||= Fugit.parse_cron('* * * * abc')
=> nil
[3] pry(main)> @parser ||= Fugit.parse_cron(nil)
=> nil
[4] pry(main)> @parser ||= Fugit.parse_cron('* * * * *')
=> #<Fugit::Cron:0x000056255ad0a590 @cron_s=nil, @day_and=nil, @hours=nil, @minutes=nil, @monthdays=nil, @months=nil, @original="* * * * *", @seconds=[0], @timezone=nil, @weekdays=nil, @zone=nil>
```

What that does is have nil failures in `next_occurrence_time` method. Here is the workflow that raises that..

```
[5] pry(main)> recurring_logic = ::ForemanTasks::RecurringLogic.new_from_cronline('* * * * abc')
=> #<ForemanTasks::RecurringLogic:0x000056255ac31fb0 id: nil, cron_line: "* * * * abc", end_time: nil, max_iteration: nil, iteration: 0, task_group_id: nil, state: nil, triggering_id: nil, purpose: nil>
[6] pry(main)> recurring_logic.valid_cronline?
NoMethodError: undefined method `previous_time' for nil:NilClass
from /home/vagrant/foreman/.vendor/ruby/2.7.0/gems/foreman-tasks-8.0.0/app/models/foreman_tasks/recurring_logic.rb:101:in `next_occurrence_time'
```

PR #712, opened 2023-04-11
GH actions tweaks

adamruzicka:
- cancel running jobs when new changes are pushed
- timeout after 15 minutes


adamruzicka:
Well, it is sad that the tests didn't pass, but at least it has shown it works.

PR #714, opened 2023-05-04
set upstream_tag_template in packit config

evgeni:
this allows to find the "last" release and produce a more meaningful changelog


adamruzicka:
Thank you @evgeni !

PR #716, opened 2023-05-05
Fixes #36351 - Send out email notification when tasks linger in running or paused for longer than two days

adamruzicka:
- [x] users should be able to opt-out/in
  - users are able to opt-in/opt-out in email preferences
  - toast notification is now tied to the email notification which isn't exactly great, but we don't really have any other means of configuring it
- [x] always enabled ~~by default~~ for admin
  - ~~this is tricky. I thought I'd add a migration, but the migration needs to run after data (the template) is seeded. Currently the migration creates the template and then the seeds override it. I'm not particularly fond of this, but I don't really see another way~~
- [x] honor taxonomy boundaries
- [x] time should be configurable
  - done, sort of. We don't really have the means for doing this nicely as recurring logics cannot be edited
  - by default set to midnight
  - However, as a workaround the RL can be cancelled, env variable `FOREMAN_TASKS_CHECK_LONG_RUNNING_TASKS_CRONLINE` can be set to a new cronline and then on service restart the RL should get re-created with the new schedule
  - Alternatively `rake foreman_tasks:reschedule_long_running_tasks_checker FOREMAN_TASKS_CHECK_LONG_RUNNING_TASKS_CRONLINE=$cron` can be used
- [x] have a recurring task monitor the rest of tasks

### DIFF
--- a/app/lib/actions/check_long_running_tasks.rb
+++ b/app/lib/actions/check_long_running_tasks.rb
@@ -1,5 +1,7 @@
 module Actions
   class CheckLongRunningTasks < ::Actions::EntryAction
+    include Actions::RecurringAction
+
     INTERVAL = 2.days
     STATES = ['running', 'paused'].freeze
 
--- a/app/lib/actions/check_long_running_tasks.rb
+++ b/app/lib/actions/check_long_running_tasks.rb
@@ -1,0 +1,39 @@
+module Actions
+  class CheckLongRunningTasks < ::Actions::EntryAction
+    INTERVAL = 2.days
+    STATES = ['running', 'paused'].freeze
+
+    def plan
+      time = Time.now.utc
+      cutoff = time - INTERVAL
+      notification = ::ForemanTasks::TasksMailNotification.find_by(name: "long_running_tasks")
+      users = User.where(id: UserMailNotification.where(mail_notification_id: notification.id))
+
+      query = "state ^ (#{STATES.join(', ')}) AND state_updated_at <= \"#{cutoff}\""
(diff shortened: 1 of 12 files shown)


ofedoren (commented):
Thanks, @adamruzicka, code-wise LGTM, apart one small inline suggestion.

> However, as a workaround the RL can be cancelled, env variable FOREMAN_TASKS_CHECK_LONG_RUNNING_TASKS_CRONLINE can be set to a new cronline and then on service restart the RL should get re-created with the new schedule

I wouldn't like server restart as the only option to change the task as a user :/ Could we maybe write this logic into a Rake task at least? I mean, we can? find the recurring task and stop? it and create a new one with the same `ForemanTasks.register_scheduled_task(Actions::CheckLongRunningTasks, ENV['FOREMAN_TASKS_CHECK_LONG_RUNNING_TASKS_CRONLINE'] || '0 0 * * *')`?

Also, a question for dummies: how can I test this out and not wait for 2 days?..

adamruzicka:
Ermm, what?

adamruzicka:
I don't really see a sensible way how to have enable it by default for admin but make them able to opt-out. The migration idea falls flat because the user is added from seeds. If we put this into seeds then they won't really be able to opt-out.

adamruzicka:
Alright, after some internal discussions, we agreed to leave this always enabled for admins, where admins are either users explicitly marked as admins or users which have the org admin role assigned

ofedoren (approved):
Thanks, @adamruzicka, LGTM and thank you for waiting so long :/

PR #721, opened 2023-07-20
Fixes #36606 - Properly translate headers

adamruzicka:
in long running tasks email notification

### DIFF
--- a/app/views/tasks_mailer/long_tasks.html.erb
+++ b/app/views/tasks_mailer/long_tasks.html.erb
@@ -8,11 +8,11 @@
 <div class="dashboard">
   <table>
     <tr>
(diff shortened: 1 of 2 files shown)


ofedoren (commented):
Thanks, @adamruzicka, sorry for missing this. Do we also want to translate these:
 - https://github.com/theforeman/foreman-tasks/blob/master/app/views/tasks_mailer/long_tasks.html.erb#L29
 - https://github.com/theforeman/foreman-tasks/blob/master/app/views/tasks_mailer/long_tasks.text.erb#L9-L14

ofedoren (approved):
Thanks, @adamruzicka, LGTM.

PR #722, opened 2023-07-21
Fixes #36700 - Switch to v2 of sub plans

adamruzicka:
In addition to migrating to a newer implementation, it also exposes concurrency limiting for sub-plans spawned by `::Actions::BulkAction`

Requires:
- https://github.com/Dynflow/dynflow/pull/428

TODO:
- [x] Bump dependency on dynflow

### DIFF
--- a/app/lib/actions/action_with_sub_plans.rb
+++ b/app/lib/actions/action_with_sub_plans.rb
@@ -1,6 +1,6 @@
 module Actions
   class Actions::ActionWithSubPlans < Actions::EntryAction
-    include Dynflow::Action::WithSubPlans
+    include Dynflow::Action::V2::WithSubPlans
 
     def plan(*_args)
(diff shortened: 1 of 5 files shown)


ofedoren:
I guess rubygems didn't update the cache for the new version of dynflow, so the test run with the old version, could you try to re-trigger the tests, just so we're green?

Also, what about TBFL and draft state? :)

adamruzicka:
Mmm, apparently we test the internals...

adamruzicka:
No, or at least not like this. I expected there might be some issues with the newer dynflow not being packaged yet, but I don't think it would manifest itself this way.

evgeni:
/packit build

now it should pass

ofedoren (approved):
Thanks, @adamruzicka and @evgeni, let's get this in.

PR #725, opened 2023-09-14
Fixes #36746 - Make email delivery skip on failures

adamruzicka:
Without this change the execution plan would go into paused and then would be picked up by subsequent checks, which would then go into paused and so on and so on.


ofedoren (approved):
Thanks, @adamruzicka, let's get this in. Sorry for waiting :/

PR #726, opened 2023-09-20
Fixes #36764 - Emit failure events from ObservableAction

adamruzicka:


ofedoren (approved):
Hmm.. Did we really support subscription to `_failed`, but never fired them? If so, it's funny, because the same, but opposite of that I've made [recently](https://projects.theforeman.org/issues/36796) as well :/ 

Thanks, @adamruzicka, checked with webhooks, works as expected.

PR #727, opened 2023-09-27
Fixes #20391 - Hide breadcrumb title from Tasks page

kmalyjur:
The Tasks page had multiple titles because of the usage of breadcrumb as a title in the PageLayout. It is dependent on https://github.com/theforeman/foreman/pull/9830.

image

### DIFF
--- a/lib/foreman_tasks/engine.rb
+++ b/lib/foreman_tasks/engine.rb
@@ -18,7 +18,7 @@ module ForemanTasks
 
     initializer 'foreman_tasks.register_plugin', :before => :finisher_hook do |_app|
       Foreman::Plugin.register :"foreman-tasks" do
-        requires_foreman '>= 3.7.0'
+        requires_foreman '>= 3.9'
         divider :top_menu, :parent => :monitor_menu, :last => true, :caption => N_('Foreman Tasks')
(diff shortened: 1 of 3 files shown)


adamruzicka:
On the tasks page it fixes the issue, on a sub tasks page we get this
image

For reference, before the double title fix in foreman we had this
image

I can't really say which one is correct

kmalyjur:
> I can't really say which one is correct
@adamruzicka I think it was just a breadcrumb or just a title before, however, in my opinion, it's nicer and more consistent having it the first way. I'm open to more opinions of course. :)

pnovotny:
LGTM

The Tasks page is fixed (verified on downstream, therefore the different look):
tasks-fixed
ixed:

The Sub tasks page looks like on the first screenshot posted in https://github.com/theforeman/foreman-tasks/pull/727#issuecomment-1736916434
IMO it looks fine with the breadcrumb + title, but of course it's a matter of personal preference.

adamruzicka (approved):
Looks good to me, but let's keep it around for a bit until other things which don't require foreman 3.9 get in

adamruzicka:
Thank you @kmalyjur and sorry for the delay

PR #728, opened 2023-10-19
Fixes #36848 - Rework long running task watching setup

adamruzicka:
- enabled for admin-like users on create
- enabled for admin-like users on seed when the notifications is first created, excluding hidden admins
- delivery is not attempted for disabled users
- delivery is not attempted for users without email set
- delivery is attempted only for users with mail_enabled set

### DIFF
--- a/app/lib/actions/check_long_running_tasks.rb
+++ b/app/lib/actions/check_long_running_tasks.rb
@@ -8,12 +8,10 @@ module Actions
     def plan
       time = Time.now.utc
       cutoff = time - INTERVAL
-      notification = ::ForemanTasks::TasksMailNotification.find_by(name: "long_running_tasks")
-      org_admin_role = Role.find_by(name: 'Organization admin')
-      users = User.left_joins(:roles)
-                  .where(id: UserMailNotification.where(mail_notification_id: notification.id).select(:role_id))
-                  .or(User.where(admin: true))
-                  .or(User.where(id: UserRole.where(id: [org_admin_role.id] + org_admin_role.cloned_role_ids).select(:owner_id)))
+      users = User.joins(:mail_notifications)
+                  .where(mail_enabled: true, mail_notifications: { name: 'long_running_tasks' })
+                  .where.not(mail: [nil, ''])
+                  .where(disabled: [nil, false])
 
       query = "state ^ (#{STATES.join(', ')}) AND state_updated_at <= \"#{cutoff}\""
       users.each do |user|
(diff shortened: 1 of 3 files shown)


app/lib/actions/check_long_running_tasks.rb
```diff
@@ -8,12 +8,10 @@ class CheckLongRunningTasks < ::Actions::EntryAction
...
       cutoff = time - INTERVAL
-      notification = ::ForemanTasks::TasksMailNotification.find_by(name: "long_running_tasks")
-      org_admin_role = Role.find_by(name: 'Organization admin')
-      users = User.left_joins(:roles)
-                  .where(id: UserMailNotification.where(mail_notification_id: notification.id).select(:role_id))
-                  .or(User.where(admin: true))
-                  .or(User.where(id: UserRole.where(id: [org_admin_role.id] + org_admin_role.cloned_role_ids).select(:owner_id)))
+      users = User.joins(:mail_notifications)
+                  .where(mail_enabled: true, mail_notifications: { name: 'long_running_tasks' })
+                  .where.not(mail: nil)
```
ares, line 13: for the user I create through webui the mail is actually `""`, because the value comes from empty input, I think we need to skip those as well as there's a valid way to get to such state

ares (changes requested):
There's one edge case, otherwise looks good to me, could you please update this before I proceed with the testing?

db/seeds.d/95-mail_notifications.rb
```diff
@@ -24,8 +24,10 @@
     org_admin_role = Role.find_by(name: 'Organization admin')
 
     users = User.left_joins(:roles)
+                .joins(:auth_source)
                 .where(admin: true)
                 .or(User.where(id: UserRole.where(id: [org_admin_role.id] + org_admin_role.cloned_role_ids).select(:owner_id)))
+                .where.not(auth_source: { name: 'Hidden' })
```
ares, line 30: there are other potentially hiddden auth sources, e.g. https://github.com/theforeman/foreman_virt_who_configure/blob/master/app/models/foreman_virt_who_configure/auth_source_hidden_with_authentication.rb

I don't think it would cause problems though, since the user [is not](https://github.com/theforeman/foreman_virt_who_configure/blob/master/app/models/foreman_virt_who_configure/config.rb#L174-L192) admin, so that's just FYI

adamruzicka: Good to know, although even if we did set it up for hidden users, as long as they don't have mail enabled and email address set we would not attempt delivery to them anyway

ares (approved):
Tested thoroughly, all edgecases I could have thought off were working as expected. Thanks for the quick fix, happy to merge when correct Redmine is linked.

PR #730, opened 2023-11-01
i18n - pulling from tx

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka!

It seems like with REX, we'd need to try to re-pull new translations a bit later, since we've got added few more new strings.

PR #731, opened 2023-11-20
Fixes #36931 - Pass keywords around the old way

adamruzicka:
This is not ideal, but it should be in line with what we had before the move to v2 bulk actions.

To handle this properly, the changes will need to start in Dynflow and then in things that build on top of it. This is something we might ultimately have to do for Ruby 3.


parthaa (approved):
Tested on a couple of bulk actions in Katello. ACK

ofedoren (approved):
Thanks, @adamruzicka, seems reasonable and easier than rewriting more sophisticated stuff :)

PR #732, opened 2023-11-20
Use foreman-devel:el8 module

adamruzicka:


adamruzicka (commented):
We've done this over and over in other repos already. Packit is green, let's go

PR #733, opened 2023-11-22
Fixes #36937 - Rework proxy batch triggering

adamruzicka:
Now it should be more strict about when it finishes. Previously this action could finish too early, even before everything was dispatched to the proxy.


ofedoren (approved):
Thanks @adamruzicka, LGTM. The good old boolean flag is always the way to go :)

PR #735, opened 2023-11-28
Generate translations for actions only coming from the current plugin

adamruzicka:


ekohl (approved):
I tested this out and in foreman_virt_who_configure (which depends on katello) it generates:
```ruby
# Autogenerated!
_("Report")
```

PR #736, opened 2024-01-08
Fixes #37036 - Hide next occurrence unless RL is active

adamruzicka:
in API


ofedoren (approved):
Thanks, @adamruzicka !

PR #737, opened 2024-01-12
Test on ruby 3

adamruzicka:


ekohl:
Why weren't https://github.com/theforeman/foreman-tasks/pull/739 and https://github.com/theforeman/foreman-tasks/pull/740 uncovered by this? Is that a lack of testing coverage?

adamruzicka:
My take is that dynflow and tasks didn't use keywords at all, so it was ruby 3 compatible and this passed. Once we started adding keywords, all the small differences between how 2.7 and 3.0 handle them started to pop up

PR #743, opened 2024-02-09
Ruby 3 minimal changes

ofedoren:


app/lib/actions/bulk_action.rb
```diff
@@ -9,7 +9,8 @@ class BulkAction < Actions::ActionWithSubPlans
     #   Arguments that all the targets share
     def plan(action_class, targets, *args, concurrency_limit: nil, **kwargs)
       check_targets!(targets)
-      limit_concurrency_level!(concurrency_limit) if concurrency_limit
+      extracted_concurrency_limit = extract_concurrency_limit(args, concurrency_limit)
+      limit_concurrency_level!(extracted_concurrency_limit) if extracted_concurrency_limit
```
adamruzicka, line 13: I think the concurrency_limit keyword is not used anywhere yet, so we might be able to change the interface if that would help

ofedoren:
Jeez.. 2.7 failure is just... such random. I'll rebase the branch just in case :/

app/lib/actions/bulk_action.rb
adamruzicka: Now I'm confused. It must have worked somehow otherwise the test that is now failing on 2.7 would never pass have passed, no?

app/lib/actions/bulk_action.rb
```diff
@@ -72,5 +73,11 @@ def batch(from, size)
     def total_count
       input[:target_ids].count
     end
+
+    private
+
+    def extract_concurrency_limit(args = [], limit = nil)
+      args.find { |arg| arg.is_a?(Hash) && arg.key?(:concurrency_limit) }&.fetch(:concurrency_limit, limit)
```
evgeni, line 80: My Ruby is bad, but doesn't this mean:
1. take `args`
2. find the first element in it that is a has and has a `concurrency_limit` key
3. if found, fetch the value. if not found, don't do anything as `find` returned `nil` and the safe operator skips the `.fetch`

so instead, how about:
```suggestion
      args.find { |arg| arg.is_a?(Hash) && arg.key?(:concurrency_limit) }&.fetch(:concurrency_limit) || limit
```

evgeni:
Also, no pressure, but this PR is the only thing needed to be able to deploy (not pass tests, alas) a Katello on EL9 :)

app/lib/actions/bulk_action.rb
evgeni: (and depending on the Ruby version, calling with `(…, concurrency_limit: 25)` places that in either `args` or `concurrency_limit` and produces 2.7 failures

evgeni:
The tests are passing, and Katello seems to work fine with these changes (even bats is passing once the Ruby3 patch for Katello is applied) \o/

adamruzicka (approved):

PR #749, opened 2024-03-20
Fixes #37285 - Properly process data we get from checking on proxy 

adamruzicka:


app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
```diff
@@ -46,7 +46,7 @@ def process_task_results(tasks, results)
         notify ::Actions::ProxyAction::ProxyActionMissing.new, missing if missing.any?
 
         stopped = present.select { |task| %w[stopped paused].include? results.dig(task.remote_task_id, 'state') }
-        notify ::Actions::ProxyAction::ProxyActionStopped.new, stopped if stopped.any?
```
ofedoren, line 49: Do I understand right, that this should correspond with https://github.com/theforeman/foreman-tasks/blob/124e9c64296c4ed997eabfdc2ad6c2384d4c0d6a/app/lib/actions/proxy_action.rb#L48-L69 ?

app/lib/actions/proxy_action.rb
```diff
@@ -93,15 +93,8 @@ def proxy_input(task_id = task.id)
 
     def check_task_status
       response = proxy.status_of_task(proxy_task_id)
-      if %w[stopped paused].include? response['state']
-        if response['result'] == 'error'
-          raise ::Foreman::Exception, _('The smart proxy task %s failed.') % proxy_task_id
```
ofedoren, line 98: Where this will be checked instead? Or we don't need this?

app/lib/actions/middleware/watch_delegated_proxy_sub_tasks.rb
adamruzicka: Yes, exactly

app/lib/actions/proxy_action.rb
adamruzicka: I might have taken a shortcut here that I can't really take.

There's the ::Actions::ProxyAction in here, which (as far as I know) is only used in remote execution. Remote execution has a subclass which inherits from ::Actions::ProxyAction and adds a bunch of extensions. The changes here rely on those extensions. Thank you for pointing this out

adamruzicka: I kept it mostly intact here and extracted the rex specific parts to https://github.com/theforeman/foreman_remote_execution/pull/900

adamruzicka:
@ofedoren anything else to do here?

ofedoren (approved):
Sorry, @adamruzicka, let's get it.

PR #752, opened 2024-05-14
Fixes #37452 - Support Zeitwerk loader

ofedoren:
~WIP~

### DIFF
--- a/lib/foreman_tasks/engine.rb
+++ b/lib/foreman_tasks/engine.rb
@@ -16,116 +16,118 @@ module ForemanTasks
       SETTINGS[:foreman_tasks] = { :assets => { :precompile => assets_to_precompile } }
     end
 
-    initializer 'foreman_tasks.register_plugin', :before => :finisher_hook do |_app|
-      Foreman::Plugin.register :"foreman-tasks" do
-        requires_foreman '>= 3.9'
-        divider :top_menu, :parent => :monitor_menu, :last => true, :caption => N_('Foreman Tasks')
-        menu :top_menu, :tasks,
-             :url_hash => { :controller => 'foreman_tasks/tasks', :action => :index },
-             :caption  => N_('Tasks'),
-             :parent   => :monitor_menu,
-             :last     => true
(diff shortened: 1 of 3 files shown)


lib/foreman_tasks/engine.rb
```diff
@@ -186,6 +185,16 @@ class Engine < ::Rails::Engine
       ::Dynflow::ActiveJob::QueueAdapters::JobWrapper.include Actions::TaskSynchronization
     end
 
+    initializer "foreman_tasks.observable_events" do
```
ofedoren, line 188: Can be dropped from this PR.

That's my attempt to fix at least some `autoloadable constant loading during initialization` warnings. Seems I was able to fix some of them in the plain Foreman, but with each plugin the number of warnings rises exponentially. Not required for Zeitwerk, but in the future Rails will make these warnings be errors.

ekohl (commented):
Should we make `ID` an acronym to avoid renaming?

lib/foreman_tasks/engine.rb
ekohl: Is that something we can already merge today or wouldn't it make sense without Zeitwerk?

ofedoren:
> Should we make `ID` an acronym to avoid renaming?

We could, but it will imply that we would need to either make custom inflector for the plugin, which won't affect core and other plugins or make it acronym in the core, so we have consistency. The later will bring even more changes across the plugins :/

lib/foreman_tasks/engine.rb
ofedoren: That's experimental and I'll drop this from the PR.

I think we could deal with the warnings without Zeitwerk enabled, but the warnings appear only if we enable it. I'd postpone that after we enable it for all the plugins, maybe devs will get more interested in resolving that with us.

ofedoren:
Ready to merge, dropped the commit for testing with Foreman PR, the latest run was green:
Screenshot_20240903_143351

adamruzicka (approved):

adamruzicka:
Thank you @ofedoren !

PR #753, opened 2024-06-27
Fixes #37605 - missing formatting in Step Canceled toast

MariaAga:
missed a `$`


adamruzicka:
Thank you @MariaAga !

PR #754, opened 2024-06-28
Fixes #37608 - cleanup TaskDetailsActions.test.js api mock

MariaAga:
its not needed as the mock is loaded from webpack/__mocks__/foremanReact/redux/API/index.js


adamruzicka:
Thank you @MariaAga !

PR #756, opened 2024-07-26
Fixes #37688 - Do not build action names without actions

adamruzicka:
If the file already exists but would be empty, it gets deleted.


ofedoren (approved):
Thanks, @adamruzicka, LGTM!

PR #761, opened 2024-09-19
Fix gettext:store_action_names

adamruzicka:
The Object.const_source_location stopped being reliable since the zeitwerk changes landed. Each action is expected to define at least one of the three methods, so a source location of a method should be a reliable indicator of where the action comes from.

An alternative to #759


ianballou (commented):
This fixed the following:

```
RAILS_ENV=production bundle exec rake plugin:gettext[katello]
```

ekohl (approved):

ofedoren (approved):
Thanks, @adamruzicka, @ianballou and @ekohl, let's get this in then :)

PR #762, opened 2024-09-20
Fixes #37846 - Support Rails 7.0

ofedoren:


ofedoren (commented):
I'm afraid this PR depends on Foreman's one, I don't seem to find `to_fs` on TimeWithZone in Rails 6.1 :/

lib/foreman_tasks/cleaner.rb
```diff
@@ -252,7 +252,7 @@ def orphaned_dynflow_tasks
 
     def prepare_filter
       filter_parts = [filter]
-      filter_parts << %(started_at < "#{after.ago.to_s(:db)}") if after > 0
+      filter_parts << %(started_at < "#{after.ago.to_fs(:db)}") if after > 0
```
ekohl, line 255: Rails 6.1 already has [to_formatted_s](https://edgeapi.rubyonrails.org/classes/DateTime.html#method-i-to_formatted_s). It's there an alias to `to_s` while in 7.0 and newer it's an alias to `to_fs`. That means you can be compatible with both.

ekohl (changes requested):

lib/foreman_tasks/cleaner.rb
ofedoren: It's even more readable name, updated.

Also, dropped the last commit, so it's gtg (if :green_apple:).

adamruzicka:
Thank you @ofedoren & @ekohl !

PR #763, opened 2024-11-11
Fixes #37938 - Reject start-at dates in the past when a job is scheduled

pafernanr:


app/models/foreman_tasks/triggering.rb
```diff
@@ -121,5 +122,10 @@ def can_start_future
       parse_start_at!
       errors.add(:start_before_raw, _('The task could not be started')) if !start_before.nil? && start_before < start_at
     end
+
+    def start_at_is_not_past
+      errors.add(:start_at, _('is in the past')) if !start_at.nil? && Time.zone.parse(start_at_raw) < Time.zone.now
+    end
   end
 end
+future
```
adamruzicka, line 131: This shouldn't be there

adamruzicka (commented):
Code changes look reasonable, apart from that one thing.

image

@MariaAga how much work would it be to mark the date and time fields as incorrect? Also, I vaguely recall the form actively checking that the date is in the future. Has it stopped working in the meantime or am i misremembering?

MariaAga:
> @MariaAga how much work would it be to mark the date and time fields as incorrect? Also, I vaguely recall the form actively checking that the date is in the future. Has it stopped working in the meantime or am i misremembering?``

We have it only for starts before, we can probably easliy copy it to starts after as well

adamruzicka:
@pafernanr could you please add the last newline so that the linter is happy?

adamruzicka:
@pafernanr ping

pafernanr:
Hi, sorry for late response. 
I was checking the tests and I think the culprit is https://github.com/theforeman/foreman-tasks/blob/master/test/unit/triggering_test.rb#L12 

If I understand it well, that test is exactly executing a job in the past (-120). Dunno why, could be past-times needed for any reason?

adamruzicka:
> If I understand it well, that test is exactly executing a job in the past (-120). Dunno why, could be past-times needed for any reason?

I can't think of anything. The only thing we did there was that jobs scheduled for the past started as soon as possible, making them essentially the same as immediate jobs

pafernanr:
In that case I removed that test line as it is no longer needed. Executing past dates returns an error which make Unit test to fail.

pafernanr:
hmm, I thought it is better idea to keep the test but modifying it in order to check pasta start_at dates returns false. Please check triggering_test.rb. Does it makes sense for you?

adamruzicka:
I took the liberty rebasing this, let's see how this goes.

PR #764, opened 2024-11-11
Fix exception name

m-bucher:
Not sure if this requires opening an issue :grin:

### DIFF
--- a/lib/foreman_tasks.rb
+++ b/lib/foreman_tasks.rb
@@ -33,7 +33,7 @@ module ForemanTasks
                 finished.wait(timeout)
                 task = ForemanTasks::Task::DynflowTask.where(:external_id => id).first
                 if task.nil? || (!task.paused? && task.pending?)
-                  raise TimeoutError, "The time waiting for task #{task.try(:id)} to finish exceeded the 'foreman_tasks_sync_task_timeout' (#{timeout}s)"
+                  raise Timeout::Error, "The time waiting for task #{task.try(:id)} to finish exceeded the 'foreman_tasks_sync_task_timeout' (#{timeout}s)"
                 end
               end
               task || ForemanTasks::Task::DynflowTask.where(:external_id => id).first!


adamruzicka:
Hrmm, it has been there for ages, has it stopped working?

m-bucher:
I guess you do not really see the difference, because if this is hit then you get an exception, just not the one that is expected, but a `NameError`:
```
home/vagrant/foreman/.vendor/ruby/3.0.0/gems/dynflow-1.9.0/lib/dynflow/dispatcher/client_dispatcher.rb:258:in `with_ping_request_caching'
/home/vagrant/foreman/.vendor/ruby/3.0.0/gems/dynflow-1.9.0/lib/dynflow/dispatcher/client_dispatcher.rb:116:in `publish_request'
[ concurrent-ruby ]
rake aborted!
NameError: uninitialized constant ForemanTasks::TimeoutError
/home/vagrant/foreman/.vendor/ruby/3.0.0/gems/foreman-tasks-10.0.1/lib/foreman_tasks.rb:36:in `block (2 levels) in trigger_task'
/home/vagrant/foreman/.vendor/ruby/3.0.0/gems/algebrick-0.7.5/lib/algebrick/matchers/abstract.rb:74:in `assigns'
/home/vagrant/foreman/.vendor/ruby/3.0.0/gems/algebrick-0.7.5/lib/algebrick/matching.rb:56:in `match_value'
/home/vagrant/foreman/.vendor/ruby/3.0.0/gems/algebrick-0.7.5/lib/algebrick/matching.rb:36:in `block in match?'
/home/vagrant/foreman/.vendor/ruby/3.0.0/gems/algebrick-0.7.5/lib/algebrick/matching.rb:35:in `each'
```
It might also be that this error is not encountered that often. In my case it was some weird problem in my forklift dev-environment :sweat_smile: 

The only way this would really show is if you explicitly tried to `rescue` this specific exception :thinking:

adamruzicka:
Oh, I think I see it. Are you on ruby >=3.1? `TimeoutError` seems to have been removed[1] in timeout-0.2.0 which is shipped with ruby-3.1 

[1] - https://github.com/ruby/timeout/commit/14d421252b54027108da183783a7a26d0cf5aac1

adamruzicka (approved):
Anyway, TimeoutError used to be an alias for Timeout::Error, let's get thin

adamruzicka:
Thank you @m-bucher !

PR #768, opened 2025-06-16
Fixes #37938 - Reject start at in the past

adamruzicka:
Successor to #763 , hopefully with green tests


ofedoren (approved):
Thanks, @pafernanr and @adamruzicka !

PR #769, opened 2025-06-16
use @theforeman/actions for JS tests

MariaAga:


.github/workflows/js_tests.yml
```diff
@@ -1,33 +1,14 @@
...
+
 on:
+  push:
+    branches:
+      - master
   pull_request:
-    paths:
-      - 'webpack/**'
-      - 'package.json'
-      - '.github/workflows/js_tests.yml'
```
ofedoren, line 7: This would mean we want to run JS tests on every PR, is that correct?

MariaAga: yes, I trusted https://github.com/theforeman/foreman_remote_execution/pull/964/files on this

ofedoren (approved):
Thanks, @MariaAga !

PR #774, opened 2025-07-30
Fixes #38626 - Hide Sinatra error pages in production

adamruzicka:
Before
image

After
image

### DIFF
--- a/lib/foreman_tasks/dynflow.rb
+++ b/lib/foreman_tasks/dynflow.rb
@@ -18,6 +18,11 @@ module ForemanTasks
           { _('Back to tasks') => "/#{ForemanTasks::TasksController.controller_path}" }
         end
         set(:world) { ::Rails.application.dynflow.world }
+        # Do not show Sinatra's pretty error pages in production
+        set(:show_exceptions) { ::Rails.env.development? }
+        # Let the errors propagate out from Sinatra to Rails.
+        # Without this, we'd only get a mostly blank 500 ISE page
+        set(:raise_errors) { !::Rails.env.development? }
       end
     end
   end


lib/foreman_tasks/dynflow.rb
```diff
@@ -18,6 +18,11 @@ def web_console
           { _('Back to tasks') => "/#{ForemanTasks::TasksController.controller_path}" }
         end
         set(:world) { ::Rails.application.dynflow.world }
+        # Do not show Sinatra's pretty error pages in production
+        set(:show_exceptions) { ::Rails.env.development? }
+        # Let the errors propagate out from Sinatra to Rails.
+        # Without this, we'd only get a mostly blank 500 ISE page
+        set(:raise_errors) { !::Rails.env.development? }
```
ofedoren, line 25: Do we have a useful page then? I mean either a page which shows "user digestible" error or a page with redirection maybe?..

In your `After` screenshot I see a mostly blank page. Is that this one mentioned here or with this we'll get a different one?..

adamruzicka: The root cause is that if you hit `/foreman_tasks/dynflow/$whatever`, it goes to dynflow's sinatra app, which treats `$whatever` as an execution plan uuid. If this can't be found, it raises an exception. By default, sinatra catches this and displays a nice page.

> Do we have a useful page then?

Not really, I'm afraid we'd have to roll our own.

> page with redirection maybe?..

If you explicitly go to a bogus url, where would you expect that redirect go?

> Is that this one mentioned here or with this we'll get a different one?..

That's the railsy-one we get with raise_errors=true. With it being false we'd get just a page with a single line saying "Internal server error"

ofedoren: > If you explicitly go to a bogus url, where would you expect that redirect go?
> 

https://github.com/boguspage :D In other words, the app shouldn't be seen as "down" or there is a problem within the app.

> That's the railsy-one we get with raise_errors=true. With it being false we'd get just a page with a single line saying "Internal server error"

Eh.. But that's something should be treated in the main app, so plugins can re-use it.

ofedoren (approved):
Thanks, @adamruzicka, for now that's probably the best we can do.

PR #798, opened 2026-03-13
Fixes #39149 - Update tasks table to Patternfly 5

MariaAga:
First commit - update test mocks from core
Depends on https://github.com/theforeman/foreman/pull/10906 
Updates foreman tasks to use the mocks setup from core, and removes all the core mocks that the repo has. This is done to preform better tests that test tasks together with core.
I updated some of the tests in the first commit. But the commit cant go in as its own PR as many other tests break, these tests are either deleted (because the component is deleted) or updated to the new component changes in the second commit (Update tasks table to Patternfly 5)

Second commit Update tasks table to Patternfly 5
Uses TableIndexPage from core to do most of the work 
This also fixes a bug where bulk action (select all specifically, not select page/manual select) never ran and the server returned an error due to a bad query.

Tests code is AI assisted


webpack/ForemanTasks/Components/TasksTable/TasksColumns.js
```diff
@@ -0,0 +1,85 @@
...
+      return (
+        <span className="param-value" title={duration.tooltip}>
+          {duration.text}
+        </span>
+      );
+    },
+  },
+};
+
+export const getRowKebabItems = (setClickedTask, openModal) => ({
```
Lukshio: I think the old way of available actions was more clear, now user has to do one more click every time just to check available options.

Old:

Image

New:
Image

Lukshio (changes requested):

Lukshio:
New react routing works as expected.

pnovotny:
QA-wise: LGTM.

I've verififed the following acceptance criteria (copying from [SAT-32118](https://redhat.atlassian.net/browse/SAT-32118)):
-    Tasks are shown in the table
-    Listed tasks are searchable by using the search field
-    Listed tasks are filterable by the cards above the table that act like preset searches
-    Each task has the correct action enabled (cancel/resume/force cancel)
-    A user can select 1 or more tasks and run bulk actions. If the tasks are from the page, and none can have "action", the bulk_action will not run. if the tasks are selected from all the tasks in the table (compared to tasks selected in the page), bulk action will try to run on the search query, and perform the action
-    Only users with edit task permission should be able to perform actions on tasks.
-    A user can refreshes the table data by using a refresh button (not the user's browser refresh button)
-    A user can export all tasks matching the current search/preset filters. This export includes the data from all pages.
-    Pagination and bookmarks should work as normal to other tables.
-    A user can sorting by duration and started on fields.
-    Clicking on the task name should lead to the task details.

Also addressing other comments from here:


> Firefox only: I'm not able to fully scroll down to the end of the table to see the pagination controls, I always get "bounced back" to some offset

The page scrolling has been fixed in the latest commit. Tried with Firefox 136 & Chromium 146.


> looking at the search select button, I think it behaves strangely, though not sure if related to your changes. When having a selection like this: image clicking the genreal button area deselects everything, so it behaves like hitting the checkbox. I think the general expectation of users would be to see the dropdown when hitting the button and deselect only when hitting the checkbox specifically

Indeed, as mentioned by @MariaAga , this is a general table filter issue, not just on this page (verified). It's definitely worth fixing, but in a separate PR.


One small visual nitpick that I found is the page title alignment:
tasks-title-new-page

The vertical alignment is a little bit off. Here is comparison with the "old" page, where it looks fine:
tasks-title-old-page
Overall, it's a minor thing, I am leaving this as a non-blocking issue.

MariaAga:
Title alignment fits with the new PF5 pages: 
Might just feel a bit different because of the dashboard background colour
Screenshot From 2026-04-01 11-19-27
Screenshot From 2026-04-01 11-19-09
Screenshot From 2026-04-01 11-19-05

Lukshio:
Thanks @MariaAga @pnovotny @pondrejk

PR #808, opened 2026-05-08
Fixes #39294 - PF5 Locks and RTL tests

andreilakatos:


webpack/ForemanTasks/Components/TaskDetails/Components/Locks.js
```diff
@@ -17,35 +25,36 @@ ConditionalLink.defaultProps = {
 
 const Locks = ({ locks }) => (
   <div>
-    <Alert type="info">
+    <Alert
+      variant={AlertVariant.info}
+      ouiaId="task-locks-info"
+      isInline
+      className="pf-u-mb-lg"
+    >
```
MariaAga: New alert style needs a title 
Image

webpack/ForemanTasks/Components/TaskDetails/Components/Locks.js
```diff
@@ -17,35 +25,36 @@ ConditionalLink.defaultProps = {
...
+        <GalleryItem key={key}>
+          <ConditionalLink link={lock.link}>
+            <Card isCompact ouiaId={`task-lock-${key}`}>
+              <CardTitle>
+                <span
+                  className={`pf-u-mr-sm fa ${
+                    lock.exclusive ? 'fa-lock' : 'fa-unlock-alt'
+                  }`}
+                  aria-hidden
+                />
```
MariaAga: Please use PF icons instead of this.

webpack/ForemanTasks/Components/TaskDetails/Components/Locks.js
```diff
@@ -17,35 +25,36 @@ ConditionalLink.defaultProps = {
...
+                {lock.resource_type}
+              </CardTitle>
```
MariaAga: Icon is too close to the title
Image

MariaAga (changes requested):
Thanks!
Could you add space between the alert and the cards ? 

Image

webpack/ForemanTasks/Components/TaskDetails/Components/Locks.js
MariaAga: Could the content in the card be centred again?  (instead of aligned to the left) 

Image

Image

andreilakatos:
image

For the below I just mocked the exclusive resources:
image

@MariaAga

adamruzicka (changes requested):
Please drop the locale changes in `locale` and `app/assets/javascripts/foreman-tasks/locale`. String extraction is usually handled separately from the changes that introduced those string changes

Lukshio (commented):
Hi, maybe I missed something, but why you replaced the Card component with Stack/Flex design? I think that the Card one was a lot better, so I would like to bring it back.

andreilakatos:
> Hi, maybe I missed something, but why you replaced the Card component with Stack/Flex design? I think that the Card one was a lot better, so I would like to bring it back.

No particular reason, just found those and thought that they may be of use. Will replace with Card.

MariSvirik:
Figma: https://www.figma.com/design/ajROfKYDdOEpWyJS2raP84/Task-detail-page?node-id=0-1&p=f&m=dev

Make sure the lock icons next to the titles are center-aligned with the text
The spacing between the header and description should be 8 px (sm). It should be optically closer to the header than to the list/table.

andreilakatos:
> Figma: https://www.figma.com/design/ajROfKYDdOEpWyJS2raP84/Task-detail-page?node-id=0-1&p=f&m=dev
> 
> Make sure the lock icons next to the titles are center-aligned with the text The spacing between the header and description should be 8 px (sm). It should be optically closer to the header than to the list/table.

Don't have access to developer mode figma anymore :(
But I think I did what you asked:
image

Lukshio:
Changes looks good.

Lukshio:
@andreilakatos Can you please replace css flex breakpoints by using Flex components from PF?

.gitignore
```diff
@@ -1,4 +1,5 @@
 .bundle/
+.cursor/*
```
MariaAga: This should be in a separate pr

webpack/ForemanTasks/Components/TaskDetails/Components/Locks.js
```diff
@@ -1,56 +1,181 @@
...
+          headingLevel="h3"
+          size="lg"
+          ouiaId={`${ouiaSectionId}-title`}
+          className="pf-v5-u-m-0"
+        >
+          {title}
+        </Title>
+        <RowIcon aria-hidden />
+      </Flex>
+    </CardHeader>
```
MariaAga: Using a card (bottom padding is 16px) goes against Maria's (S) request of a 8px padding, flex might be a better solution:
(added margin:0 since we have a global css rule to add margin to all titles)
Also it makes the description grey instead of black
```
<Flex direction={{ default: 'column' }}>
    <FlexItem spacer={{ default: 'spacerSm' }}>
      <Flex
        alignItems={{ default: 'alignItemsCenter' }}
        spaceItems={{ default: 'spaceItemsSm' }}
      >
        <FlexItem>
          <Title headingLevel="h3" size="lg"  style={{ margin: '0' }}>
            {title}
          </Title>
        </FlexItem>
        <FlexItem>
          <RowIcon aria-hidden />
        </FlexItem>
      </Flex>
    </FlexItem>
    <FlexItem>
      <TextContent>
        <Text component={TextVariants.p}>{description}</Text>
      </TextContent>
    </FlexItem>
```

MariaAga (changes requested):

webpack/ForemanTasks/Components/TaskDetails/Components/Locks.js
andreilakatos: This goes against the comment from @Lukshio though 😁 
https://github.com/theforeman/foreman-tasks/pull/808#pullrequestreview-4262549878

Lukshio: Well, I thought that it will be better approach using Card, but the suggested changes looks better after all. Anyway there still we be need to use classes for the table width.

MariaAga: This is the result I get with the following code, that doesnt use classes
image

This is the design:
image
```js
<FlexItem>
      <Grid>
        <GridItem span={5}>
          <Table
            aria-label={title}
            variant="compact"
            ouiaId={`${ouiaSectionId}-table`}
          >
            <Tbody>
              {items.map((lock, index) => (
                <Tr
                  key={`${lock.resource_type}-${lock.resource_id}-${index}`}
                  ouiaId={`${ouiaSectionId}-row-${index}`}
                >
                  <Td>
                    <Flex
                      alignItems={{ default: 'alignItemsCenter' }}
                      spaceItems={{ default: 'spaceItemsSm' }}
                      flexWrap={{ default: 'nowrap' }}
                    >
                      <FlexItem>
                        <Icon size="sm">
                          <RowIcon />
                        </Icon>
                      </FlexItem>
                      <FlexItem>
                        {lock.link ? (
                          <a
                            href={lock.link}
                            data-ouia-component-id={`${ouiaSectionId}-resource-type-link-${index}`}
                          >
                            {lock.resource_type}
                          </a>
                        ) : (
                          lock.resource_type
                        )}
                      </FlexItem>
                    </Flex>
                  </Td>
                  <Td>
                    {sprintf(__('id: %s'), String(lock.resource_id))}
                  </Td>
                </Tr>
              ))}
            </Tbody>
          </Table>
        </GridItem>
      </Grid>
```

andreilakatos: image
It needs margin

MariaAga:
```js
    <Flex
      direction={{ default: 'column' }}
      gap={{ default: 'gap2xl' }}
    >
      {nonExclusive.length > 0 && (
        <FlexItem>
...
      {exclusive.length > 0 && (
        <FlexItem>
```

andreilakatos: That doesn't make it look better:
image
I'll just add margin-bottom-2xl:
image

MariaAga: This is what I see 
image

Lukshio (approved):
Code looks good, packit tested.

PR #812, opened 2026-05-14
Fixes #39324 - Preserve sub-second precision in task duration API response

pablomh:
The workaround for ruby/json#957 (commit 88faac6) replaced the plain
`duration` attribute with `.in_seconds.to_s` to avoid a `SystemStackError`
when serializing `ActiveSupport::Duration`. However, `in_seconds` is an
alias for `to_i`, which truncates sub-second precision.

Use `.to_f` instead — also returns a primitive type (Float), avoiding the
serialization recursion, but preserves the fractional seconds.

Before: `"duration": "60"`
After: `"duration": "60.123456"`

### DIFF
--- a/app/views/foreman_tasks/api/tasks/show.json.rabl
+++ b/app/views/foreman_tasks/api/tasks/show.json.rabl
@@ -6,6 +6,6 @@ attributes :id, :label, :pending, :action
 attributes :username, :started_at, :ended_at, :state, :result, :progress
 
 # A workaround for https://github.com/ruby/json/issues/957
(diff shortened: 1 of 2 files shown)


adamruzicka (changes requested):
Test failures need to be addressed. Apart from that I have no objections, both hammer and new PF5 frontend seem to work just fine with it.

pablomh:
Just noticed it. Tests updated.

adamruzicka:
Thank you @pablomh !

PR #815, opened 2026-05-20
Fixes #39339 - use React component for TaskDetails page rendering

andreilakatos:
This PR is needed to be merged before this one:
https://github.com/theforeman/foreman/pull/10992


test/controllers/tasks_controller_test.rb
```diff
@@ -82,15 +82,6 @@ def in_taxonomy_scope(organization, location = nil)
...
 
-      describe 'show' do
-        it 'does not allow user without permissions to see task details' do
-          setup_user('view', 'foreman_tasks', 'owner.id = current_user')
-          get :show, params: { id: FactoryBot.create(:some_task).id },
-                     session: set_session_user(User.current)
-          assert_response :not_found
-        end
-      end
-
```
MariaAga, line 93: This should still be a valid test

config/routes.rb
```diff
@@ -12,7 +12,7 @@
       end
     end
 
-    resources :tasks, :only => [:show] do
+    resources :tasks, :only => [:index] do
```
Lukshio:
```suggestion
    resources :tasks, :only => [:index] do 
```

the `only` can be removed

Lukshio (changes requested):

test/controllers/tasks_controller_test.rb
MariaAga: But should probably be edited to work with the new way

config/routes.rb
andreilakatos: It seems like it's like that in multiple places, not sure if I should remove it

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -7,8 +7,19 @@ class TasksController < ::ApplicationController
     before_action :find_dynflow_task, only: [:unlock, :force_unlock, :cancel, :abort, :cancel_step, :resume]
 
     def show
-      @task = resource_base.find(params[:id])
-      render :layout => !request.xhr?
+      @task = task_detail_scope.find(params[:id])
+
+      respond_to do |format|
+        format.html do
+          response.headers['X-Request-Path'] = request.path
```
adamruzicka: Why is this needed?

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -131,5 +142,17 @@ def filter(scope, paginate: true)
...
+    # Same scoped lookup as SPA entry (authorize + taxonomy + explicit search params).
+    def task_detail_scope
+      scope = ForemanTasks::Task.authorized('view_foreman_tasks')
+      sq = search_query
+
+      if sq.blank?
+        scope
+      else
+        scope.search_for(sq)
+      end
```
adamruzicka: `search_query` is always a string, calling `#search_for` with an empty string is a noop

```suggestion
      ForemanTasks::Task.authorized('view_foreman_tasks')
                        .search_for(search_query)
```

config/routes.rb
```diff
@@ -48,7 +49,7 @@
         end
       end
 
-      resources :tasks, :only => [:show, :index] do
+      resources :tasks, :only => [:index, :show] do
```
adamruzicka: I'm pretty sure we could live without this

app/controllers/foreman_tasks/tasks_controller.rb
andreilakatos: Not 100% sure, ruby is not my area of expertise, did this with AI only. To also validate permissions on the route, but render the react component. Want me to see if it can be removed?

config/routes.rb
andreilakatos: Without the whole resource?

adamruzicka: Without reordering what was already there

app/controllers/foreman_tasks/tasks_controller.rb
adamruzicka: Yes please. In general setting http headers directly is something that shouldn't really be done unless there's a really good reason for it.

app/controllers/foreman_tasks/api/tasks_controller.rb
```diff
@@ -324,7 +324,7 @@ def controller_permission
 
       def action_permission
         case params[:action]
-        when 'bulk_search', 'summary', 'details', 'sub_tasks'
+        when 'bulk_search', 'summary', 'show', 'details', 'sub_tasks'
```
adamruzicka: This should be already covered by the call to super by https://github.com/theforeman/foreman/blob/develop/app/controllers/api/base_controller.rb#L399

adamruzicka: This still applies

andreilakatos:
Tests won't pass on this PR until the foreman-core PR is merged

app/controllers/foreman_tasks/api/tasks_controller.rb
andreilakatos: I actually already changed this if you look at the files changed again 😄

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -107,7 +116,7 @@ def controller_permission
 
     def action_permission
       case params[:action]
-      when 'sub_tasks', 'summary', 'summary_sub_tasks'
+      when 'show', 'sub_tasks', 'summary', 'summary_sub_tasks'
```
adamruzicka: This should be already covered by the call to super by https://github.com/theforeman/foreman/blob/develop/app/controllers/application_controller.rb#L176 . A spin on https://github.com/theforeman/foreman-tasks/pull/815#discussion_r3280337556

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -131,5 +140,11 @@ def filter(scope, paginate: true)
       scope = scope.paginate(page: params[:page], per_page: params[:per_page]) if paginate
       scope.distinct
     end
+
+    # Same scoped lookup as SPA entry (authorize + taxonomy + explicit search params).
+    def task_detail_scope
+      ForemanTasks::Task.authorized('view_foreman_tasks')
+                        .search_for(search_query)
```
adamruzicka: If this is only used in `#show`, then the search doesn't really make sense and `#resource_scope` already checks permissions

config/routes.rb
adamruzicka: Why put index back? A `/tasks` route is already defined on L37

app/controllers/foreman_tasks/api/tasks_controller.rb
adamruzicka: Oh, sorry, I missed there were two different places with this

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -7,8 +7,17 @@ class TasksController < ::ApplicationController
...
+      @task = task_detail_scope.find(params[:id])
+
+      respond_to do |format|
+        format.html do
+          render('react/index', :layout => 'layouts/react_application', :formats => [:html])
+        end
+
+        format.csv do
+          respond_with_tasks(Task.where(:id => @task.id))
+        end
```
adamruzicka: Also, why would we need a csv format for details of a single task?

andreilakatos: Another instance of AI doing more than asked initially

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -7,8 +7,13 @@ class TasksController < ::ApplicationController
...
 
     def show
-      @task = resource_base.find(params[:id])
-      render :layout => !request.xhr?
+      @task = resource_scope.find(params[:id])
+
+      respond_to do |format|
+        format.html do
+          render('react/index', :layout => 'layouts/react_application', :formats => [:html])
+        end
```
adamruzicka: This should be equivalent
```suggestion
      render('react/index', :layout => 'layouts/react_application')
```

config/routes.rb
```diff
@@ -35,6 +35,7 @@
     resources :tasks, :only => [:index], constraints: ->(req) { req.format == :csv }
 
     match '/tasks', to: '/react#index', via: :get
+    get '/tasks/:id', :to => 'tasks#show', :defaults => { :format => 'html' }
```
adamruzicka: I don't think the defaults are necessary
```suggestion
    get '/tasks/:id', :to => 'tasks#show'
```

adamruzicka (commented):
Last two comments

app/controllers/foreman_tasks/tasks_controller.rb
```diff
@@ -7,8 +7,9 @@ class TasksController < ::ApplicationController
     before_action :find_dynflow_task, only: [:unlock, :force_unlock, :cancel, :abort, :cancel_step, :resume]
 
     def show
-      @task = resource_base.find(params[:id])
-      render :layout => !request.xhr?
+      @task = resource_scope.find(params[:id])
+
+      render('react/index', :layout => 'layouts/react_application')
```
adamruzicka: If I'm reading things right then L10 is unnecessary because the component calls the api to fetch the details anyway.

Building on top of that, if we don't need anything from the controller apart from just rendering a full react page that then gets picked up by react router, we should skip the roundtrip and direct the route directly to `/react#index`?

andreilakatos: I did this route because of the testing that @MariaAga said should still be there.
image

For permissions check. Is that working?

adamruzicka: I would lean into Maria's second comment. Yes, the page must still be checking 
user's permissions and that behaviour should be covered with tests, but one could argue that the way it is implemented now is a bit of an antipattern.

andreilakatos: I tried a few options but I can't really test the user permissions for the /tasks/:id without actually routing through ruby first. What do you suggest to do here?

Lukshio: Little bit late to the party, but I think that the `show` function can be deleted, so as the test. As @adamruzicka said, the page loads the data via API, so it should work without this function. But wait for his confirmation before implementing.

andreilakatos:
@adamruzicka recommended to add a wrapper over the TasksDetailPage render function to replace that test that I removed @MariaAga.
image

Lukshio:
@coderabbitai full review

test/controllers/tasks_controller_test.rb
adamruzicka, line 93: Rather than dropping it, it would be nice to convert it to a integration/capybara test

adamruzicka:
> @Lukshio, we couldn't start this review because you've reached your PR review rate limit.

Oh come on, @coderabbitai full review

adamruzicka:
Is a proper loading state planned for later?

While data is being loaded, it shows some sort of a in-between state. Showing the result as `? error` when we just don't know is a bit odd.

image

image

Lukshio:
> Is a proper loading state planned for later?
> 
> While data is being loaded, it shows some sort of a in-between state. Showing the result as `? error` when we just don't know is a bit odd.
> 

When the state is loading, or it's unknown, it fallbacks to default. But some skeleton could be implemented. I would leave that to the next step, when some tabs will be merged and the design will change. So we can do all these loading skeletons at once.

webpack/ForemanTasks/Routes/ShowTaskDetails/TaskDetailsPage.js
```diff
@@ -0,0 +1,71 @@
...
+      <TextContent>
+        <Text ouiaId="breadcrumb_title" component="h1">
+          {headerText}
+        </Text>
+      </TextContent>
+      <FlexItem component="span">
+        {taskResultIconEl(taskState, taskResult)}
+      </FlexItem>
+    </Flex>
+  );
```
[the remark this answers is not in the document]
Lukshio: The component itself does not have this prop, i would replace it only with `id` so it can be easily selected by some future QA automation.

https://v5-archive.patternfly.org/layouts/flex#flex

adamruzicka:
> I would leave that to the next step

No objections against that, but I'd prefer to have na issue for it.

Lukshio:
> > I would leave that to the next step
> 
> No objections against that, but I'd prefer to have na issue for it.
Sure, created new sub task.

https://redhat.atlassian.net/browse/SAT-45708

adamruzicka:
Now the permission wrapper only checks if the user has that permission _at all_. If the user has that permission in general, but does not have the permission to view that particular task or if the task does not exist at all, the backend returns a 404 which the frontend doesn't seem to be able to deal with.

As a minimal reproducer, just navigate to `foreman_tasks/tasks/asdasd`

andreilakatos:
> foreman_tasks/tasks/asdasd

This is what I get:
image

andreilakatos:
@Lukshio @adamruzicka 
Added error states:
image
image
image

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
```diff
@@ -48,15 +52,32 @@ const TaskDetails = ({
...
+
     return (
-      <MessageBox
-        key="task-details-error"
-        icontype="error-circle-o"
-        msg={sprintf(__('Could not receive data: %s'), APIerror?.message)}
+      <EmptyState
+        icon={<ExclamationCircleIcon />}
+        header={__('Error')}
+        description={apiErrorMessage || __('Could not receive task data.')}
```
adamruzicka: `apiErrorMessage` seems to be some sort of a generic error, even if the backend supplies more details

image

webpack/ForemanTasks/Routes/ShowTaskDetails/__tests__/TaskDetailsPage.test.js
```diff
@@ -0,0 +1,254 @@
...
+});
+
+const rootReducer = combineReducers({
+  API: (state = {}, action) => state,
+  intervals: intervalsReducer,
+  breadcrumbBar: breadcrumbBarReducer,
+  foremanTasks: (state = {}, action) => state,
+});
+
+function renderPage(apiPayloadOverrides = {}, propsOverrides = {}) {
```
adamruzicka: _nitpick:_
I'd prefer to pick one style of function definition (`function name` vs `const name = () => { ... }` and sticking with it, at least within the scope of a file.

adamruzicka (commented):
Left some inline comments. The backend looks alright, the edge cases seem to be covered, but I'd prefer if someone else could bless the js parts

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
Lukshio: It will probably need our casual parsing

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
```diff
@@ -48,15 +52,32 @@ const TaskDetails = ({
...
+      return <PermissionDenied missingPermissions={['view_foreman_tasks']} />;
+    }
+
+    if (apiErrorCode === 404) {
+      return (
+        <EmptyState
+          icon={<ExclamationCircleIcon />}
+          header={__('Task not found')}
+          description={
+            apiErrorMessage || __('The requested task could not be found.')
```
Lukshio: Nitpick, the error messages could be used as const

andreilakatos: Not sure what you mean. To put the message in a constant variable?

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
andreilakatos: Did some parsing with the last commit

test/controllers/tasks_controller_test.rb
andreilakatos: Are those tests written inside foreman core? I can't find any capybara / integration tests in this repo.

adamruzicka: > Are those tests written inside foreman core?

No, directly in the plugin.

> I can't find any capybara / integration tests in this repo.

It is possible we just don't have any yet. You could draw inspiration from webhooks for example https://github.com/theforeman/foreman_webhooks/blob/master/test/integration/webhooks_test.rb

andreilakatos: Done, thanks for the link 😀

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
Lukshio: yes, put somewhere on top `const NOT_FOUND_ERR = _('The requested task could not be found.')` and then use the const. We are using it in other files, so just to keep consistency.

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
```diff
@@ -48,15 +52,32 @@ const TaskDetails = ({
...
-  if (status === STATUS.ERROR) {
+  if (status === STATUS.ERROR || apiStatus === STATUS.ERROR) {
+    if (apiErrorCode === 403) {
+      return <PermissionDenied missingPermissions={['view_foreman_tasks']} />;
+    }
+
+    if (apiErrorCode === 404) {
+      return (
+        <EmptyState
+          icon={<ExclamationCircleIcon />}
```
MariaAga: this logic starts to look like https://github.com/theforeman/foreman/pull/10986
```
  if (httpStatus === 403) return FAILURE_REASON.FORBIDDEN;
  if (viewPermissions?.length > 0 && !hasViewPermission) {
    return FAILURE_REASON.FORBIDDEN;
  }
  if (httpStatus === 404) return FAILURE_REASON.NOT_FOUND;
  if (viewPermissions?.length > 0 && hasViewPermission) {
    return FAILURE_REASON.NOT_FOUND;
  }
```

andreilakatos: I guess we can adapt this if that gets merged

webpack/ForemanTasks/Routes/ShowTaskDetails/TaskDetailsPage.js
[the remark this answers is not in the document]
andreilakatos: This is actually failing because the other foreman core PR is not merged yet. Tests won't pass until [this](https://github.com/theforeman/foreman/pull/10992) is merged.

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
Lukshio: @andreilakatos core PR has got merged, so you can adapt it

andreilakatos:
@Lukshio 
Used Reusable Empty State:
image
image

Lukshio (commented):
@andreilakatos CI failures are related, tests needs to be updated

andreilakatos:
> @andreilakatos CI failures are related, tests needs to be updated

Tests won't pass until the PR from foreman core is merged
https://github.com/theforeman/foreman/pull/10992

config/routes.rb
```diff
@@ -27,14 +27,15 @@
         post :cancel_step
       end
     end
-    resources :tasks, :only => [:show], constraints: ->(req) { req.format == :csv } do
+    resources :tasks, :only => [], constraints: ->(req) { req.format == :csv } do
       member do
         get :sub_tasks
       end
     end
```
Lukshio, line 34: I didn't found any usage of this, can someone else confirm it?

ofedoren:
> > @andreilakatos CI failures are related, tests needs to be updated
> 
> Tests won't pass until the PR from foreman core is merged [theforeman/foreman#10992](https://github.com/theforeman/foreman/pull/10992)

Just to be sure: is this PR is dependent on a PR in Foreman? If so, we need to bump the required version in https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L23

UPD: Also, to be sure that the PR in Foreman actually fixes the tests here, we can use a temporary commit to update CI here to use that PR, similarly to https://github.com/theforeman/foreman_webhooks/pull/102

andreilakatos:
> > > @andreilakatos CI failures are related, tests needs to be updated
> > 
> > 
> > Tests won't pass until the PR from foreman core is merged [theforeman/foreman#10992](https://github.com/theforeman/foreman/pull/10992)
> 
> Just to be sure: is this PR is dependent on a PR in Foreman? If so, we need to bump the required version in https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L23
> 
> UPD: Also, to be sure that the PR in Foreman actually fixes the tests here, we can use a temporary commit to update CI here to use that PR, similarly to [theforeman/foreman_webhooks#102](https://github.com/theforeman/foreman_webhooks/pull/102)

Hmm, is the CI running the tests from foreman using specific version? And is a version always created when one PR is merged into core?

ofedoren:
> is the CI running the tests from foreman using specific version

CI for foreman-tasks runs the tests defined within this repo, but since it's a plugin, it requires Foreman to be present as well. In this case we fetch whatever version is on `develop` branch of Foreman repo by default (https://github.com/theforeman/actions/blob/v1/.github/workflows/foreman_plugin.yml#L19-L21).

> And is a version always created when one PR is merged into core

Not sure what you mean by "created", but if you mean bump in a plugin then depends: e.g. if this plugin is currently requires Foreman >= 3.19, but there are changes in Foreman that the plugin needs and those changes are available only on Foreman >= 3.20, then we need to make sure that the plugin requires that *new* minimum.

In development setup it doesn't matter, we always use latest-greatest, but in released versions it does matter. It also makes sure that if we bump minimal version, we don't ship that released version with incompatible Foreman version.

config/routes.rb
MariaAga: Which part is `this`? `sub_tasks` or `show`?

webpack/Routes/routes.test.js
```diff
@@ -53,17 +62,39 @@ describe('ForemanTasks routes', () => {
...
+      },
+      {
+        ...routerProps,
+        match: {
+          ...routerProps.match,
+          params: { id: '99' },
+          path: '/foreman_tasks/tasks/:id',
+          url: '/foreman_tasks/tasks/99',
+        },
+      },
```
MariaAga: Why is this duplicated?

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
```diff
@@ -15,6 +15,8 @@ import { TaskSkeleton } from './Components/TaskSkeleton';
 
 import './TaskDetails.scss';
 
+const TASKS_PATH = '/foreman_tasks/tasks';
```
MariaAga: Needs to be in TaskDetailsConstants

config/routes.rb
Lukshio: Pulling `.csv` files for `sub_tasks`, originally there was `:only => [:show]`. I didn't found any usage of it in UI. Only the `Export All` button on `/tasks` page

webpack/ForemanTasks/Routes/ShowTaskDetails/TaskDetailsPage.js
```diff
@@ -0,0 +1,71 @@
...
+import { taskResultIconEl } from '../../Components/common/taskResultIcon';
+
+const TaskDetailsPage = props => {
+  const { id } = props.match.params;
+  const action = useSelector(selectAction);
+  const taskState = useSelector(selectState);
+  const taskResult = useSelector(selectResult);
+  const headerText = action
+    ? sprintf(__('Details of %s task'), action)
+    : __('Task details');
```
MariaAga:
```suggestion
    : __('Task Details');
```

webpack/ForemanTasks/Routes/ShowTaskDetails/TaskDetailsPage.js
```diff
@@ -0,0 +1,71 @@
...
+        </Text>
+      </TextContent>
+      <FlexItem component="span">
+        {taskResultIconEl(taskState, taskResult)}
+      </FlexItem>
+    </Flex>
+  );
+
+  return (
+    <Permitted requiredPermissions={['view_foreman_tasks']}>
```
MariaAga: no need to use permitted, as the `ResourceLoadFailedEmptyState` will take care of showing the empty state

webpack/ForemanTasks/Routes/ShowTaskDetails/TaskDetailsPage.js
```diff
@@ -0,0 +1,71 @@
...
+      <PageLayout
+        customHeader={header}
+        header={headerText}
+        searchable={false}
+        breadcrumbOptions={{
+          breadcrumbItems: [
```
MariaAga: Missing `  switcher_item_url: foreman_tasks_task_path(:id => ':id')` logic that was deleted in `show.html.erb`
before:

Image

after: 

Image

webpack/ForemanTasks/Components/common/taskResultIcon.js
```diff
@@ -0,0 +1,52 @@
...
+        <Icon status="warning">
+          <ExclamationTriangleIcon />
+        </Icon>
+      );
+    default:
+      return (
+        <Icon>
+          <QuestionCircleIcon />
+        </Icon>
+      );
```
MariaAga, line 51: I think it would be nice to add a title/tooltip to the Icons

webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js
MariaAga: Should be in the consts file `webpack/ForemanTasks/Components/TaskDetails/TaskDetailsConstants.js`

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsSelectors.js
```diff
@@ -103,8 +105,24 @@ export const selectCanEdit = state =>
 
 export const selectStatus = state => selectTaskDetailsResponse(state).status;
 
+export const selectAPIStatus = state =>
+  selectAPIStatusByKey(state, FOREMAN_TASK_DETAILS);
+
```
MariaAga, line 107: `selectStatus = selectTaskDetailsResponse(state).status;` => 
`selectTaskDetailsResponse =  selectAPIResponse(state, FOREMAN_TASK_DETAILS);`
` selectAPIResponse =  selectAPIByKey(state, key).response || {};`

`selectAPIStatus = selectAPIStatusByKey(state, FOREMAN_TASK_DETAILS);` =>
`selectAPIStatusByKey =  selectAPIByKey(state, key).status;`

selectStatus, and selectAPIStatus seem to fulfil the same purpose. When I check the TasksDetails props, it seems that `selectStatus` doesnt get updated and is stuck being `PENDING` while `selectAPIStatus` is RESOLVED.

MariaAga (changes requested):
Went through the code, and looked at the page a bit. Will go over the tests after review comments are addressed

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsSelectors.js
andreilakatos: Yep, `selectStatus` is never updated it seems. That's why I'm using `selectAPIStatus`. I kept it because it seemed to be used later in the Task component. But it's not used anywhere. I'll remove it.

andreilakatos:
> Went through the code, and looked at the page a bit. Will go over the tests after review comments are addressed

Comments addressed 😁

MariaAga (changes requested):
I think the latest change with the status is causing, that when a task is pending, with every reload the page goes into a skeleton view for a second, and then shows the data. The logic should be - if pending & no data -> show skeleton.

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsSelectors.js
```diff
@@ -122,6 +123,7 @@ export const selectAPIErrorCode = state =>
   selectAPIError(state)?.response?.status;
 
 export const selectIsLoading = state =>
+  !selectAPIByKey(state, FOREMAN_TASK_DETAILS).response &&
```
MariaAga: Why not use `selectTaskDetailsResponse.length`?

webpack/ForemanTasks/Routes/ShowTaskDetails/TaskDetailsPage.js
```diff
@@ -0,0 +1,98 @@
...
+        requiredPermissions={[VIEW_FOREMAN_TASKS]}
+        ouiaIdPrefix="task-details-page-empty-state"
+        primaryAction={{
+          label: __('Back to tasks'),
+          url: TASKS_PATH,
+          ouiaId: 'task-details-page-empty-state-tasks-list',
+        }}
+      />
+    );
+  }
```
MariaAga: Can `ResourceLoadFailedEmptyState` be only loaded once in `webpack/ForemanTasks/Components/TaskDetails/TaskDetails.js` ?

webpack/Routes/routes.test.js
```diff
@@ -53,17 +62,30 @@ describe('ForemanTasks routes', () => {
...
     ForemanTasksRoutes.forEach((route, index) => {
       const { unmount } = render(route.render(propsByIndex[index]));
 
-      if (index === 2) {
+      if (index === 1) {
+        expect(
+          screen.getByTestId('task-details-page-stub')
+        ).toBeInTheDocument();
+      } else if (index === 3) {
         expect(screen.getByTestId('show-task-stub')).toBeInTheDocument();
```
MariaAga: there is no `show-task-stub` or index === 3 (ForemanTasksRoutes length is 3)

webpack/Routes/routes.test.js
```diff
@@ -53,17 +62,30 @@ describe('ForemanTasks routes', () => {
...
+          url: '/foreman_tasks/tasks/99',
+        },
+      },
+      {
+        ...routerProps,
+        match: {
+          ...routerProps.match,
+          params: { id: '42' },
+          path: '/foreman_tasks/ex_tasks/:id',
+          url: '/foreman_tasks/ex_tasks/42',
```
MariaAga: ex_tasks doesnt exist anymore

webpack/ForemanTasks/Routes/ShowTaskDetails/__tests__/TaskDetailsPage.test.js
```diff
@@ -0,0 +1,271 @@
...
+describe('TaskDetailsPage', () => {
+  afterEach(() => {
+    window.history.pushState({}, '', '/');
+    mockUseForemanPermissions.mockImplementation(() => new Set(['view_foreman_tasks']));
+  });
+
+  beforeEach(() => {
+    mockUseForemanPermissions.mockImplementation(() => new Set(['view_foreman_tasks']));
+  });
+
```
MariaAga: both reset the mock to the same value - the afterEach reset is redundant since beforeEach runs before every test anyway.

webpack/ForemanTasks/Routes/ShowTaskDetails/__tests__/TaskDetailsPage.test.js
```diff
@@ -0,0 +1,271 @@
...
+  describe('permissions', () => {
+    const VIEW_PERM = 'view_foreman_tasks';
```
MariaAga: thats already defined in the consts file

webpack/ForemanTasks/Components/TaskDetails/TaskDetailsConstants.js
```diff
@@ -1,4 +1,5 @@
 export const FOREMAN_TASK_DETAILS = 'FOREMAN_TASK_DETAILS';
 export const FOREMAN_TASK_DETAILS_SUCCESS = 'FOREMAN_TASK_DETAILS_SUCCESS';
-
+export const TASKS_PATH = '/foreman_tasks/tasks';
 export const TASK_STEP_CANCEL = 'TASK_STEP_CANCEL';
+export const VIEW_FOREMAN_TASKS = 'view_foreman_tasks';
```
MariaAga:
```suggestion
export const VIEW_FOREMAN_TASKS = 'view_foreman_tasks';

```

MariaAga (changes requested):

MariaAga (approved):
Thanks! Tested and it works as expected

PR #819, opened 2026-05-26
Refs #39294 - keep result to the props object

andreilakatos:


Lukshio:
Thanks for this hotfix.

Lukshio (approved):
packit + manually tested, thanks @andreilakatos

PR #821, opened 2026-05-29
Fixes #39376 - remove placeholder showtask component

andreilakatos:


adamruzicka (commented):
What if we dropped both the test and the component? It is just a placeholder anyway

andreilakatos:
> What if we dropped both the test and the component? It is just a placeholder anyway

Seemed like a placeholder to me. Although I was tasked to update snapshot testing for files. We can also remove it I guess, @MariaAga ?

adamruzicka (approved):
`+0 -61`, what's not to like?

adamruzicka:
Thank you @andreilakatos !

PR #836, opened 2026-09-22
Update Transifex host to app.transifex.com

ogajduse:
The legacy www.transifex.com endpoint is no longer served by current tx
clients, which fail with EOF when pulling translations. Every other
Foreman plugin already uses app.transifex.com.


ogajduse:
Does this patch need a Redmine issue?

lhellebr (approved):
LGTM

PR #837, opened 2026-09-22
i18n - pulling from tx

ogajduse:
Re-extracts translatable strings from the source and pulls the current translations from Transifex.

**Translations pulled.** The four languages supported downstream (fr, ja, ko, zh_CN) go to 284 translated strings each, including the Satellite 6.20 round that Red Hat's translators completed last week.

**Strings re-extracted.** The last extraction on `master` was 2026-03-10 (`bb72f08`); since then there have been 49 commits touching `app`/`lib`/`webpack` and five releases. Re-extraction therefore adds **36 msgids** that have never been in Transifex and drops **23** that no longer exist in the source. The 36 land here as untranslated entries — Transifex picks them up from the `.pot` once this merges, and they can then be translated.

Note that `locale/action_names.rb` is gitignored but is an extraction *input*: `gettext:find` scans sources before regenerating it, so running the extraction on a tree where that file is missing silently drops the action-name strings. It was present for this run and all six references are intact.

Extraction and compilation ran in a container with Foreman `develop` and this plugin mounted (`rake plugin:gettext[foreman-tasks]`, `plugin:po_to_json`, `make -C locale mo-files`). Depends on #836, now merged.

Ref: https://issues.redhat.com/browse/SAT-51008


lhellebr:
Superseded by github.com/theforeman/foreman-tasks/pull/838

